PR #3, opened 2016-02-29
src/devices/nm-device.c refactored, g_object_notify() replaced with _…

atulhjp:
…notify() and properties defined with NM_GOBJECT_PROPERTIES_DEFINE()

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -94,8 +94,7 @@ enum {
 };
 static guint signals[LAST_SIGNAL] = { 0 };
 
-enum {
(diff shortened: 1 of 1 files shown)


atulhjp:
src/devices/nm-device.c refactoring , properties now defined with NM_GOBJECT_PROPERTIES_DEFINE( ) and g_object_notify ( ) replaced with _notify ( ),
the pattern we already have in src/nm-ip4-config.c

thom311:
Merged as commit 6d409653b1bc9116bfb58b074f0c61407aef1038

Thanks!

PR #4, opened 2016-03-02
Fix missing commas in list-devices devtypes dict.

alikins:


lkundrak:
Thanks, applied https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=6c59443ad59e3c68915350e6865f2f96c50d3a19

PR #5, opened 2016-04-04
manager: fix crash in platform_link_added

tonyespy:
https://bugzilla.gnome.org/show_bug.cgi?id=764606


thom311:
Hi Tony,

sorry, this is the wrong fix, because we want that functions that fail set the error.

I attached a patch to the bugzilla entry.

Thank you for reporting!!

PR #7, opened 2016-06-02
dns: follow resolv.conf if it is a symlink ('rc-manager=file-follow')

thom311:
Until before 1.2.0, NetworkManager would always write resolv.conf as file, but
if /etc/resolv.conf was a symlink, it would follow the link instead of
replacing it with a file ([1], [2]).

With 1.2.0, we dropped that behavior and added a new 'rc-manager=none' which
writes resolv.conf to /var/run/NetworkManager and symlinks resolv.conf [3].
In case resolv.conf being already a symlink to another target, it would
not be replaced [4].
Also, we added 'rc-manager=file', which always writes /etc/resolv.conf as
file [5].

With 1.4.0, we will rename 'rc-manager=none' to 'rc-manager=symlink' [6].

This commit adds yet another 'rc-manager=file-follow' which restores
the pre-1.2 behavior.

[1] 5761e328b81ce8894c2657ce0994ba401923ba35
[2] https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/324233
[3] 4805be2ed27b71a6099477d86dbc109adb41b819
[4] 583568e12f9e580cd2903811637c9f9b7a2f1088
[5] 288799713dc78bc45e2b0a9cf41d228f5d95315f
[6] cd6a469668028fbc347919ed3580275f9894a1f2

This is a controversial change, because of adding more and more options. Instead we could modify the existing rc-manager=file.
I prefer this, for two reasons:
  - we already released 1.2.0 with a different rc-manager=file behavior
  - I prefer the rc-manager=file behavior over rc-manager=file-follow. I would not replace rc-manager=file with an IMO less desired behavior.


bengal:
Looks good

bengal:
LGTM

thom311:
merged to master: 87a1360505f6fd2f9f699f320fe7d08ddd0eb863

PR #8, opened 2016-06-02
core: reuse common static string buffer in nm_utils_dnsmasq_status_to…

thom311:
…_string()

Instead of declaring our local string buffer @buffer,
reuse nm_utils_to_string_buffer_init().


bengal:
LGTM

thom311:
merged: https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=d66b5e79c7940c169dbf47ca578fc626c2b89b38

PR #9, opened 2016-09-16
Announce capabilities

mvollmer:
See https://bugzilla.redhat.com/show_bug.cgi?id=1375967

### DIFF
--- a/introspection/nm-manager.xml
+++ b/introspection/nm-manager.xml
@@ -381,6 +381,15 @@
     <property name="Version" type="s" access="read"/>
 
     <!--
+        Capabilities:
+
+        The current set of capabilities. See <link
+        linkend="NMCapability">NMCapability</link> for currently
+        defined capability numbers.
+    -->
+    <property name="Capabilities" type="ai" access="read"/>
+
+    <!--
         State:
 
         The overall state of the NetworkManager daemon.
--- a/src/nm-manager.c
+++ b/src/nm-manager.c
@@ -100,6 +100,8 @@ typedef struct {
 } RadioState;
 
 typedef struct {
+	GArray *capabilities;
+
 	GSList *active_connections;
 	GSList *authorizing_connections;
 	guint ac_cleanup_id;
@@ -180,6 +182,7 @@ static guint signals[LAST_SIGNAL] = { 0 };
 
 NM_GOBJECT_PROPERTIES_DEFINE (NMManager,
 	PROP_VERSION,
+	PROP_CAPABILITIES,
 	PROP_STATE,
 	PROP_STARTUP,
 	PROP_NETWORKING_ENABLED,
@@ -5437,6 +5440,45 @@ dbus_connection_changed_cb (NMBusManager *dbus_mgr,
 
 /**********************************************************************/
 
+gboolean
+nm_manager_check_capability (NMManager *self,
+                             NMCapability cap)
+{
+	NMManagerPrivate *priv = NM_MANAGER_GET_PRIVATE (self);
+	int i;
+
+	for (i = 0; i < priv->capabilities->len; i++) {
+		NMCapability test = g_array_index (priv->capabilities, gint, i);
+		if (test == cap)
+			return TRUE;
+		if (test > cap)
+			return FALSE;
+	}
+
+	return FALSE;
+}
+
+static int
+cmp_caps (gconstpointer a, gconstpointer b)
+{
+	return *(gint *)a - *(gint *)b;
+}
+
+void
+nm_manager_set_capability (NMManager *self,
+                           NMCapability cap)
+{
+	NMManagerPrivate *priv = NM_MANAGER_GET_PRIVATE (self);
+
+	if (!nm_manager_check_capability (self, cap)) {
+		g_array_append_val (priv->capabilities, cap);
+		g_array_sort (priv->capabilities, cmp_caps);
+		_notify (self, PROP_CAPABILITIES);
+	}
+}
+
+/**********************************************************************/
+
 NM_DEFINE_SINGLETON_REGISTER (NMManager);
 
 NMManager *
@@ -5482,6 +5524,8 @@ constructed (GObject *object)
 	const NMConfigState *state;
 
 	G_OBJECT_CLASS (nm_manager_parent_class)->constructed (object);
+
+	priv->capabilities = g_array_new (FALSE, FALSE, sizeof (gint));
 
 	_set_prop_filter (self, nm_bus_manager_get_connection (priv->dbus_mgr));
 
@@ -5640,6 +5684,12 @@ get_property (GObject *object, guint prop_id,
 	switch (prop_id) {
 	case PROP_VERSION:
 		g_value_set_string (value, VERSION);
+		break;
+	case PROP_CAPABILITIES:
+		g_value_set_variant (value, g_variant_new_fixed_array (G_VARIANT_TYPE ("i"),
+		                                                       priv->capabilities->data,
+		                                                       priv->capabilities->len,
+		                                                       sizeof(gint)));
 		break;
 	case PROP_STATE:
 		nm_manager_update_state (self);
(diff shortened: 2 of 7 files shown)


mvollmer:
I can't get this to compile:

```
  GISCAN   NM-1.0.gir
[...]
tmp-introspectsx0v99/home/mvo/work/NetworkManager/libnm/tmp-introspectsx0v99/NM-1.0.o:(.data.rel.GI_GET_TYPE_FUNCS_+0x8): undefined reference to `nm_capability_get_type'
collect2: error: ld returned 1 exit status
linking of temporary binary failed: Command '[u'/bin/sh', u'../libtool', u'--mode=link', u'--tag=CC', u'--silent', 'gcc', u'-o', u'/home/mvo/work/NetworkManager/libnm/tmp-introspectsx0v99/NM-1.0', u'-export-dynamic', '-Wall', '-std=gnu89', '-Wshadow', '-Wmissing-declarations', '-Wmissing-prototypes', '-Wdeclaration-after-statement', '-Wfloat-equal', '-Wno-unused-parameter', '-Wno-sign-compare', '-Wstrict-prototypes', '-Wno-unused-but-set-variable', '-Wundef', '-Wimplicit-function-declaration', '-Wpointer-arith', '-Winit-self', '-Wmissing-include-dirs', '-Wno-pragmas', '-g', '-O2', '-Warray-bounds', '-Wunused-value', '-fno-strict-aliasing', '-fdata-sections', '-ffunction-sections', '-Wl,--gc-sections', '-Wno-error', u'tmp-introspectsx0v99/home/mvo/work/NetworkManager/libnm/tmp-introspectsx0v99/NM-1.0.o', '-L.', 'libnm.la', u'-Wl,--export-dynamic', u'-lgmodule-2.0', u'-pthread', u'-lgio-2.0', u'-lgudev-1.0', u'-lgobject-2.0', u'-lglib-2.0']' returned non-zero exit status 1
/usr/share/gobject-introspection-1.0/Makefile.introspection:155: recipe for target 'NM-1.0.gir' failed
```

mvollmer:
> I can't get this to compile:

Got it.

mvollmer:
I guess I should add a nm_manager_check_capability function to libnm.

mvollmer:
> I guess I should add a nm_manager_check_capability function to libnm.

I gave up on this for now.  It should be straigtforward, but needs more knowledge of how libnm loads D-Bus properties than I want to have. :-)

mvollmer:
> I gave up on this for now. It should be straigtforward, but needs more knowledge of how libnm loads D-Bus properties than I want to have. :-)

To expand on this:  I think the following program should just work:

```
#include <stdio.h>
#include <NetworkManager.h>

void
main (int argc, char **argv)
{
  NMClient *client = nm_client_new (NULL, NULL);
  printf ("V: %s\n", nm_client_get_version (client));
  printf ("T: %d\n", nm_client_check_capability (client, NM_CAPABILITY_TEAM));
}
```

but nmdbus_manager_get_capabilities was always returning `NULL`.  So "Capabilities" should be loaded in the same way as "Version", I guess, but I couldn't immediately see how that should be done.

introspection/nm-manager.xml
```diff
@@ -381,6 +381,19 @@
...
+
+        The current set of capabilities.  This is a list of 32 bit
+        integers where specific bits are set when a capability is
+        currently available.  For example, capability number 42 would
+        be represented by bit (42 % 32) == 10 in the element with index
+        (42 / 32) == 1 in the array.
+
+        See <link linkend="NMCapability">NMCapability</link> for currently defined capability numbers.
+    -->
+    <property name="Capabilities" type="au" access="read"/>
```
thom311: this is certainly a way to do it... I like that we can pack 32 capabilities in one number and it's reasonably simple to access the API from the client-side.

On the other hand, I was wondering if we want to reserve a range of capabilities for downstream users. In that case, those capabilities would have a high number, and could not be easily expressed in a bitfield. That is the reason why I was thinking about exposing every capability as individual number. Hm, probably the bitfield *is* nicer to use and more efficient. What are other opinions here?

src/nm-manager.c
```diff
@@ -5437,6 +5442,25 @@ dbus_connection_changed_cb (NMBusManager *dbus_mgr,
 
 /**********************************************************************/
 
+void
+nm_manager_set_capability (NMManager *self,
+						   NMCapability cap)
```
thom311: no tabs here

src/nm-manager.c
```diff
@@ -5641,6 +5665,12 @@ get_property (GObject *object, guint prop_id,
 	case PROP_VERSION:
 		g_value_set_string (value, VERSION);
 		break;
+	case PROP_CAPABILITIES:
+		g_value_set_variant (value, g_variant_new_fixed_array (G_VARIANT_TYPE ("u"),
+															   priv->capabilities,
+															   G_N_ELEMENTS (priv->capabilities),
+															   sizeof(priv->capabilities[0])));
```
thom311: we have this peculiar indention style, where you indent with tabs at first, and for line breaks the additional indents are done with spaces... yeah, it's a bit complicated to get right. Sorry

introspection/nm-manager.xml
bengal: > On the other hand, I was wondering if we want to reserve a range of capabilities for downstream users. In that case, those capabilities would have a high number, and could not be easily expressed in a bitfield. That is the reason why I was thinking about exposing every capability as individual number. Hm, probably the bitfield is nicer to use and more efficient. What are other opinions here?

I don't think that efficiency should matter much (actually I first thought to represent each capability as string and return an array of strings). I like the idea of using an array of u32 where each element is a capability which can assume any value. In this way we can also easily group them for class (for example device capabilities, settings?, etc.).

mvollmer: > On the other hand, I was wondering if we want to reserve a range of capabilities for downstream users.

This could be done by giving them the odd numbers, and we keep the even ones... But that is probably too clever.

mvollmer: >  I like the idea of using an array of u32 where each element is a capability which can assume any value.

I agree.  I used a bitfield because the code was simpler.  If everybody agrees, let's use a set of integers.  Should they be guaranteed to be sorted?  I think that would be nice.

src/nm-manager.c
mvollmer: Like this?  https://www.emacswiki.org/emacs/SmartTabs

thom311: yes, exactly. Cool, there is also a link to a vim plugin, have to try that. Tabs are such an annoyance to me... but gotta stick to the convention.

bengal: Or use this:

https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/contrib/editors/networkmanager-style.el?h=nm-1-4

introspection/nm-manager.xml
mvollmer: Okay, changed to a set of ints, represented as a sorted array.

src/nm-manager.c
mvollmer: Fixed.

mvollmer:
The `make check` failure also happens on master, so I think it is unrelated.

```
ERROR: test-keyfile - missing test plan
ERROR: test-keyfile - exited with status 134 (terminated by signal 6?)
```

thom311:
Hi Marius,

I merged the two patches together with follow-up changes to upstream master:

https://cgit.freedesktop.org/NetworkManager/NetworkManager/log/?h=0b48ad4e48c22ca4d392d5d64701df268ad5ed8e

Thanks.

What is still missing is adding the new property to libnm. But we can do that anytime, when somebody actually requests it...

Also, the `make check` failure is surprising. Doesn't happen to me. Sounds like a bug a different issue that depends on your config. Should be fixed too.

mvollmer:
> I merged the two patches together with follow-up changes to upstream master:

Thanks!

PR #10, opened 2016-09-30
Search in /usr/local before system paths for the dnsmasq binary.

AidanDelaney:
When looking for the dnsmasq (or any) binary, NetworkManager should check /usr/local before it checks any system installed version.  This allows the user to replace the binary with a newer version should they desire and is more consistent with the search behaviour commonly found in $PATH.


bengal:
LGTM

lkundrak:
Applied. Thank you!

https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=05e66d52025d7cbe8bcfafd93c31aa5a5b852566

PR #11, opened 2016-11-29
Fix compilation with Python3

mirh:
has_key() has long been dropped
https://docs.python.org/3.1/whatsnew/3.0.html#builtins

Should still work with anything from 2.3 onwards


thom311:
Thanks. Merged as https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=48402cd52693b29a42ec1653e22755d8ec215f38

PR #12, opened 2016-12-03
Add hint to configuration parameter in resolv.conf header

vinzent:
After setting the hostname with hostnamectl NetworkManager got triggered
and created a empty resolv.conf because no DNS servers had been specified in
ifcfg-* files (Fedora/EL).

As we manage it with a config management tool I wanted to disable this.
It took me some time to discover how to do it. This adds a hint to the
generated where to look for it.

Ref https://bugzilla.redhat.com/show_bug.cgi?id=1344303


lkundrak:
Thanks for the patch, but I'd prefer if we didn't merge it.

It is already rather obvious the file is generated by NetworkManager and it should be sort of natural to look into the NetworkManager manual if you find the behavior undesirable and would like to change it.

That said, only generating resolv.conf when the hostname changes is what's broken and surprising here. But that's a different bug.

thom311:
> That said, only generating resolv.conf when the hostname changes is what's broken and
> surprising here. But that's a different bug.

Sounds like https://bugzilla.gnome.org/show_bug.cgi?id=774885#c8

vinzent:
If you google for NetworkManager and resolv.conf best hits indicate people don't actually look at the NetworkManager config first. https://www.google.ch/search?q=networkmanager+resolv.conf

lkundrak:
> If you google for NetworkManager and resolv.conf best hits indicate people don't actually look at the NetworkManager config first. https://www.google.ch/search?q=networkmanager+resolv.conf

I'm not sure how to interpret those search results. The first few hits seem to be users attempting to reconfigure their resolver without much idea about how to operate their systems? I honestly believe there's nothing we could do about that.

I think that it's very reasonable to assume that the users who edit their configuration files are able to use the online manual. The "man" tool has been around since long years ago and it's essentially what every good UNIX textbook begins with.

vinzent:
of course in an ideal world all Linux users and admins are experts and will not google and will not find the answer to just disable NetworkManager or find some wrong answers because of old docs like chattr +i /etc/resolv.conf. 

But there are people outside which would profit from a hint and maybe head over to the right place for information. And on the other hand it wont hurt the experts if there is a hint the comment. 

I'll stop now. :-)

thom311:
> or find some wrong answers because of old docs like chattr +i /etc/resolv.conf.

note that `chattr +i` is still a fine way to tell NM not to touch resolv.conf ( https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/dns/nm-dns-manager.c?id=a96c819f6f5d1c4faa21b242bea18096049c9674#n1666 )

vinzent:
> note that chattr +i is still a fine way to tell NM not to touch resolv.conf 

yes. but at least for me it is not working. Puppet can't manage it then either. And I suspect this chattr +i solution was introduced because back then it was not possible to disable it?

Disabling NetworkManager is a working solution too. But it seems not to be the best solution to me. I don't wan't to disable NetworkManager and I don't wan't to set chattr +i.

PR #13, opened 2017-02-14
automation: clone from github instead of anongit.freedesktop.org

vbenes:
It looks that freedesktop blocks our git clone requests from time to
time. Github should handle more loads.


thom311:
applied, thanks!

PR #16, opened 2017-02-27
dns: check target of resolv.conf symlink with stat

yrro:
The kernel will now take care of resolving link targets and handling
multiple levels of symlinks.

https://bugzilla.gnome.org/show_bug.cgi?id=779269

### DIFF
--- a/src/dns/nm-dns-manager.c
+++ b/src/dns/nm-dns-manager.c
@@ -1638,27 +1638,28 @@ _resolvconf_resolved_managed (void)
 		"/lib/systemd/resolv.conf",
 		"/usr/lib/systemd/resolv.conf",
 	};
-	GFile *f;
-	GFileInfo *info;
-	gboolean ret = FALSE;
+	struct stat etc_resconf_st;
+	int i;
 
-	f = g_file_new_for_path (_PATH_RESCONF);
-	info = g_file_query_info (f,
-	                          G_FILE_ATTRIBUTE_STANDARD_IS_SYMLINK","\
-	                          G_FILE_ATTRIBUTE_STANDARD_SYMLINK_TARGET,
-	                          G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
-	                          NULL, NULL);
+	if (stat (_PATH_RESCONF, &etc_resconf_st) != 0)
+		/* /etc/resolv.conf does not exist or is a broken link.
+		 * NetworkManager.service starts
+		 * After=systemd-resolved.service, so the target in /run will
+		 * always be in place before this test executes. */
+		return FALSE;
 
-	if (info && g_file_info_get_is_symlink (info)) {
-		ret = nm_utils_strv_find_first ((gchar **) resolved_paths,
-		                                G_N_ELEMENTS (resolved_paths),
-		                                g_file_info_get_symlink_target (info)) >= 0;
+	for (i = 0; i < G_N_ELEMENTS (resolved_paths); i++) {
+		struct stat systemd_resconf_st;
+
+		if (stat (resolved_paths[i], &systemd_resconf_st) != 0) {
+			return FALSE;
+		}
+
+		if (etc_resconf_st.st_dev == systemd_resconf_st.st_dev && etc_resconf_st.st_ino == systemd_resconf_st.st_ino)
+			return TRUE;
 	}
 
-	g_clear_object(&info);
-	g_clear_object(&f);
-
-	return ret;
+	return FALSE;
 }
 
 static void
--- a/data/NetworkManager.service.in
+++ b/data/NetworkManager.service.in
@@ -2,7 +2,7 @@
 Description=Network Manager
 Documentation=man:NetworkManager(8)
 Wants=network.target
-After=network-pre.target dbus.service
+After=network-pre.target dbus.service systemd-resolved.service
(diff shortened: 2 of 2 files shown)


src/dns/nm-dns-manager.c
```diff
@@ -1637,28 +1637,25 @@ _resolvconf_resolved_managed (void)
 		"/run/systemd/resolve/resolv.conf",
 		"/lib/systemd/resolv.conf",
 		"/usr/lib/systemd/resolv.conf",
+		NULL
```
thom311: let's not make the array NULL terminated, instead use a index variable like:

    for (i = 0; i < G_N_ELEMENTS (resolved_paths); i++) {
          // use resolved_paths[i]
    }

src/dns/nm-dns-manager.c
```diff
@@ -1637,28 +1637,25 @@ _resolvconf_resolved_managed (void)
...
-		                                G_N_ELEMENTS (resolved_paths),
-		                                g_file_info_get_symlink_target (info)) >= 0;
-	}
+	for (p = resolved_paths[0]; *p; p++) {
+		struct stat systemd_resconf_st;
 
-	g_clear_object(&info);
-	g_clear_object(&f);
+		if (stat (p, &systemd_resconf_st) != 0)
+			continue;
```
thom311: if stat-ing the first path (`/run/systemd/resolve/resolv.conf`) fails, it is fair to assume that resolved is not running. I would just return FALSE right away.

yrro: Now I'm wondering how to deal with the race between NM and systemd-resolved starting up. What should NM's behaviour be in the case that `/etc/resolv.conf` is a symlink and is broken?

thom311: Good point. How about:
 - if /etc/resolv.conf is not a symlink, return FALSE.
 - if /etc/resolv.conf is one of the 3 well-known symlinks, return TRUE (current master's behavior)
 - if the stat() check reveals the symlink points to resolved's file, return TRUE (your stat check)
 - otherwise, return FALSE.

yrro: I think it's sufficient to add `After=systemd-resolved.service` to `NetworkManager.service`. This will guarantee that NM will only be started after resolved has signalled to systemd that it is ready to process requests. (It does *not* have the effect of causing a request to start NM to also start resolved--that's done with the `Wants=` or `Requires=` directive, which we don't use).

I can add back the check for `/etc/resolv.conf` being a symlink itself, although it is not strictly necessary, and without the check the code even works for the edge case where the administrator unintentionally hardlinked `/etc/resolv.conf` to `/lib/systemd/resolv.conf`.

thom311: I wouldn't do that... "After" delays startup, for a very obscure case (yes, I know: it delays it only if resolved is actually enabled).

The entire _resolvconf_resolved_managed() is a heuristic to determine whether to use resolved. The user already can bypass that by configuring the "dns" plugin explicitly.
The heuristic from https://github.com/NetworkManager/NetworkManager/pull/16#discussion_r103333709 should work fairly well.

bengal:
The only concern I have about steps in comment 16 is that they handle differently symlinks with absolute and relative targets (relative symlinks are matched only if the target exist, absolute are matched only by name).
Maybe it would be better to just use realpath() (or the GLib equivalent API) to resolve symlinks and then do the comparison, without stat()?

yrro:
If I'm reading the output of `systemd-analyze blame` correctly, `systemd-resolved` took 400ms to start up. If that's too long to wait then I think it would be better to go back to resolving the symlink to an absolute path, and accept that we won't handle the case where `resolv.conf` points to another symlink.

(That said, what happens if NM then tries to push DNS info to resolved before it's ready? Presumably it will wait until the service is ready, but I haven't checked.)

Anyway, I implemented that with gio at https://github.com/NetworkManager/NetworkManager/compare/master...yrro:networkd-resconf-abspath. I don't think the target of a pull request can be changed, so if you want to go for that approach, I suggest you close this PR and open a new one pointing at that branch.

thom311:
400 ms is surprisingly long. Maybe it was waiting for something else? Anyway, to me that only confirms that we don't want to add "After=".

I am not familiar with the gfile API, but if you'd use good old "realpath()" then "we won't handle the case where resolv.conf points to another symlink" is not an issue, as it will follow all symlinks to the end.

I still would do all checks:

  - if /etc/resolv.conf is a symlink:
       - if "g_file_read_link()" returns one of the 3 special paths, return TRUE
       - if "realpath()" resolves to one of the 3 special paths, return TRUE
       - otherwise, return FALSE
  - if /etc/resolv.conf is a regular file
       - stat() it and compare the inode with the inodes of each of the 3 special paths

bengal:
> I still would do all checks:
> 
>     if /etc/resolv.conf is a symlink:
>         if "g_file_read_link()" returns one of the 3 special paths, return TRUE
>         if "realpath()" resolves to one of the 3 special paths, return TRUE
>         otherwise, return FALSE
>     if /etc/resolv.conf is a regular file
>         stat() it and compare the inode with the inodes of each of the 3 special paths

Sounds good to me.

thom311:
Merged approach from comment https://github.com/NetworkManager/NetworkManager/pull/16#issuecomment-283311368 : fae84b16f850c8a9d6a5f5c4bee1c162debb8dec

PR #17, opened 2017-05-11
Ignore RTM_GETLINK messages sent by wireless extentions

mar-kolya:
We listen to all RTM_GETLINK messages to get updates on interfaces statuses.
Unfortunately wireless code in the kernel sends those messages with wireless information included
and all other information excluded. When we receive such message we wipe out our valid cached entry
with new object that is almost empty because netlink message didn't contain any information.

Solution to this is to check that incoming message contains MTU field: this field is always
set for complete messages about interfaces and is not set by wireless code.

Signed-off-by: Nikolay Martynov <mar.kolya@gmail.com>

### DIFF
--- a/src/platform/nm-linux-platform.c
+++ b/src/platform/nm-linux-platform.c
@@ -1529,6 +1529,23 @@ _new_from_nl_link (NMPlatform *platform, const NMPCache *cache, struct nlmsghdr 
 	if (!obj->link.name[0])
 		goto errout;
 
+	if (!tb[IFLA_MTU]) {
+		/* Kernel has two places that send RTM_GETLINK messages:
+		 * net/core/rtnetlink.c and net/wireless/ext-core.c.
+		 * Unfotunatelly ext-core.c sets only IFLA_WIRELESS and
+		 * IFLA_IFNAME. This confuses code in this function, because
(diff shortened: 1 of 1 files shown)


thom311:
At other places, we take the partial information that is in the netlink message and complete it with the data we have in cache. See "completed_from_cache": https://github.com/NetworkManager/NetworkManager/blob/e70dfb4df7ea4f7787d16f58695e583a4e520c8d/src/platform/nm-linux-platform.c#L1649

Are you confident that these incomplete messages are entirely useless for NM and we can safely ignore them altogether?

mar-kolya:
I'm relatively confident that those particular messages do not contain any information useful for this specific function: they all seem to be results of wifi scans.

I did realize that there is already code that merges from cache. I can see if I can rewrite this patch to merge other fields too - if that's what you would prefer.

thom311:
> I did not realize that there is already code that merges from cache. I can see if I can rewrite this patch to merge other fields too - if that's what you would prefer.

No, since you seem sufficiently confident, that seems not necessary.

Patch lgtm.

bengal:
Looks good to me

thom311:
Patch merged. 

Thanks Nikolay

PR #18, opened 2017-05-14
Handle carrier changes by bond devices

mar-kolya:
This batch changes thy way carrier changes are handled by bond devices:
* pulling cable out of one of the slaves brings slave into disconnected state.
* reconnecting cable reconnects slave automatically.
* pulling out cables of all slaves brings down bond interface - removing its IP configuration.
* reconnecting cable to one of the slaves makes bond interface reistablish its IP configuration.

With this patch bond interface are made to work like all any other interface supporting carrier.
This patch has also been tested with bond interface havind one ethernet and one wifi device - and it works as expected.

This patch allows one to have bond interface between wifi and ethernet and still be able to connect to other wifi networks and not have ip config or routing clashes.

Signed-off-by: Nikolay Martynov <mar.kolya@gmail.com>

### DIFF
--- a/src/devices/nm-device-bond.c
+++ b/src/devices/nm-device-bond.c
@@ -58,24 +58,6 @@ get_generic_capabilities (NMDevice *dev)
 }
 
 static gboolean
-is_available (NMDevice *dev, NMDeviceCheckDevAvailableFlags flags)
-{
-	return TRUE;
-}
-
-static gboolean
-check_connection_available (NMDevice *device,
-                            NMConnection *connection,
-                            NMDeviceCheckConAvailableFlags flags,
-                            const char *specific_object)
-{
-	/* Connections are always available because the carrier state is determined
-	 * by the slave carrier states, not the bonds's state.
-	 */
-	return TRUE;
-}
-
-static gboolean
 check_connection_compatible (NMDevice *device, NMConnection *connection)
 {
 	NMSettingBond *s_bond;
@@ -623,9 +605,7 @@ nm_device_bond_class_init (NMDeviceBondClass *klass)
 	NM_DEVICE_CLASS_DECLARE_TYPES (klass, NM_SETTING_BOND_SETTING_NAME, NM_LINK_TYPE_BOND)
 
 	parent_class->get_generic_capabilities = get_generic_capabilities;
-	parent_class->is_available = is_available;
 	parent_class->check_connection_compatible = check_connection_compatible;
-	parent_class->check_connection_available = check_connection_available;
 	parent_class->complete_connection = complete_connection;
 
 	parent_class->update_connection = update_connection;
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -2195,24 +2195,33 @@ carrier_changed (NMDevice *self, gboolean carrier)
 		return;
 
 	if (priv->is_master) {
-		/* Bridge/bond/team carrier does not affect its own activation,
-		 * but when carrier comes on, if there are slaves waiting,
-		 * it will restart them.
-		 */
-		if (!carrier)
-			return;
-
-		if (nm_device_activate_ip4_state_in_wait (self))
-			nm_device_activate_stage3_ip4_start (self);
-		if (nm_device_activate_ip6_state_in_wait (self))
-			nm_device_activate_stage3_ip6_start (self);
-
-		return;
-	} else if (priv->is_enslaved && !carrier) {
-		/* Slaves don't deactivate when they lose carrier; for
-		 * bonds/teams in particular that would be actively
-		 * counterproductive.
-		 */
+		if (carrier) {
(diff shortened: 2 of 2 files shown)


thom311:
Hi Nikolay,

> pulling out cables of all slaves brings down bond interface - removing its IP configuration.

are you describing here the behavior with your patch? That sounds wrong to me, the code on current master seems as intended: the master effectively ignores carrier, it doesn't go down when slaves get unenslaved or loose carrier.

@bengal, what is your opinion?

mar-kolya:
Yes, sorry, I should have been more clear: I'm describing behaviour I get with the patch.

I think current behaviour (ignoring carrier) is somewhat problematic: if all links get unplugged from one network and repplugined into another it is a good idea to renew ip configurations.

This might seem as somewhat theoretical problem in datacenter setting, but it becomes more prominent if one tries to use interface bonding to bond wifi/wired interfaces to get ability to roam between wired and wireless without loosing connections. Current behaviour of keeping IPs on completely disconnected bond interface means that those IPs may 'clash' with another networks: e.g. I take my laptop out to a cafe and cafe's wifi uses overlapping IP ranges with my home network (where I use bonding).

Also, just to clarify: my patch doesn't unenslave slaves when they are unplugged - all slave interfaces are still enslaved. Also: carrier 'value' itself is handled by link kernel: if all slaves loose carrier then bond interface is marked as lost carrier as well.

mar-kolya:
Just a kind ping.

Should I move this patch into mailing list instead?

bengal:
> pulling out cables of all slaves brings down bond interface - removing its IP configuration.

I think this is a problem and breaks valid uses cases, as the one in which there is a static configuration on the master and a daemon bound to that address. With you patch, if I understand well, it won't be possible to start the daemon until a slave gets the carrier and the master connection can activate.

mar-kolya:
@bengal, thank you for your response!

Situation that you are describing is exactly what would happen on non-bond interface with static IP configuration: if one pulls out cable static (or dynamic, for that matter) ip config goes down. Yes, with this patch same behaviour would apply to bond interfaces.

For both ethernet and bond interfaces 'ignore-carrier' attribute works and mediates this problem: with `ignore-carrier` enabled interface will not loose ip config if carrier goes down.

Also, in my unrelated experience usually services bind to '0.0.0.0' to avoid these kinds of problems of IP reconfiguration (and all necessary access control is done with firewall) . Binding to specific IP on physical interface seems like may cause all sort of other issues.

So, would you find above a satisfactory explanation?

Alternative to this patch seems to be to make this behaviour configurable - which essentially means introducing `observe-carrier` for bond devices. I can obviously try to work on that if that is what you would prefer - but in my very humble opinion this may unnecessarily complicate code.

Please let me know your thoughts.
Thanks!

mar-kolya:
Hi, I would like to make a kind ping on this again.
I think this is the only important thing in NM 'core' that prevents one from using wifi+wired bonding.

thom311:
I think the patch as-is is not correct. Even if it makes sense, it is too much of a change in behavior.

How about https://github.com/NetworkManager/NetworkManager/commits/th/master-no-ignore-carrier-pr18 ?

It does some thing differently then your patch. 

  - you have to opt-in to the new behavior.

  - your patch moves the master to DISCONNECTED state. Why not UNAVAILABLE? IMO, disconnecting a software device should eventually result in the deletion of it. Why having a disconnected bond? Later, if there is a slave ready to (auto)connect, the master will be automatically created as needed.

  - why does your patch do nm_device_update_dynamic_ip_setup()? If that's necessary, that change seems not related to the commit, and should be a separate commit.

thom311:
> * your patch moves the master to DISCONNECTED state. Why not UNAVAILABLE? IMO, disconnecting a software device should eventually result in the deletion of it. Why having a disconnected bond? Later, if there is a slave ready to (auto)connect, the master will be automatically created as needed.
>
> * why does your patch do nm_device_update_dynamic_ip_setup()? If that's necessary, that change seems not related to the commit, and should be a separate commit.

Ok, I think I understand now... I adjusted the patch to make it similar to yours. Maybe that's correct... Needs more testing.

mar-kolya:
@thom311 thanks for looking into this, I really appreciate it!
I'll test out your branch and report back.

mar-kolya:
Also, I should mention that I have been running my laptop with this patch since a few days before it was submitted and it seems to be working fine at least in my setup.

bengal:
In branch th/master-no-ignore-carrier-pr18:

    If the user decides to configure the master device like

	[device-with-carrier]
	match-device=type:bond,type:bridge,type:master

s/master/team/ ?

The rest LGTM.

thom311:
branch merged to master as commit https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=a84facb90c8ed7214e2aeba1ce54909c00a82b2c

PR #21, opened 2017-05-30
nmtui connect: avoid segfault when iface is not found

BlackYoup:
Hi,

I'm not sure why but when I use `nmtui connect` with a network that doesn't exist, the iface (not sure what that is). `nm_device_get_ip_iface` returns `NULL` and the following `strcmp` segfaults

This patch avoids the segfault


thom311:
merged to master as 	1fcbb69ae22ed4a6047e36c816f49b2a67a13583 and backported to nm-1-8 as 	cc36b9f6eb7870d02e0df7c26a7a980709b620fd .

thanks!!

PR #22, opened 2017-06-03
Th/wifi ap hidden

thom311:
Minor fix for hidden flag of AP connection

If you create a connection with `nmcli device wifi hotspot`, you will see that it is wrongly marked as hidden, and also wrongly probe-scanned.


bengal:
LGTM

thom311:
merged as 14529f5b4635f8b856b7698a2eebc687b31c516e

PR #23, opened 2017-07-28
Add support for disabling connectivity checking via the D-Bus interface 

cassidyjames:
This was mentioned in the [Ubuntu Desktop weekly newsletter](https://lists.ubuntu.com/archives/ubuntu-desktop/2017-July/005070.html), but wasn't PR'd here. Patch files also available over at [Bugzilla](https://bugzilla.gnome.org/show_bug.cgi?id=785117)

### DIFF
--- a/introspection/org.freedesktop.NetworkManager.xml
+++ b/introspection/org.freedesktop.NetworkManager.xml
@@ -444,6 +444,28 @@
     <property name="Connectivity" type="u" access="read"/>
 
     <!--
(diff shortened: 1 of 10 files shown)


thom311:
Hi Cassidy,

it's fine to send patches either to the mailing list, attach to a bugzilla bug, or here on github. But let's keep the discussion focused on one place only.

I close this PR, and let's follow up on https://bugzilla.gnome.org/show_bug.cgi?id=785117 .

PR #28, opened 2017-09-14
th/ifcfg route parse

thom311:
Refactor parsing of routes in ifcfg-rh plugin.

Soon we will add more route attributes ("table", maybe "onlink").

I think it's fundamentally wrong to do a fuzzy parsing, with accepting any invalid configuration (wrongly).
I think parsing should be strict, and individual cases/options must be explicitly implemented.

Yes, this change results in rejecting some hand-written files, that contain unexpected fields.


bengal:
LGTM

thom311:
merged to master: https://github.com/freedesktop/NetworkManager/commit/63104b721356dce8f31113a85232b219d45bfd04

PR #31, opened 2017-10-05
dhcp: add support for dhcpcanon client

juga0:
https://github.com/juga0/dhcpcanon is a DHCP client implementing RFC 7844.
This PR works with current version of dhcpcanon and with nm version in commit https://github.com/NetworkManager/NetworkManager/commit/142ebb10376ec592593f15b0359f38be89c97620.
With yesterday nm version it is giving:
    ERROR:src/dhcp/nm-dhcp-client-logging.h:36:_nm_dhcp_client_get_domain: code should not be reached
Comments and suggestions very welcomed. Thanks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -852,6 +852,25 @@ else
 fi
 AM_CONDITIONAL(WITH_OFONO, test "${with_ofono}" = "yes")
 
+# DHCP client support with dhcpcanon
+AC_ARG_WITH([dhcpcanon],
+            AS_HELP_STRING([--with-dhcpcanon=yes|no|path], [Enable dhcpcanon support (experimental)]))
+if test "$with_dhcpcanon" != "no"; then
+    with_dhcpcanon_="$with_dhcpcanon"
+    AC_PATH_PROGS(with_dhcpcanon, dhcpcanon, no, /sbin:/usr/sbin:/usr/local/sbin:/usr/bin:/usr/local/bin)
+    if test "$with_dhcpcanon" == "no"; then
+        if test "$with_dhcpcanon_" == yes; then
+            AC_MSG_WARN([dhcpcanon not found, assume path /sbin/dhcpcanon])
+            with_dhcpcanon=/sbin/dhcpcanon
+        fi
+    fi
+fi
+if test "$with_dhcpcanon" != "no"; then
+    AC_DEFINE(WITH_DHCPCANON, 1, [Define if you have dhcpcanon])
+    AC_SUBST(DHCPCANON_PATH, $with_dhcpcanon)
+else
+    AC_DEFINE(WITH_DHCPCANON, 0, [Define if you have dhcpcanon])
+fi
 # DHCP client support
 AC_ARG_WITH([dhclient],
             AS_HELP_STRING([--with-dhclient=yes|no|path], [Enable dhclient 4.x support]))
@@ -933,6 +952,7 @@ AC_ARG_WITH(config-dhcp-default,
 if test "$config_dhcp_default" = yes -o "$config_dhcp_default" = no; then
 	config_dhcp_default=''
 fi
+test -z "$config_dhcp_default" -a "$with_dhcpcanon" != "no" && config_dhcp_default='dhcpcanon'
 test -z "$config_dhcp_default" -a "$with_dhclient" != "no" && config_dhcp_default='dhclient'
 test -z "$config_dhcp_default" -a "$with_dhcpcd"   != "no" && config_dhcp_default='dhcpcd'
 test -z "$config_dhcp_default"                             && config_dhcp_default='internal'
@@ -1019,7 +1039,7 @@ AC_SUBST(DNSSEC_TRIGGER_SCRIPT)
 
 # system CA certificates path
 AC_ARG_WITH(system-ca-path,
-            AS_HELP_STRING([--with-system-ca-path=/path/to/ssl/certs], [path to system CA certificates])) 
+            AS_HELP_STRING([--with-system-ca-path=/path/to/ssl/certs], [path to system CA certificates]))
 if test "x${with_system_ca_path}" = x; then
 	SYSTEM_CA_PATH=/etc/ssl/certs
 else
@@ -1358,6 +1378,7 @@ echo "  config-dns-rc-manager-default: ${config_dns_rc_manager_default}"
 echo
 
 echo "DHCP clients (default $config_dhcp_default):"
+echo "  dhcpcanon: $with_dhcpcanon"
 echo "  dhclient: $with_dhclient"
 echo "  dhcpcd: $with_dhcpcd"
 echo "  dhcpcd-supports-ipv6: $with_dhcpcd_supports_ipv6"
--- a/src/dhcp/nm-dhcp-dhcpcanon.c
+++ b/src/dhcp/nm-dhcp-dhcpcanon.c
@@ -1,0 +1,272 @@
+/* -*- Mode: C; tab-width: 4; indent-tabs-mode: t; c-basic-offset: 4 -*- */
+/* nm-dhcp-dhcpcanon.c - dhcpcanon specific hooks for NetworkManager
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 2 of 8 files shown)


configure.ac
```diff
@@ -787,6 +787,25 @@ else
 fi
 AM_CONDITIONAL(WITH_OFONO, test "${with_ofono}" = "yes")
 
+# DHCP client support with dhcpcanon
+AC_ARG_WITH([dhcpcanon],
+            AS_HELP_STRING([--with-dhcpcanon=yes|no|path], [Enable dhcpcanon support]))
+if test "$with_dhcpcanon" != "no"; then
```
thom311: Could you also add `--with-dhcpcanon=no` to ` contrib/fedora/rpm/NetworkManager.spec`? After `--with-dhcpcd=no`

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+	}
+}
+
+static void
+state_changed (NMDhcpClient *client,
+               NMDhcpState state,
+               GObject *ip_config,
+               GHashTable *options)
+{
+	gs_unref_bytes GBytes *client_id = NULL;
```
thom311: src/dhcp/nm-dhcp-dhcpcanon.c:248:25: error: unused variable 'client_id' [-Werror,-Wunused-variable]

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Copyright (C) 2017 juga <juga at riseup dot net>
+ */
+
+#include <config.h>
```
thom311: drop this include, and the following until "nm-default.h". Have "nm-default.h" included as first.

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+#if WITH_DHCPCANON
+
+#include <string.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <ctype.h>
```
thom311: none of the above includes are required AFAIS. Just drop them.

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+#include "nm-dhcp-manager.h"
+#include "NetworkManagerUtils.h"
+#include "nm-dhcp-listener.h"
+#include "nm-dhcp-client-logging.h"
+
+/*****************************************************************************/
+
+/* TODO: uncomment to merge with current master
+ * static const char *
+ * _addr_family_to_path_part (int addr_family)
```
thom311: on master, there is now `nm_utils_addr_family_to_char()`. Use that. It gives '4' instead of "" for IPv4, but that seems desired.

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+	const char *iface, *system_bus_address, *dhcpcanon_path = NULL;
+	char *binary_name, *cmd_str, *pid_file = NULL, *system_bus_address_env = NULL;
+	/* TODO: uncomment to merge with current master
+	int addr_family; */
+
+	g_return_val_if_fail (priv->pid_file == NULL, FALSE);
+
+	iface = nm_dhcp_client_get_iface (client);
+	/* TODO: uncomment to merge with current master
+	addr_family = nm_dhcp_client_get_addr_family (client); */
```
thom311: can you rebase the branch on master and fix these TODO?

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+	g_ptr_array_add (argv, NULL);
+
+	cmd_str = g_strjoinv (" ", (gchar **) argv->pdata);
+	g_free (cmd_str);
+
+	if (g_spawn_async (NULL, (char **) argv->pdata, NULL,
+	                   G_SPAWN_DO_NOT_REAP_CHILD | G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL,
+	                   nm_utils_setpgid, NULL, &pid, &error)) {
+		 g_assert (pid > 0);
+ 		_LOGI ("dhclient started with pid %d", pid);
```
thom311: not "dhclient"

src/dhcp/nm-dhcp-dhcpcanon.c
```diff
@@ -0,0 +1,308 @@
...
+	_LOGW ("the dhcpcd backend does not support IPv6");
+	return FALSE;
+}
+static void
+stop (NMDhcpClient *client, gboolean release, const GByteArray *duid)
+{
+	NMDhcpDhcpcanon *self = NM_DHCP_DHCPCANON (client);
+	NMDhcpDhcpcanonPrivate *priv = NM_DHCP_DHCPCANON_GET_PRIVATE (self);
+
+	/* Chain up to parent */
```
thom311: let's remove comments stating the obvious

src/dhcp/nm-dhcp-dhcpcanon.c
juga0: Thanks, i think then this can also be changed: https://github.com/NetworkManager/NetworkManager/blob/master/src/dhcp/nm-dhcp-dhclient.c#L127

juga0:
@thom311 thanks a lot for the review, i think i've solved the comments, further review would be great.

juga0:
> I wish we wouldn't add new DHCP plugins, instead make the existing one working better. 

I understand this point, though it's possitive to have implementations in other languages.

> For example, there certainly is the aim to support anonymization options for internal and dhclient.

The good new is that internal (systemd) (will?) already have this feature: https://github.com/systemd/systemd/commit/504cef76d4ce62e8c58adb31c169ad8bb3879e33
I don't know whether is there any work in this direction in dhclient.

thom311:
> The good new is that systemd already have this feature: systemd/systemd@504cef7, and is already in 
> internal: https://github.com/NetworkManager/NetworkManager/blob/master/src/systemd/src/libsystemd-network/sd-dhcp-client.c#L67 !! :)
> So now we should only add an option to nm config to use "Anonymize=yes" with internal.

Yes, except that the "only" part is the a major part. It interacts with other options, it's some effort to add that to all the places where it matters. Also, only working with "internal" is very unsatisfying, because internal isn't the default DHCP client on many installations. But yes. TODO.

src/dhcp/nm-dhcp-dhcpcanon.c
thom311: nm-dhcp-dhclient.c uses `_addr_family_to_path_part()` which returns "" for AF_INET. It is different from `nm_utils_addr_family_to_char()`, which would return '4'. It needs to stay this way, to preserve previous behavior. But for the new client, let's not do that.

thom311:
Can you please apply the fixup from https://paste.fedoraproject.org/paste/MaxeIQ5jssc9Yi-yv9HVZw and rebase to master again?

Then it looks good to me. Thanks!

juga0:
I'm sorry i lost previous changes when merging with the new ones, now it should be fine, applied your patch and rebased to master. Thanks for your review.

bengal:
Since the new backend didn't get as much testing as the existing ones, can we mark is as experimental for now? I mean:

  AS_HELP_STRING([--with-dhcpcanon=yes|no|path], [Enable dhcpcanon support (experimental)])) 

And maybe build it only when --with-dhcpcanon=yes|path is explicitly set.

Is there any plan to add IPv6 support?

src/dhcp/nm-dhcp-dhcpcanon.c
juga0: Thanks for the clarification.

juga0:
> Since the new backend didn't get as much testing as the existing ones, can we mark is as experimental for now? I mean:

> AS_HELP_STRING([--with-dhcpcanon=yes|no|path], [Enable dhcpcanon support (experimental)]))

I've added "experimental".

> And maybe build it only when --with-dhcpcanon=yes|path is explicitly set.

I thought that having "--with-dhcpcanon=no" (https://github.com/NetworkManager/NetworkManager/pull/31/commits/cc728d4dc1264184580b1d67501a84176085a229#diff-9fcbb86337fcde4e7560f5faa541f7bbR376), as @thom311 commented, will make fedora to do not build it by default.

> Is there any plan to add IPv6 support?

Yes. I thought to create other PR when that is ready.

Thanks for the comments @bengal

juga0:
@thom311 is there something else that should be changed to get this PR merged?.
Thanks.

thom311:
> @thom311 is there something else that should be changed to get this PR merged?.

Nope. It's good, just forgotten about it. Merged now, thanks!!

PR #34, opened 2017-10-27
contrib/fedora: Add mockbuild.sh

mopsfelder:
This script comes handy to build from git tree without having to
install build dependencies on the system, the actual build happens
inside mock environment.


thom311:
merged. Thanks!!

PR #35, opened 2017-11-01
Add mDNS support for DNS backends which support them.

ipuustin:
Add a new setting ``mdns`` to connections. This setting can be used for enabling support for mDNS resolving and mDNS host registration for a given interface. Also add a function (``update_mdns``) to DNS plugin interface and implement this function for systemd-resolved plugin.

So, in other words, if a connection has ``mdns=1``, enable mDNS support in systemd-resolved for the interface that the connection uses.


ipuustin:
Rebased...

libnm/libnm.ver
```diff
@@ -948,6 +949,7 @@ global:
 	nm_setting_connection_autoconnect_slaves_get_type;
 	nm_setting_connection_get_autoconnect_slaves;
 	nm_setting_connection_get_lldp;
+	nm_setting_connection_get_mdns;
```
thom311: As 1.10-rc1 is already released, we better don't change API anymore (not even extend)

These new entries, must move to a new linker section libnm_1_12_0

libnm-core/nm-setting-connection.c
```diff
@@ -1982,4 +2005,34 @@ nm_setting_connection_class_init (NMSettingConnectionClass *setting_class)
...
+	 * do not register hostname but allow resolving of mDNS host names.
+	 * When updating this property on a currently activated connection,
+	 * the change takes effect immediately.
+	 *
+	 * This feature requires a plugin which supports mDNS. One such
+	 * plugin is dns-systemd-resolved.
+	 **/
+	/* ---ifcfg-rh---
+	 * property: mdns
+	 * variable: CONNECTION_MDNS(+)
```
thom311: You didn't implement this flag in ifcfg-rh reader/writer, but it's kinda important to do (yeah, that sucks).

libnm-core/nm-setting-connection.c
```diff
@@ -1982,4 +2005,34 @@ nm_setting_connection_class_init (NMSettingConnectionClass *setting_class)
 		                   G_PARAM_CONSTRUCT |
 		                   NM_SETTING_PARAM_FUZZY_IGNORE |
 		                   G_PARAM_STATIC_STRINGS));
+	/**
+	 * NMSettingConnection:mdns:
+	 *
```
thom311: If you build with --enable-gtk-doc, you'll see that this comment ends up in `./clients/common/settings-docs.c` file (this also breaks `make check`).

You need to generated file over to `./clients/common/settings-docs.c.in`, and commit the changes as well.

src/nm-policy.c
```diff
@@ -919,6 +919,49 @@ get_best_ip_config (NMPolicy *self,
...
 static void
+update_mdns(NMPolicy *self, NMDnsManager *dns_mgr, const char *ip_iface)
+{
+	guint i;
+	NMPolicyPrivate *priv = NM_POLICY_GET_PRIVATE (self);
+	gs_free NMSettingsConnection **connections = NULL;
+
+	connections = nm_settings_get_connections_sorted (priv->settings, NULL);
+
+	for (i = 0; connections[i]; i++) {
```
thom311: this is not right. You iterate over all connection-profiles, without regard whether they are currently active or not.
also, `nm_setting_connection_get_interface_name()` is not required to be present.

I would merge this commit in the previous one. But it seems to me, this needs to be re-thought.

Note that for regular DNS settings, the manager tracks it for all devices at the same time. Meaning, whenever a device's configuration changes, we propagate the change to the NMDnsManager via `nm_dns_manager_add_ip6_config()`. But note that that is not only done during `update_ip6_dns()` (which gets called when a different device gets the best default-route), it happens every time something on the device changed.
On the other hand, your `update_mdns()` iterates over all connections, setting nm_dns_manager_set_mdns() for each ifindex it finds. That seems wrong. Instead, pass the mdns configuration for every device during nm_dns_manager_add_ip6_config()

ipuustin:
@thom311: Thanks for the review! The code base is new to me, and I really appreciate the help. I did the changes you requested, but some for some things (such as the policy changes) I'm still not sure if I got your meaning right.

thom311:
Hi. I pushed a few suggestions to https://github.com/NetworkManager/NetworkManager/commits/th/pr/35 for your consideration.

The second commit is the non-trivial one. I think the logic there is not right. `update_ip_dns()` is called when the "best-device" changes. Then, in your code we would find the current best-device, and call `nm_dns_manager_set_mdns()` for its ifindex/mdns setting. I think that's wrong, because you want to set nm_dns_manager_set_mdns() for every device which is currently active. Also, when the device deactivates, you need to reset the mdns setting for the ifindex (currently the code will only configure mdns on the next best device).

It's not entirely clear to me how it should be, but what I tried to say was that NMDnsManager already tracks **for each** device some configuration. See `nm_dns_manager_add_ip_config()` and `nm_dns_manager_remove_ip_config()`. One problem is, that it tracks the configuration by ifname. That is not very suitable, because devices can be renamed. The ifindex would be the correct identifier for a device instead. Of course, you chould hack it, to use the ifname in `nm_dns_manager_set_mdns(dnsmgr, ifname, mdns)` (and in the layers below resolve the ifindex, before calling `"SetLinkMulticastDNS"`), but that seems ugly.
Also, while NMDnsManager tracks the per-device configs (IPv4 and IPv6), it's currently not very suitable to be extended for your new mdns setting. You could hack around that, by putting the mdns setting into NMIP4Config/NMIP6Config, but that seems ugly. A better solution would be to refactor NMDnsManager, to be able to track various per-device data (ifname, NMIP4Config, NMIP6Config, mdns) by ifindex. Doing that is a bit of an effort however...

@bengal, what do you think?

bengal:
I think that as a first step we could add functions modeled as the existing API:

```
 nm_dns_manager_add_mdns_config (NMDnsManager, ifname, NMSettingConnectionMdns)
 nm_dns_manager_remove_mdns_config (NMDnsManager, ifname)
```

The DNS manager would store this information and update the plugin
when we call update_dns().

Later, it would be nice to do what Thomas suggests and change the API to be more generic and use the ifindex.

ipuustin:
Ok, I refactored the connection settings to be separate from the ip settings in both dns manager and policy. @thom311, I incorporated your `fixup!` patches (those which were still relevant after the refactoring), I hope that was your intention?

thom311:
> @thom311, I incorporated your fixup! patches (those which were still relevant after the refactoring), I hope that was your intention?

Yes, it was. Thanks you!!

ipuustin:
Now the ifindex information should be carried to the DNS manager. Also updating ifindex for a given NMdevice is supported.

thom311:
> Now the ifindex information should be carried to the DNS manager. Also updating ifindex for a given 
> NMdevice is supported.

I think it goes to the right direction (thanks!!), but could we change it a bit more?

You currently track two distinct entries
  *  `NMDnsIPConfigData`, in `priv->ip_configs`, for `NMIP6Config` and `NMIP4Config`.
  * `NMDnsConnectionConfigData` in `priv->connection_configs` for mdns setting.

I already think that it's wrong that for one ifindex/ifname, we have possibly two `NMDnsIPConfigData` instances.

It think `NMDnsConnectionConfigData` and `NMDnsIPConfigData` should get merged in one, to have fields `ifname`, `ifindex`, `ip4_config`, `ip6_config`, `mdns`. Then, each ifindex could have zero or one such data. The data might have only parts of the settings present (e.g. `ip4_config` could be set, but `ip6_config` not. Or maybe `ip4_config` set but `mdns` not -- in which case, you might need two flags `mdns_value` and `mdns_present` to indicate whether the value is set at all).

I would only have one master function to set the data:

```
    typedef enum {
        NM_DNS_MANAGER_CONFIG_MODE_IGNORE,
        NM_DNS_MANAGER_CONFIG_MODE_SET,
        NM_DNS_MANAGER_CONFIG_MODE_UNSET,
    } NMDnsManagerConfigMode; 
    nm_dns_manager_config_set (NMDnsManager *dns_manager,
                               int ifindex,
                               NMDnsManagerConfigMode ifname_mode,
                               const char *ifname,
                               NMDnsManagerConfigMode mdns_mode,
                               gboolean mdns,
                               NMDnsManagerConfigMode ip4_config_mode,
                               NMIP4Config *ip4_config,
                               NMDnsManagerConfigMode ip6_config_mode,
                               NMIP4Config *ip6_config);
```
a function like `nm_dns_manager_add_connection_config` would be just a shortcut like
```
    void nm_dns_manager_add_ip_config (NMDnsManager *dns_manager, 
                                       int ifindex,
                                       NMIPConfig *ip_config)
   {
        if (NM_IS_IP4_CONFIG (ip_config)) {
            nm_dns_manager_config_set (dns_manager,
                                       ifindex,
                                       NM_DNS_MANAGER_CONFIG_MODE_IGNORE,
                                       NULL,
                                       NM_DNS_MANAGER_CONFIG_MODE_IGNORE,
                                       FALSE,
                                       NM_DNS_MANAGER_CONFIG_MODE_SET,
                                       ip_config,
                                       NM_DNS_MANAGER_CONFIG_MODE_IGNORE,
                                       NULL);
        } else {
                // SAME
        }
    }
```

I think that's how it should look like eventually. If you would like to work on that, that would be great. Otherwise, I think what you did so far could be the base for that change, and we could do the changes on top your patches.

bengal:
The branch looks fine to me now; I agree with Thomas that we can merge it now and improve it later, or we can wait for you to do the suggested changes. Thanks!

bengal:
There is one failure in 'make check', can you fix it? It's enough to add the new setting to the array in libnm-core/tests/test-general.c - test_connection_diff_a_only()

ipuustin:
I fixed the check failure and the compilation warning. I think it might be a good idea to merge now and improve later, because refactoring the API has a lot to do with project style preferences and not that much to do with mDNS itself. :-)

bengal:
LGTM. @thom311, what do you think?

ipuustin:
Rebased to fix the `libnm.ver` conflict.

thom311:
I took the patches, incorporated them in #44, and expanded on them.
I think this pull request should be closed and discussion should continue at #44.

Closing this pull request, but please reopen if need be. Thanks.

ipuustin:
Ok, thanks!

PR #36, opened 2017-11-15
use siphash24() almost everywhere [[th/nm-hash-all]]

thom311:
except in libnm-util and libnm-glib and a few parts of test-code.


bengal:
LGTM

PR #37, opened 2017-11-27
Add dhcpcanon as an option in the configuration

juga0:
Would it be possible to add this to the man?.
Thanks.

### DIFF
--- a/man/NetworkManager.conf.xml
+++ b/man/NetworkManager.conf.xml
@@ -183,13 +183,17 @@ plugins-=remove-me
         <term><varname>dhcp</varname></term>
         <listitem><para>This key sets up what DHCP client
         NetworkManager will use. Allowed values are
-        <literal>dhclient</literal>, <literal>dhcpcd</literal>, and
-        <literal>internal</literal>. The <literal>dhclient</literal>
-        and <literal>dhcpcd</literal> options require the indicated
+        <literal>dhclient</literal>, <literal>dhcpcd</literal> and
+        <literal>internal</literal>. <literal>dhcpcanon</literal>
+        (experimental) is also available when NetworkManager has
+        been compiled with "--with-dhcpcanon" option. 
+        The <literal>dhclient</literal> and <literal>dhcpcd</literal>
+        (as well as <literal>dhcpcanon</literal>)options require the indicated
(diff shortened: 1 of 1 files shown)


thom311:
I don't think we should do this (yet).

Is dhcpcanon packaged in any distribution?

I think not, those distros will all build NM without dhcpcanon support (and the manual entry is just confusing in the majority of cases).

A user who installs dhcpcanon by hand, also needs to rebuild NetworkManager with dhcpcanon support. Such a user probably wouldn't need the manual entry, but follow a Howto.

If you package NetworkManager for a particular distribution where you enable dhcpcanon, you can patch the manual page downstream.

juga0:
> Is dhcpcanon packaged in any distribution?

It's packaged for Debian [1]. Hopefully will be packaged for Fedora soon.

> A user who installs dhcpcanon by hand, also needs to rebuild NetworkManager with dhcpcanon support. Such a user probably wouldn't need the manual entry, but follow a Howto.

Well, that wouldn't be the case if NetworkManager is build with dhcpcanon support by default (currently it doesn't [2]). In any case dhcpcanon won't be used unless specified manually in the configuration.

> If you package NetworkManager for a particular distribution where you enable dhcpcanon, you can patch the manual page downstream.

That's right, i can do that for now.

Thanks. 
[1] http://packages.debian.org/sid/dhcpcanon
[2] https://github.com/NetworkManager/NetworkManager/pull/31#issuecomment-337881457

thom311:
I guess, we wouldn't mind a patch that modifies the manual page depending on configure options. We do that usually to patch the configuration default (like `<literal>&NM_CONFIG_DEFAULT_MAIN_AUTH_POLKIT_TEXT;</literal>`). But in this case, it would be an entire paragraph that needs to change depending on the configure option. That seems a bit more complicated.

If you would, please provide a patch that refactors the current manual page to make the list for the dhcp plugins. And then, a second patch that adds one list-item for dhcpanon, depending on whether it's enabled on compile time.

thom311:
I am closing this for now, because I think we don't want to merge the present patch as-is.

Please reopen if you disagree or open a new pull request for an updated patch.

Thanks!!

PR #39, opened 2017-12-11
th/libnm attribute handling

thom311:
Some fixes regarding attribute handling

Already partly discussed on IRC. As there was no conclusion there, I open a PR.

TODO: keyfile is unable to store/load address attributes.


thom311:
patches merged as 3cc6aa68655f1c07a831261a5f61cc36ed468c34

PR #40, opened 2017-12-11
th/static_asserts

thom311:
Compile NM with -Wvla and don't use variable length arrays.

It makes static-asserts against a non-compile-time constant work. Otherwise, the assertion is silently ignored.

### DIFF
--- a/libnm-core/nm-setting-bond.c
+++ b/libnm-core/nm-setting-bond.c
@@ -537,7 +537,7 @@ _nm_setting_bond_option_supported (const char *option, NMBondMode mode)
 
 	for (i = 0; i < G_N_ELEMENTS (bond_unsupp_modes); i++) {
 		if (nm_streq (option, bond_unsupp_modes[i].option))
-		    return !NM_FLAGS_HAS (bond_unsupp_modes[i].unsupp_modes, BIT (mode));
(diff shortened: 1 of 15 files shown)


bengal:
> build: add "-Wvla" to warn about uses of variable-length arrays 

> But the worst is, G_STATIC_ASSERT() is implemented by declaring an array of negative length. Usually, the checked condition should be a compile time constant, but with VLAs the compiler would not evaluate the static-assert at compile time and instead accept it silently. It's easy to mess up static asserts to wrongly have a non-constant condition, especially when the static-assert is used inside a macro.

I struggle to understand how this works; can you make an example of where G_STATIC_ASSERT is ignored due to VLAs?

thom311:
```
const int A = 1;
int a = 1;

G_STATIC_ASSERT (A != 0);
G_STATIC_ASSERT (a != 0);
```
both expressions are not static, hence they compile to nothing.

Sometimes it's less obvious that an expression is not static, especially when you use G_STATIC_ASSERT in a macro

```
#define XX(arg) \
    ({ \
        G_STATIC_ASSERT (sizeof (arg) == 5); \
        memset ((arg), 0, 5); \
    })
{
    int LEN = 3;
    char xx[LEN];
    XX (xx);
}
```

PR #41, opened 2017-12-13
th/wifi_iwd_config

thom311:
Make `wifi.backend` configurable per device.

Note https://mail.gnome.org/archives/networkmanager-list/2017-December/msg00027.html , where Andrew said:

> > Would it ever make sense that one device is managed by supplicant and
> > another by iwd? I think it would.
> 
> I'm not sure, multiple real wifi devices in one system is generally a
> rare situation and it's hard to tell what is expected from the
> software.

### DIFF
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -50,6 +50,7 @@
 %bcond_without wwan
 %bcond_without team
 %bcond_without wifi
+%bcond_with iwd
 %bcond_without ovs
 %bcond_without ppp
 %bcond_without nmtui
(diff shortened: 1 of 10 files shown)


contrib/fedora/rpm/NetworkManager.spec
```diff
@@ -221,6 +227,9 @@ Summary: Wifi plugin for NetworkManager
 Group: System Environment/Base
 Requires: %{name}%{?_isa} = %{epoch}:%{version}-%{release}
 Requires: wpa_supplicant >= 1:1.1
+%if %{with wifi}
```
balrog-kun: Should this rather be %{with iwd}?

thom311: fixed. Thanks

PR #42, opened 2017-12-15
IWD backend 8021x networks support proposal

balrog-kun:
The main idea here is to not require secrets for 8021x networks that we know have been configured with IWD already, and fail early for 8021x networks that have not been configured.

This should not conflict with the Wifi backend config option changes.


thom311:
Hi Andrew,

I merged your patches as-is (after rebase to master). All I did, was adding a trivial follow-up commit with a style fix.

See commit 60d4b370ba726287343efe66c1293b0f77afc486

Thanks!!

PR #43, opened 2017-12-20
th/setting_plugin_ifnet

thom311:
I think it's time to drop the `ifnet` settings plugin. Please review.

Note that `configure.ac` used to partly detect `--with-hostname-persist=` based on the chosen distro plugin. Please note that this might change some behavior there. Does it affect anybody in a bad way?
Probably distributions should anyway explicitly select `--with-hostname-perist`.


bengal:
The ifnet plugin is not maintained since long time, so LGTM.

If someone still configures with --enable-ifnet, maybe we should print
an error instead of succeeding without notice (at least for some time)?

thom311:
> If someone still configures with --enable-ifnet, maybe we should print
> an error instead of succeeding without notice (at least for some time)?

The same could be done for `--enable-ifcfg-suse`, but `./configure` already warns about unknown --enable switches. I think that is good enough, I really doubt somebody is using these plugins successfully.

I dislike the change in behavior how we detect the persist-hostname option, but I again think that nobody cares.

floppym:
Gentoo maintainer here: I'm fine with dropping it. We have been disabling it unconditionally since 2014.

thom311:
mreged

PR #44, opened 2017-12-20
Rework NMDnsManager and add MDNS support for systemd-resolved (obsoletes #35)

thom311:
I took the patches from https://github.com/NetworkManager/NetworkManager/pull/35 and expanded on them. The original patches are applied first, with minimal changes (only rebasing). Follow-up commits are based on them.  I propose this pull request to obsolete #35.

After rethinking, I think #35 needs to be solved a bit different, so I reworked `NMDnsManager` to track their data by ifindex (not by ifname), and reworked how the MDNS setting is passed to the DNS plugin.

Still totally untested :) Please review.


bengal:
I wonder if NMSettingConnection is the right place for a mdns
property. Currently we put all DNS-related parameters in ipv4 and ipv6
settings, but we discussed the idea of moving them to a dedicated
'dns' setting because the parameters are not
IP-protocol-specific.

Maybe we could add now the new dns setting with the only property 'mdns'
and later add the other properties (servers, searches, options,
priority). Or we can put 'mdns' in the connection setting for now and
move it later, before 1.12 is released.

I have quickly looked at commits and the approach looks good to me. Can
you schedule a CI run to see if it breaks anything?

thom311:
after rework and CI run, merged to master.

PR #46, opened 2017-12-29
remove --run-from-build-dir option from NetworkManager

pavlix:
It used to be possible to successfully run NetworkManager from the build
directory at least using `--run-from-build-dir`. The code is relatively
complicated for that simple task and doesn't even achieve the results.

For example the path to device plugins is hardcoded and the actual device
plugins are in their own different build directories. There is no trivial
way to fix that.

In future we should probably implement running from build directory using a
set of environment variables with search paths and other tweaks together
with `NetworkManager` and `nmcli` shell wrappers that would make it all
work. In addition to plugin and helper search paths there appears to be a
problem with the dbus and polkit configurations in the running system. This
seems to be a drawback related to the usage of D-Bus for all communication.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -71,7 +71,6 @@ static struct {
 	gboolean print_config;
 	gboolean become_daemon;
 	gboolean g_fatal_warnings;
-	gboolean run_from_build_dir;
 	char *opt_log_level;
 	char *opt_log_domains;
 	char *pidfile;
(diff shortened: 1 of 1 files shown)


thom311:
Hi Pavel,

yesterday on IRC we discussed about dropping this option.

I opinionated that to my knowledge nobody uses it. Given that it's probably(?) quite broken, I agree it should be fixed or dropped. 

I am also not sure that I would make use of it, even if it would be fixed. So, to me personally, this option isn't useful. But maybe I could change my mind. Also, if it's useful for somebody else, that is fine with me.

If I understood you correctly, you intend to fix it instead.

Waiting on an update from you :)

bengal:
Since the lookup of plugins is broken in --run-from-build-dir mode and thus the daemon is going to behave in a different way when launched in such way, I don't think that the option is particularly useful. I'm fine with dropping it.

pavlix:
This became part of https://github.com/NetworkManager/NetworkManager/pull/52 and is therefore no longer needed.

PR #47, opened 2017-12-29
dhcp: remove dhcpcd plugin

pavlix:
The dhcpcd plugin doesn't support IPv6 and there is no good reason to use
it with NetworkManager. The power of dhcpcd is when it is used as a
standalone IPv4/IPv6 configuration daemon rather than a DHCP helper for one
of the big network configuration services. The default DHCP plugin for
NetworkManager is now dhclient even on Gentoo and we are eventually going
to switch away from that as well.

Removing dhcpcd plugin ensures that we don't have additional maintainance
burden with a plugin that isn't even fully working and nobody was willing
to fix it for many years.


thom311:
I also want get rid of code, especially if badly maintained, little used, or not well functioning.
I am all for reducing the feature matrix.
Better one awesome way to do it, instead of several ones with issues.

But I also hate it when a project deprecates something which works for me, and with that excuse drop it in the next release entirely. It's just so annoying to break working user setups (I have projects in mind).

I have no idea who is using dhcpcd and how well it works for those users. But if in doubt, I'd rather not drop it. The dhcpcd code is there, but it doesn't really bother me too much.

bengal:
I agree with Thomas, the maintenance burden of the plugin is negligible and removing it has the potential of breaking users for whom the plugins works well now. Instead, if we plan to get rid of it in the future, can we start by marking it as deprecated in the man pages?

thom311:
Ok, I think after discussion, we agreed not to do this.

My main argument for not dropping the plug-in now, is that it really doesn't cause much additional ugliness. It's an independent piece of code.

Sure, we have the entire DHCP plug-in-infrastructure, which indeed is a maintenance burder. Dropping that altogether would be a simplification. But we need this infrastructure anyway for `dhclient` and `internal` plug-ins. So, as we have it, the `dhcpcd` plug-in itself is not much additional trouble.

As also discussed, a patch for the manual page to discourage the use of `dhcpcd` would be appreciated.

Closing for now, maybe do this in the future. Only after having the `dhcpcd` plug-in documented as deprecated for a **long** time.

PR #49, opened 2017-12-29
RFC: rename ifcfg-rh to ifcfg now that there is just one variant

pavlix:


thom311:
I cherry-picked the two commits that adjust the auto-detection of the ifcfg plugin.
Merged to master early as 6341b6a5daff3e6506207e046d9b3be0eaec8593 and 1402fa7487b29fc1ea39a6bf7659fee7f30bb0e0.


Regarding renaming. While I agree that `ifcfg` would be a better name, the old name is well established. E.g. you would at least need something like

```
   if (nm_streq (pname, "ifcfg-rh"))
       pname = "ifcfg";
```
in [load_plugins()](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/settings/nm-settings.c?id=1402fa7487b29fc1ea39a6bf7659fee7f30bb0e0#n690) to not break existing settings in `NetworkManager.conf`.

Also, probably the man page should mention the deprecated alias.

Also, we install `man nm-settings-ifcfg-rh`. With the change, we would need either to symlink/hardlink it with a new `man nm-settings-ifcfg` or make it a stub that tells the user that this manual page is a deprecated alias and where to look at.


Renaming stuff without breaking existing configurations means to indefinitely support the old name. It ends up being more ugly then the misnomer in the first place. I am not convinced about such a change.

Conan-Kudo:
@thom311 I disagree with the premise that the old names have to stick around forever.

Provided that there's some notice, packagers and users can be prepared for a change.

bengal:
> Provided that there's some notice, packagers and users can be prepared for a change.

Packagers hopefully read release notes when a new version is available; but how do you prepare users?

As a user, I find annoying that I have to update configuration files after upgrading packages (fortunately it happened only very few times).

Conan-Kudo:
@bengal At least for one version series where the change occurs, supporting the old name as an alias is fine, but NetworkManager could rewrite its configuration when it detects legacy keys and change them. Alternatively, it could print out a message to the syslog warning of the deprecated entry.

thom311:
> @thom311 I disagree with the premise that the old names have to stick around forever.

The patch doesn't first deprecated the current name to drop it in the future. It drops it right away and there is no grace period for which both names are accepted.
Even then, (in my opinion) taking a deprecation process as an excuse to break existing configurations it a bad attitude. A user doesn't really care why he has to invest work to adjust his perfectly working setup.

Luckily, in this case, it's trivial for NM to keep supporting the old name indefinitely. See the 3 points in https://github.com/NetworkManager/NetworkManager/pull/49#issuecomment-355026019 

I personally don't care about the name strongly enough, to implement the (albeit simple) workaround to preserve the old name. But if the branch implements the workarounds to keep the old name functioning, it's fine with me.

Another thing is, to take care that renaming the man page `nm-settings-ifcfg-rh` doesn't break the URL https://developer.gnome.org/NetworkManager/stable/nm-settings-ifcfg-rh.html

bengal:
> @bengal At least for one version series where the change occurs, supporting the old name as an alias is fine, but NetworkManager could rewrite its configuration when it detects legacy keys and change them. 

Rewriting the configuration may not work well is some cases and requires non-trivial implementation effort (I imagine) ... I don't know if it's worth it.

> Alternatively, it could print out a message to the syslog warning of the deprecated entry.

Yeah, but usually people don't look at logs when things work...

I agree with Thomas that, if we want to do the name change, it's easier to support both the old and the new name.

vhumpa:

thom311:
As discussed, this change needs some work. I am closing this pull request for now. Please re-open, if there is something new. Thanks you!

PR #50, opened 2017-12-29
th/c_list_sort_nonrec

thom311:
Replace the recursive implementation for `c_list_sort()` with an non-recursive to avoid the overhead of the function call.


bengal:
LGTM

thom311:
merged

PR #51, opened 2018-01-03
th/nm_compilation_type

thom311:
We used already `NETWORMANAGER_COMPILATION` define so that (mainly) `nm-default.h` and `nm-test-utils.h` would behave accordingly for the component in which they were included. However, the distinction we not very clear, e.g. we would have build `src/dhcp/nm-dhcp-helper` with `NM_NETWORKMANAGER_COMPILATION_DEFAULT`, which would include `<glib/gi18n.h>`.

In fact, `nm-dhcp-helper` doesn't need translation, hence it should not include either translation headers.

Make the compilation type more granular and specify it everywhere correctly.


shared/nm-default.h
```diff
@@ -22,19 +22,95 @@
...
 #define __NM_DEFAULT_H__
 
-/* makefiles define NETWORKMANAGER_COMPILATION for compiling NetworkManager.
- * Depending on which parts are compiled, different values are set. */
-#define NM_NETWORKMANAGER_COMPILATION_DEFAULT             0x0001
-#define NM_NETWORKMANAGER_COMPILATION_INSIDE_DAEMON       0x0002
-#define NM_NETWORKMANAGER_COMPILATION_LIB                 0x0004
-#define NM_NETWORKMANAGER_COMPILATION_SYSTEMD             0x0008
-#define NM_NETWORKMANAGER_COMPILATION_LIB_LEGACY          0x0010
+#define NM_NETWORKMANAGER_COMPILATION_WITH_GLIB                 (1 <<  0)
```
bengal: It seems to me that 

```
NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_PRIVATE
NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_CORE_INTERNAL
NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_CORE_PRIVATE
NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_UTIL_PRIVATE
NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_GLIB_PRIVATE
NM_NETWORKMANAGER_COMPILATION_WITH_SYSTEMD
```

are not used anywhere; can they be dropped?

shared/nm-utils/nm-test-utils.h
```diff
@@ -670,6 +670,20 @@ nmtst_test_quick (void)
...
 
+#define NMTST_EXPECT(domain, level, msg)        g_test_expect_message (domain, level, msg)
+
+#if (NETWORKMANAGER_COMPILATION) & NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_UTIL
+#define NMTST_EXPECT_LIBNM_U(level, msg)        NMTST_EXPECT ("libnm-util", level, msg)
+#define NMTST_EXPECT_LIBNM_G(level, msg)        NMTST_EXPECT ("libnm-glib", level, msg)
+
+#define NMTST_EXPECT_LIBNM_U_CRITICAL(msg)      NMTST_EXPECT_LIBNM_U (G_LOG_LEVEL_CRITICAL, msg)
+#define NMTST_EXPECT_LIBNM_G_CRITICAL(msg)      NMTST_EXPECT_LIBNM_G (G_LOG_LEVEL_CRITICAL, msg)
+#else
```
bengal: No strong opinion, but I don't find those macros very useful; they are few characters shorter, but the name encodes the same information as the original function, with the disadvantage that you have to look up what they do if you don't know it.

thom311: I mostly find the `NMTST_EXPECT_NM_*` macros useful, because they also check for the right logging level, so we get
```
-    g_test_expect_message ("NetworkManager", G_LOG_LEVEL_MESSAGE,
-                           "*<warn> * keyfile: 802-1x.client-cert: certificate or key file '/CASA/dcbw/Desktop/certinfra/client.pem' does not exist*");
+    NMTST_EXPECT_NM_WARN ("keyfile: 802-1x.client-cert: certificate or key file '/CASA/dcbw/Desktop/certinfra/client.pem' does not exist*");
```
for free. It's shorter, and tests more specific.

The `NMTST_EXPECT_LIBNM_CRITICAL` variant exists for completeness and consistency. If you already understand what `NMTST_EXPECT_NM_WARN` does, the libnm version should look familiar. What it provides, is that it encapsulates the `"libnm"` domain, so vim autosuggests the function with `NMTST_E<C-n>`, instead of `g_test_e<C-n> ("libnm", G_LOG_L<C-n>, `.

shared/nm-default.h
thom311: `NM_NETWORKMANAGER_COMPILATION_WITH_SYSTEMD` is actually needed. That was a bug.

Regarding the others: maybe. The `*_PRIVATE` flag indicates that you are allowed to make use of private API -- that is, you are building the library itself. Although unused, I added them for completeness, hoping that the code shows the intend of how it should/could be used (instead of some random comment that gets ignored). To me, it should be a very different thing whether you build libnm-core.a or libnm.so. The former can access `libnm-core/nm-utils-private.h`, the latter not. Since those private headers are in the same path as the public headers, we only don't include them by convention -- which is easy to get wrong. One idea was to do:
```
    #if !((NETWORKMANAGER_COMPILATION) & NM_NETWORKMANAGER_COMPILATION_WITH_LIBNM_PRIVATE)
    #error Cannot use private header
    #endif
```
in those headers.

But maybe such headers should instead move to separate directories, like
```
   libnm-core/private/nm-utils-private.h
   libnm-core/internal/nm-core-internal.h
```
and when building libnm-core.a, you need to explicitly specify those include paths. Downside is, that our list of `-I` arguments is already long, so I am not convinced that more directories are better.


Ok, I now dropped the libnm-util and libnm-glib releated define, and made use of all others. If course, if you don't agree that they are used in a useful manner, you could argue that it's an invalid excuse for having them :)

bengal: > Ok, I now dropped the libnm-util and libnm-glib releated define, and made use of all others. If course, if you don't agree that they are used in a useful manner, you could argue that it's an invalid excuse for having them :)

Looks good.

shared/nm-utils/nm-test-utils.h
bengal: > I mostly find the NMTST_EXPECT_NM_* macros useful, because they also check for the right logging level, so we get
> 
> -    g_test_expect_message ("NetworkManager", G_LOG_LEVEL_MESSAGE,
> -                           "*<warn> * keyfile: 802-1x.client-cert: certificate or key file '/CASA/dcbw/Desktop/certinfra/client.pem' does not exist*");
> +    NMTST_EXPECT_NM_WARN ("keyfile: 802-1x.client-cert: certificate or key file '/CASA/dcbw/Desktop/certinfra/client.pem' does not exist*");
> 
> for free. It's shorter, and tests more specific.

Okay, it's fine with me.

thom311:
Merged.

PR #52, opened 2018-01-03
development: make it possible to run NetworkManager from build directory

pavlix:
I know we were talking about keeping the `NetworkManager --run-from-build-dir` option but right now it feels much easier to just use `./autogen.sh --enable-develop` and have the feature cleanly applied just to the development build and nowhere else.

This way we can try to keep devel builds safe from loading system NetworkManager libraries and plugins and at the same time the system build doesn't get polluted by code that only applies to the build directory.

I propose to merge this pull request now because it's extremely useful to have the feature in master and thus be able to build branches in develop mode immediately. We can reconsider the approach later if needed.


thom311:
I merged two patches from the branch to master (96658d2fd11410b01d11ace62a89899dd7958ee4).
It needed some trivial adjustments.

There are two patches left:

- remove --run-from-build-dir option from NetworkManager
- development: make it possible to run NetworkManager from build directory

I am not convinced about a compile time option. Changing a configure option is cumbersome, because it requires rebuild most of the code. Also, you effectively cannot `make install` such a build. How about:

 - keep `--run-from-build-dir` option. All the decisions that are done with `#if DEVELOP` can also be made at runtime, with a global variable `nm_global_option_run_from_build_dir`.
 - additionally, only enable `--run-from-build-dir` and `nm_global_option_run_from_build_dir` when compiling `--with-more-asserts`. That means, guard it with `#if NM_MORE_ASSERTS`. I think it makes sense that if you don't do a devel build, you can also not run from the build directory. That especially applies when the NM binary is installed. At that point, the option makes no more sense and will not be present in release builds (`--without-more-asserts`).

vhumpa:

thom311:
As I said, I think the remaining two patches need some rework.

I am closing this pull request for now, please re-open, if there is anything new. Thank you!

PR #53, opened 2018-01-04
IWD backend improvements

balrog-kun:


src/devices/wifi/nm-device-iwd.c
```diff
@@ -809,6 +814,27 @@ check_scanning_prohibited (NMDeviceIwd *self, gboolean periodic)
 	return prohibited;
 }
 
+static void
+scan_cb (GObject *source, GAsyncResult *res, gpointer user_data)
+{
+	NMDeviceIwd *self = user_data;
+	NMDeviceIwdPrivate *priv = NM_DEVICE_IWD_GET_PRIVATE (self);
```
thom311: Below, we pass `priv->cancellable` to `g_dbus_proxy_call()`. That is good, usually, either an async request must be cancellable, or we must take an additional reference to keep the object alive. Taking an additional reference is usually bad, because it means the object cannot be destroyed, until the callback returns. So, cancel is good.

But, in the callback you must *first* check whether 
```
    if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
```
before accessing `self`.

src/devices/wifi/nm-device-iwd.c
```diff
@@ -858,8 +884,14 @@ dbus_request_scan_cb (NMDevice *device,
...
 	}
 
-	g_dbus_proxy_call (priv->dbus_proxy, "Scan", g_variant_new ("()"),
-	                   G_DBUS_CALL_FLAGS_NONE, -1, NULL, NULL, NULL);
+	if (!priv->scanning && !priv->scan_requested) {
+		g_dbus_proxy_call (priv->dbus_proxy, "Scan",
+		                   g_variant_new ("()"),
+		                   G_DBUS_CALL_FLAGS_NONE, -1,
+		                   priv->cancellable, scan_cb, self);
+		priv->scan_requested = TRUE;
```
thom311: maybe instead of having `priv->scan_requested`, use a separate `priv->cancellable_scan` instance? It feels confusing to me to reuse the same cancellable for multiple tasks.

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1443,6 +1475,47 @@ activation_failure_handler (NMDevice *device)
...
+	if (current_state <= NM_DEVICE_STATE_UNAVAILABLE)
+		return;
+
+	if (current_state == NM_DEVICE_STATE_DISCONNECTED)
+		interval = 10;
+	else
+		interval = 20;
+
+	if (priv->periodic_scan_id)
+		g_source_remove (priv->periodic_scan_id);
```
thom311: `nm_clear_g_source (&priv->periodic_scan_id);` ?

src/devices/wifi/nm-device-iwd.c
balrog-kun: `priv->scan_requested` is used so that we avoid queuing a second scan even through `priv->scanning` is still FALSE while we're waiting for the scan to start.  So I don't think those two things are related.
I can add a `priv->cancellable_scan` but being new to glib I thought that was the whole benefit of using GCancellables.  For things like device shutdown where everything needs to be cancelled you don't have to keep one variable for each single task that the device may be doing, it seemed like it was made for this purpose.  The doc only says that it can be done but doesn't say if it's recommended or not.

src/devices/wifi/nm-device-iwd.c
balrog-kun: Ok, can do that although we're assigning `priv->periodic_scan_id` soon anyway.

src/devices/wifi/nm-device-iwd.c
balrog-kun: Ok, good point.  I'm new to GCancellables.

thom311:
It seems to me, that this pull request is at least partly obsolete by #60.

I am closing this pull request for now, please reopen (or maybe just create a new pull request) if these patches are consolidated with #60.

Thank you!

PR #54, opened 2018-01-10
th/build_meson_on_travis

thom311:
Some fixes for meson build, and enable meson build on travis.

Please review

### DIFF
--- a/libnm/meson.build
+++ b/libnm/meson.build
@@ -2,13 +2,7 @@ libnm_inc = include_directories('.')
 
 sources = files('nm-libnm-utils.c')
 
-if have_fake_typelibs
-  subdir('fake-typelib')
-endif
-
 deps = [
-  # FIXME: this makes GIR mandatory?
-  gir_dep,
   libnmdbus_dep,
   shared_dep
 ]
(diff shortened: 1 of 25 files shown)


thom311:
various options are still disabled in meson build (in fact, even with autotools, we don't build everything in our CI).

Of course, we should test/build as much as possible... TODO laster.

bengal:
LGTM

meson_options.txt
```diff
@@ -44,7 +44,7 @@ option('bluez5_dun', type: 'boolean', value: false, description: 'enable Bluez5
 
 # configuration plugins
 option('config_plugins_default', type: 'string', value: '', description: 'Default configuration option for main.plugins setting, used as fallback if the configuration option is unset')
-option('config_plugin_ibft', type: 'boolean', value: true, description: 'enable ibft configuration plugin')
+option('ibft',     type: 'boolean', value: false, description: 'enable ibft configuration plugin')
```
inigomartinez: There are a couple whitespaces in front of type that would be nice to remove.

libnm/meson.build
```diff
@@ -19,6 +13,11 @@ cflags = [
   '-DNMRUNDIR="@0@"'.format(nm_pkgrundir)
 ]
 
+if have_fake_typelibs
+  deps += [ gir_dep ]
```
inigomartinez: Square brackets are not necessary when adding a dependency to a list.

meson_options.txt
thom311: the white space is intentional, to vertically align with the neighbouring lines, that do something similar.

libnm/meson.build
thom311: fixed

thom311:
> Why not create each linker script strings in the main meson.build file and reuse them when needed?

done. See commit 34cb6f9877a9b7a7032899dbccca1def88da4191

libnm/meson.build
thom311: fixed by commit 98b46537fa1a91941b5f7a8252b3fce85554a855

PR #55, opened 2018-01-11
th/modem_ip_iface

thom311:
Rework handling of ip-iface vs. data-port in NMModem and related.

The path tries to do two things (for NMModem):

- there should always be a clear distinction between an networking interface (ip-iface) and other identifiers, like the data port for ppp.

- rely less on interface names, but try to resolve the ifindex as early as possible. The ifindex can be used as an ID for a network link, the ifname cannot, because interfaces can be renamed. Sometimes, we get only the name from external API (like ModemManager or PPP). In this case, we should resolve the ifindex as early as possible, to minimize the time for renaming-race, and then proceed with the ifindex.

NMDevice currently has `nm_device_get_ifindex()` and `nm_device_get_ip_ifindex()`. Eventually, this should be separated, so that we don't track a control-device and a data-device in one. This branch is a small step in that direction.

Please review.

### DIFF
--- a/src/devices/wwan/nm-device-modem.c
+++ b/src/devices/wwan/nm-device-modem.c
@@ -263,17 +263,17 @@ static void
 data_port_changed_cb (NMModem *modem, GParamSpec *pspec, gpointer user_data)
 {
 	NMDevice *self = NM_DEVICE (user_data);
-	gboolean changed;
+	gboolean has_ifindex;
 
 	/* We set the IP iface in the device as soon as we know it, so that we
 	 * properly ifup it if needed */
-	changed = nm_device_set_ip_iface (self, nm_modem_get_data_port (modem));
+	has_ifindex = nm_device_set_ip_iface (self, nm_modem_get_data_port (modem));
 
 	/* Disable IPv6 immediately on the interface since NM handles IPv6
 	 * internally, and leaving it enabled could allow the kernel's IPv6
 	 * RA handling code to run before NM is ready.
 	 */
-	if (changed)
+	if (has_ifindex)
 		nm_device_ipv6_sysctl_set (self, "disable_ipv6", "1");
 }
 
--- a/src/devices/wwan/nm-device-modem.c
+++ b/src/devices/wwan/nm-device-modem.c
@@ -629,11 +629,7 @@ set_modem (NMDeviceModem *self, NMModem *modem)
 	g_signal_connect (modem, NM_MODEM_STATE_CHANGED, G_CALLBACK (modem_state_cb), self);
 	g_signal_connect (modem, NM_MODEM_REMOVED, G_CALLBACK (modem_removed_cb), self);
 
-	/* In the old ModemManager the data port is known from the very beginning;
-	 * while in the new ModemManager the data port is set afterwards when the bearer gets
-	 * created */
 	g_signal_connect (modem, "notify::" NM_MODEM_DATA_PORT, G_CALLBACK (data_port_changed_cb), self);
-
 	g_signal_connect (modem, "notify::" NM_MODEM_DEVICE_ID, G_CALLBACK (ids_changed_cb), self);
 	g_signal_connect (modem, "notify::" NM_MODEM_SIM_ID, G_CALLBACK (ids_changed_cb), self);
 	g_signal_connect (modem, "notify::" NM_MODEM_SIM_OPERATOR_ID, G_CALLBACK (ids_changed_cb), self);
@@ -706,9 +702,9 @@ nm_device_modem_init (NMDeviceModem *self)
 NMDevice *
 nm_device_modem_new (NMModem *modem)
 {
+	NMDevice *self;
 	NMDeviceModemCapabilities caps = NM_DEVICE_MODEM_CAPABILITY_NONE;
 	NMDeviceModemCapabilities current_caps = NM_DEVICE_MODEM_CAPABILITY_NONE;
-	NMDevice *device;
 	const char *data_port;
 
 	g_return_val_if_fail (NM_IS_MODEM (modem), NULL);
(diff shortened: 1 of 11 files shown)


src/devices/wwan/nm-device-modem.c
```diff
@@ -708,34 +708,23 @@ nm_device_modem_new (NMModem *modem)
...
-	                                    NM_DEVICE_MODEM_CAPABILITIES, caps,
-	                                    NM_DEVICE_MODEM_CURRENT_CAPABILITIES, current_caps,
-	                                    NULL);
-
-	/* If the data port is known, set it as the IP interface immediately */
-	data_port = nm_modem_get_data_port (modem);
-	if (data_port) {
-		nm_device_set_ip_iface (device, data_port);
-		nm_device_ipv6_sysctl_set (device, "disable_ipv6", "1");
-	}
```
bengal: I don't understand, why is this no longer needed?

thom311: My theory is, that nobody sets data-port when constructing the object. Hence, immediately after construction of the modem, the data port cannot be set. See commit ceb5c569921eb1f0b4ab66b7391e80721c7092d0.

thom311: I  addressed this concern by a follow-up commit "device: don't set NMDeviceModem's ip-iface right after constuction"

src/devices/wwan/nm-device-modem.c
```diff
@@ -743,9 +743,10 @@ dispose (GObject *object)
 {
 	NMDeviceModemPrivate *priv = NM_DEVICE_MODEM_GET_PRIVATE ((NMDeviceModem *) object);
 
-	if (priv->modem)
+	if (priv->modem) {
 		g_signal_handlers_disconnect_by_data (priv->modem, NM_DEVICE_MODEM (object));
-	g_clear_object (&priv->modem);
+		g_clear_object (&priv->modem);
```
dcbw: Don't need the if (priv->modem) anymore, do we?

src/devices/wwan/nm-modem.c
```diff
@@ -1395,6 +1405,38 @@ nm_modem_get_data_port (NMModem *self)
...
+_set_ip_ifindex (NMModem *self, int ifindex)
+{
+	NMModemPrivate *priv = NM_MODEM_GET_PRIVATE (self);
+
+	nm_assert (ifindex >= -1);
+
+	if (priv->ip_ifindex != ifindex) {
+		gboolean notify;
+
+		notify = priv->ip_ifindex > 0;
```
dcbw: This notify behavior causes crashes when the IPv4 config comes in from the ppp-manager.  This is because the NMDevice itself is "ttyUSB0" and has no ifindex yet (because it's not a netdev).  It also has no ip_ifindex yet for the same reason.  So when the ppp-manager signals NM_PPP_MANAGER_SIGNAL_IFINDEX_SET with a valid index, in this function priv->ip_ifindex == -1, and no notification is sent.  That means the parent NMDevice never gets its ip_ifindex set.

And that makes ip4_config_merge_and_apply() create a new composite NMIP4Config without an ifindex in _ip4_config_new(), and then when we go to merge addresses to this new composite config from the WWAN one, _nm_ip_config_add_obj() hits nm_assert (ifindex > 0); and we crash.

What was the motivation for gating the notification on > 0 here?

dcbw:
Besides the crashes, seems OK to me :)

src/devices/wwan/nm-modem.c
```diff
@@ -1438,6 +1435,91 @@ _set_ip_ifindex (NMModem *self, int ifindex)
...
+	           || !NM_IN_SET (ip6_method, NM_MODEM_IP_METHOD_UNKNOWN, NM_MODEM_IP_METHOD_STATIC, NM_MODEM_IP_METHOD_AUTO)
+	           || (   ip4_method == NM_MODEM_IP_METHOD_UNKNOWN
+	               && ip6_method == NM_MODEM_IP_METHOD_UNKNOWN)) {
+		g_set_error_literal (error, NM_UTILS_ERROR, NM_UTILS_ERROR_UNKNOWN,
+		                     "invalid ip methods");
+		return FALSE;
+	}
+
+	if (is_ppp) {
+		if (data_port[0] != '/') {
```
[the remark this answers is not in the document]
thom311: Ok, but then `port_speed_is_zero()` is wrong, because the cwd of  NM is undefined and it cannot call `open()` with a relative path. How do you recommend to fix this?

src/devices/wwan/nm-modem.c
thom311: The idea was, that `nm_modem_get_ip_ifindex()` always returns a value >= 0. So, when setting ip-ifindex from -1 to 0, the value doesn't appear to change from outside. Such a no-change notification should be suppressed. The condition was of course wrong, and maybe the whole intent is wrong. Added a fixup to always emit the signal.

However, odd that an ignored NM_PPP_MANAGER_SIGNAL_IFINDEX_SET signal results in an assertion failure. Need to check whether that hints to another bug... (todo)

src/devices/wwan/nm-modem.c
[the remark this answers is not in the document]
dcbw: @thom311 it'll be a TTY, so I guess we can just prepend "/sys/class/tty/" to the device name and just treat ENOENT as "port speed is zero" in port_speed_is_zero().

src/devices/wwan/nm-modem.c
dcbw: > However, odd that an ignored NM_PPP_MANAGER_SIGNAL_IFINDEX_SET signal results in an assertion failure. Need to check whether that hints to another bug... (todo)

@thom311 if there's a bug, it would be in doing any IPv4/IPv6 operations on a device that doesn't yet have an ip_ifindex.  What I had always wanted to do but never got around to was the parent/child stuff, where the ttyUSB0 NMDevice wouldn't do anything IP-related itself, but it's child ppp0 device would, and then perhaps (for compatibility) the IP4Config from the ppp0 device would be copied to ttyUSB0 only for D-Bus compat.  Same for VPNs, PPPoE, etc.  That would have removed the whole ip_iface thing since the only IP-related operations would be performed on actual netdevices.

Obviously that would also have to be paired with API changes to expose those parent/child relationships, and not all parents would get the IPConfig copied (eg WWAN would for ppp0->ttyUSB0 but bridges wouldn't necessarily for eth0->br0).  So maybe that's confusing.  But where we have split control + data interfaces (WWAN, Bluetooth, ATM/ADSL, PPPoE, etc) it might make things simpler.  Not sure...

src/devices/wwan/nm-modem.c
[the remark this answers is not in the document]
dcbw: @thom311 the path will either be (a) absolute already in which case we can just use it, or (b) not absolute but including a directory separator in which case we should error out (this would be very unexpected), or (c) bare device name eg ttyUSB0 in which case prepend and open.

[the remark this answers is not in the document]
thom311: addressed with a fixup

src/devices/wwan/nm-device-modem.c
aleksander0m: data-port was known at construct time in MM 0.6 and earlier; since MM 1.x data port is known once the modem has been connected

vhumpa:
rebuild

vhumpa:
rebuild

thom311:
merged as 0b25e302382e2ec5d68474f14e2e0545de19b2d7

PR #56, opened 2018-01-12
th/libnm_jansson

thom311:
A different approach to wrapping the libjansson symbols.

See commit cd476e4dc922f0acfd65b02639032ee67339ad95.

Preferably, we don't use any `json_*` names, and only include `<jansson.h>` in the glue code. However, defines like `json_int_t` (which depends on configure options of libjansson) and `json_error_t` (which are supposed to be stack-allocated by us), are hard to re-implement.

So, we still include `<jansson.h>` header and make use of such `json*` typedefs. However, don't use any symbols from the ABI.

cd476e4dc922f0acfd65b02639032ee67339ad95 achieved that by re-defining symbols before including the header. I think that is confusing. How about a function table instead?

### DIFF
--- a/libnm-core/nm-json.c
+++ b/libnm-core/nm-json.c
@@ -18,99 +18,134 @@
 
 #include "nm-default.h"
 
-#define NM_JANSSON_C
 #include "nm-json.h"
 
 #include <dlfcn.h>
 
-void *_nm_jansson_json_object_iter_value;
-void *_nm_jansson_json_object_key_to_iter;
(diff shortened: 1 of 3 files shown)


thom311:
I merged the first (uncontroversial?) part. Rest rebased.

thom311:
Ok, then close.

thom311:
this branch was heavilty reworked, and again proposed for merging: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/559

PR #58, opened 2018-01-15
th/wifi_cleanup

thom311:
Some cleanup

also, move the netlink code to a new `nm-netlink.h` header.

We will need generic netlink for wireguard.

Please review.


bengal:
Looks good.

thom311:
merged

PR #60, opened 2018-01-15
IWD state change handling

balrog-kun:
This pull request contains part of https://github.com/NetworkManager/NetworkManager/pull/53 except the periodic scan addition which need fixing.


thom311:
the patches lgtm.

Regarding the last RFC, it seems to me that the best way for a subclass to prevent auto-activation (based on some internal state) is indeed

```
    parent_class->get_autoconnect_allowed = get_autoconnect_allowed;
```
So, also that patch lgtm.

If you could reword the commit message and rebase to master, it's good to merge IMO. Thanks

balrog-kun:
Ok, rebased the changes but somehow github shows as if I'd removed the original commits from the pull request and added new ones... not sure if this is normal.

thom311:
Merged as 14cfb3d3735227b0f3b2ead52d449ecef2252ef4

Thanks!

PR #62, opened 2018-01-22
More IWD fixes

balrog-kun:
With this I seem to be able to use NM+iwd for my daily desktop work including VPN connections.

### DIFF
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -75,10 +75,12 @@ typedef struct {
 	GCancellable *  cancellable;
 	NMDeviceWifiCapabilities capabilities;
 	NMActRequestGetSecretsCallId *wifi_secrets_id;
+	guint           periodic_scan_id;
 	bool            enabled:1;
 	bool            can_scan:1;
 	bool            can_connect:1;
 	bool            scanning:1;
+	bool            scan_requested:1;
 } NMDeviceIwdPrivate;
 
 struct _NMDeviceIwd {
@@ -100,6 +102,9 @@ G_DEFINE_TYPE (NMDeviceIwd, nm_device_iwd, NM_TYPE_DEVICE)
 #define NM_DEVICE_IWD_GET_PRIVATE(self) _NM_GET_PRIVATE(self, NMDeviceIwd, NM_IS_DEVICE_IWD)
 
 /*****************************************************************************/
+
+static void schedule_periodic_scan (NMDeviceIwd *self,
+                                    NMDeviceState current_state);
 
 static void
 _ap_dump (NMDeviceIwd *self,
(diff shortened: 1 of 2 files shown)


src/devices/wifi/nm-device-iwd.c
```diff
@@ -864,6 +869,30 @@ check_scanning_prohibited (NMDeviceIwd *self, gboolean periodic)
...
+scan_cb (GObject *source, GAsyncResult *res, gpointer user_data)
+{
+	NMDeviceIwd *self = user_data;
+	NMDeviceIwdPrivate *priv = NM_DEVICE_IWD_GET_PRIVATE (self);
+	NMDeviceState state = nm_device_get_state (NM_DEVICE (self));
+	gs_free_error GError *error = NULL;
+
+	if (   !_nm_dbus_proxy_call_finish (G_DBUS_PROXY (source), res,
+	                                   G_VARIANT_TYPE ("()"), &error)
+	    && g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
```
thom311: `scan_cb` must not dereference `scan` before checking that the request was not cancelled. Otherwise, `self` might be a dangling pointer, and `NM_DEVICE_IWD_GET_PRIVATE (self);` and `nm_device_get_state (NM_DEVICE (self));` will lead to crash.

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -648,9 +649,14 @@ static void
...
 {
 	NMIwdManagerPrivate *priv = NM_IWD_MANAGER_GET_PRIVATE (self);
+	NMManager *manager = nm_manager_get ();
 
 	priv->cancellable = g_cancellable_new ();
+
 	prepare_object_manager (self);
+
+	g_signal_connect (manager, "device-added",
+	                  G_CALLBACK (device_added), self);
```
thom311: The later commit "iwd: disconnect from NMManager signals in cleanup" should be squashed into this one.

When you connect to a signal, you commonly have to disconnect too. If you are going to disconnect later, you must be sure that the object is still alive. `nm_manager_get()` is a singleton, but during shutdown singletons are deconstructed (well, currently we leak a lot of singletons, that is a bug). If a singleton uses another singleton just for the moment, it might not need to take a reference. But if you connect and disconnect to a signal that implies that the source singleton needs to stay alive during that time.

I guess, you could use `g_signal_connect_object()`, or maybe better: add a `NMIwdManagerPrivate.manager` field that keeps a reference to the manager instance. Initialize it in `nm_iwd_manager_init()`, and use it instead of `nm_manager_get()`.

src/devices/wifi/nm-device-iwd.c
balrog-kun: Oops

src/devices/wifi/nm-iwd-manager.c
balrog-kun: Good point, let's keep a reference to NMManager. 
I think this is a fix unrelated from the gdbus workaround patch so let me just reorder the patches instead of squashing them.

vhumpa:

thom311:
Looks good to me. Sorry for the long delay.

Rebased on master (without conflicts) and merged as 873be8a37a31411a6b655b6fe35b4fc46158bcf7

Thanks!

PR #63, opened 2018-01-31
th/ifcfg_simple_dbus

thom311:


vhumpa:
Heya Thomas, please disregard the 'default' check result, I've just created a trigger for pull requests here and it isn't yet finished coding. Disabled that until done. It's good to see the triggering works though.

vhumpa:
rebuild

vhumpa:
rebuild

vhumpa:
rebuild

thom311:
arguably, I ended up not using `NM_UNCONST_PTR()` and `NM_UNCONST_PPTR()`, and they are only in an intermediate step. That was, because originally I wanted to do it that way, but later I change it further.

It's still keep these commits, because `NM_UNCONST_PTR()` does something I find (unfortunately) lacking in C: a form of C++'s const_cast<>().

thom311:
@bengal better now?

thom311:
merged to master

PR #65, opened 2018-02-08
th/ip6_temp_addr_sync_rh1542609

thom311:
Fix syncing IPv6 addresses

(the last commit is just some extra verbose logging for debugging. Ignore that one).

See also: https://bugzilla.redhat.com/show_bug.cgi?id=1542609

### DIFF
--- a/src/platform/nm-platform.c
+++ b/src/platform/nm-platform.c
@@ -3193,6 +3193,9 @@ array_ip6_address_position (const GPtrArray *addresses,
 	for (i = 0, pos = 0; i < len; i++) {
 		NMPlatformIP6Address *candidate = NMP_OBJECT_CAST_IP6_ADDRESS (addresses->pdata[i]);
 
+		if (!candidate)
+			continue;
+
 		if (   IN6_ARE_ADDR_EQUAL (&candidate->address, &address->address)
 		    && candidate->plen == address->plen
 		    && nm_utils_lifetime_get (candidate->timestamp,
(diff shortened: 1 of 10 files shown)


lkundrak:
> platform: add nm_platform_refresh_all() API
...
> It's only use is for testing
> really, because if there is a known condition in which the cache gets
> out of sync, that should be fixed instead.

Why include the patch then? Please elaborate on what is the "known condition"
because it is not known to me.


> platform: maintain proper order of addresses in platform cache during RTM_NEWADDR
...
>+       /* new objects are usually appended to the list. Except for
>+        * addresses, which are prepended.

Why?

> TEST

I guess you're not going to merge this one?

thom311:
> >   It's only use is for testing
> >   really, because if there is a known condition in which the cache gets
> >   out of sync, that should be fixed instead.
>
> Why include the patch then? Please elaborate on what is the "known condition"
> because it is not known to me.

Reworded commit message. It is a useful function like nm_ip4_config_dump(), which is never used in production code, but it's useful to have around during a printf debugging session. Same here, that is why I would merge the patch.



> >          * addresses, which are prepended.
>
> Why?

Improved(?) comment




> >    TEST
>
> I guess you're not going to merge this one?

Correct. I drop this one.


Repushed.

lkundrak:
LGTM

src/platform/nm-platform.c
```diff
@@ -3545,17 +3512,16 @@ gboolean
...
+	gs_unref_hashtable GHashTable *known_addresses_idx = NULL;
 	NMPLookup lookup;
 	guint32 ifa_flags;
-	gboolean remove = FALSE;
 
-	if (!_addr_array_clean_expired (AF_INET6, ifindex, known_addresses, now, NULL))
+	if (!_addr_array_clean_expired (AF_INET6, ifindex, known_addresses, now, &known_addresses_idx))
 		known_addresses = NULL;
 
 	/* @plat_addresses and @known_addresses are in increasing priority order */
```
bengal: The comment seems now obsolete, as the commit messages says:

> Note that @known_addresses contains the desired addresses in order of
    least-important first, while @plat_addresses contains them in opposite
    order.

bengal:
Otherwise LGTM

bengal:
The following crash was caught by CI running the th/ip6-temp-addr-sync-rh1542609 branch:

https://bugzilla.redhat.com/show_bug.cgi?id=1543640

bengal:
New commits look good.

thom311:
I merged this to master early, because the branch got some review.

I hurry on this one, because in my opinion the issue is quite severe (it makes IPv6 unusable for me at home).

I think this should make it to master/nm-1-10 as soon as possible, so that it gets early exposure to more testing and potential remaining issues can be found sooner. This issue should be fixed ASAP and there is almost no time left for including it in rhel-7.5. If we are going to include it, it must get early testing before it's 100% perfect.

PR #66, opened 2018-02-14
th/dhcp_client_id_bgo790660

thom311:
Extend `ipv4.dhcp-client-id` to use the MAC address of the current device.

Also, make the property configurable via global default.
For example with a file `/etc/NetworkManager/conf.d/30-dhcp-client-id.conf`:

```
[connection-30-dhcp-client-id]
ipv4.dhcp-client-id=mac
```

Addresses: https://bugzilla.gnome.org/show_bug.cgi?id=790660


thom311:
Note that `"mac"` and `"perm-mac"` generates a `01:xx` client id, while `"stable"` generates one with `00:...`. I think that makes sense, does it?

bengal:
LGTM

thom311:
Thanks. Merged to master.

freakynl:
Initial tests here are looking great. Boot file is attached. It's a somewhat modified version of
dnf --installroot=/some/path --releasever=27 group install 'LXDE Desktop' 
as initrd (/etc/machine-id cleared, /etc/fstab added, selinux disabled) with the RPM's I build from the pull request sources installed over the original ones.

Log shows it reading the suggested config file (in NetworkManager.conf under main dhcp=internal was also set) and getting an IP. On the DHCP server side I can see the MAC address as ID.

[log.txt](https://github.com/NetworkManager/NetworkManager/files/1740696/log.txt)

Thanks a lot :).

PR #67, opened 2018-02-16
th/netlink (not more libnl3)

thom311:
This branch drops the dependency on libnl3 and implements all the netlink related functions ourself.

This was done by copying all the missing bits over from libnl3 library, which is compatible licence-wise.

I did not try to simplify the code, except at a few places where it simplified the import itself. Hence, this is not yet as simple as it could be. For example, I think that `nl_cb` and `nl_socket` should essentially cease to exist and instead the relevant parameters should be passed on as needed -- instead of attaching them to the `nl_socket` object. That is not yet done intentionally, to first get a function equivalent form.


vbenes:
crash here: https://bugzilla.redhat.com/show_bug.cgi?id=1546656

thom311:
> crash here: https://bugzilla.redhat.com/show_bug.cgi?id=1546656

This crash can also happen on master. Attached patch to downstream bug (same patch is also part of this pull request for the moment).

bengal:
Looks good to me.

PR #69, opened 2018-02-19
fg/dhcp_lease-rh1503587

fgiudici:
Last commit somehow "optional": if more work needed we may move it in a branch apart.
We should instead do something for the IPv4 lease renewal issue.
https://bugzilla.redhat.com/show_bug.cgi?id=1503587


thom311:
Post merge review: lgtm

Merged as b9e22ece2d51510c81fa4286a6a26932d93d3e8a

PR #70, opened 2018-02-19
th/connectivity_rework

thom311:
Rework of NMConnectivity.

This serves to purposes:

  - properly cancel asynchronous connectivity checks when NetworkManager is about to terminate. This is necessary if we want to cleanup and destruct all objects in a timely manner.

  - rework the API, so that it becomes simpler to schedule and cancel items. This will be useful for https://bugzilla.gnome.org/show_bug.cgi?id=792240 , where we no longer will schedule connectivity checks every 30 seconds, but after connectivity changes we will first check more frequently to better handle short outages.

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -164,6 +164,7 @@ struct _NMDeviceConnectivityHandle {
 	gpointer user_data;
 	NMConnectivityCheckHandle *c_handle;
 	guint64 seq;
+	bool is_periodic:1;
 };
 
 /*****************************************************************************/
@@ -543,7 +544,19 @@ typedef struct _NMDevicePrivate {
 
 	NMConnectivity *concheck_mgr;
 
-	gulong concheck_periodic_id;
+	/* if periodic checks are enabled, this is the source id for the next check. */
+	guint concheck_p_cur_id;
+
+	/* the currently configured max periodic interval. */
+	guint concheck_p_max_interval;
+
+	/* the current interval. If we are probing, the interval might be lower
+	 * then the configured max interval. */
+	guint concheck_p_cur_interval;
+
+	/* the timestamp, when we last scheduled the timer concheck_p_cur_id with current interval
+	 * concheck_p_cur_interval. */
+	gint64 concheck_p_cur_basetime_ns;
 
 	NMConnectivityState connectivity_state;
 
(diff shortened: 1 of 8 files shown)


thom311:
crash report http://faf.lab.eng.brq.redhat.com/faf/reports/5601/

thom311:
> crash report http://faf.lab.eng.brq.redhat.com/faf/reports/5601/

fixed and repushed

src/nm-connectivity.c
```diff
@@ -41,7 +41,7 @@ typedef struct {
 	struct curl_slist *request_headers;
 	guint timeout_id;
 	char *ifspec;
-} ConCheckCbData;
+} NMConnectivityCheckHandle;
```
bengal: Since the structure is local, can we avoid to namespace it with the NM
prefix? For example, "ConnectivityCheckHandle"

src/devices/nm-device.c
```diff
@@ -2157,17 +2170,213 @@ nm_device_get_physical_port_id (NMDevice *self)
...
+		goto out;
+	}
+
+	nm_assert (interval_ns >= 0);
+
+	if (!nm_device_is_real (self))
+		goto out;
+
+	/* we enable periodic checks for every device state (except UNKNOWN). Especially with
+	 * unmanaged devices, it is interesting to know whether we have connectivity on that device. */
```
bengal: We should at least check that the device is realized, and probably also avoid loopback.

I wonder if we should keep ignoring unmanaged devices as before, because the promise is that they don't exist from
NM point of view. But I agree that if there is a unmanaged device that can reach the internet, it's useful to know it to show the real connectivity state.

src/nm-connectivity.c
thom311: This is a trival commit that does only renaming in preparation of the follow up commits. The follow up commit makes this handle public (at least, it's name. It will still be an opaque structure).

src/devices/nm-device.c
thom311: > We should at least check that the device is realized,

a few lines above there is a check for `nm_device_is_real()`. Added check for loopback.

Let's enable connectivity check eagerly for the moment, and see how that works. I think there is a point in doing connectivity checks for unmanaged devices.


Repushed.

bengal: > a few lines above there is a check for nm_device_is_real(). Added check for loopback.

Right, it's the manual connectivity check that doesn't verify whether a device is realized and so we start the check also on devices that don't exist:

    connectivity: (pppoe) start request to 'http://fedoraproject.org/static/hotspot.txt'
    connectivity: (pppoe) check completed: LIMITED; check failed with curl status 45

thom311: Ah... the idea was that `nm_device_check_connectivity()` never fails and also doesn't schedule it's own idle handler -- because `nm_connectivity_check_start()` already schedules an idle handler when there is nothing to do. The point was a consistent behavior. You are right, that is not great. Pushed a fixup. How about now?

bengal: LGTM

thom311:
merged part of to master: 9c4919be764493b35553583723f77018e7974569

vbenes:
Rebuild

thom311:
merged to master

src/nm-connectivity.c
```diff
@@ -105,111 +130,172 @@ NM_DEFINE_SINGLETON_GETTER (NMConnectivity, nm_connectivity_get, NM_TYPE_CONNECT
...
+		curl_easy_setopt (cb_data->concheck.curl_ehandle, CURLOPT_WRITEDATA, NULL);
+		curl_easy_setopt (cb_data->concheck.curl_ehandle, CURLOPT_HEADERFUNCTION, NULL);
+		curl_easy_setopt (cb_data->concheck.curl_ehandle, CURLOPT_HEADERDATA, NULL);
+		curl_easy_setopt (cb_data->concheck.curl_ehandle, CURLOPT_PRIVATE, NULL);
+		curl_easy_setopt (cb_data->concheck.curl_ehandle, CURLOPT_HTTPHEADER, NULL);
+
+		priv = NM_CONNECTIVITY_GET_PRIVATE (self);
+
+		curl_multi_remove_handle (priv->concheck.curl_mhandle, cb_data->concheck.curl_ehandle);
+		curl_easy_cleanup (cb_data->concheck.curl_ehandle);
```
andersk: I’m trying to investigate a heap corruption bug at [LP #1792745](https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/1792745), and I find this line suspicious in light of the above comment. Valgrind seems to be [showing](https://launchpadlibrarian.net/388662598/nm.19802.log) invalid writes through the easy handle after it’s been freed here.

thom311: This is a duplicate of https://bugzilla.gnome.org/show_bug.cgi?id=797136
Yes, there is a bug related to this.

PR #71, opened 2018-02-21
th/keyfile_cert_parser

thom311:
Some refactoring / cleanup of keyfile parsing certificates.

I think the only visible change in behavior, is that I think previously an empty certificate would have caused an assertion `handle_as_scheme: assertion 'data && data_len > 0' failed`. This is fixed.


thom311:
merged

PR #72, opened 2018-02-21
iwd: Simpler handling of wifi secrets

balrog-kun:
I'll appreciate a thorough look at the changes as I'm not 100% sure that for example changing the NMDevice state to NEED_AUTH and back to CONFIG from the subclass won't break anything.  I'm also not sure whether nm_connection_clear_secrets() actually does anything without some other call to make the change be saved to file (doesn't really matter as I'm setting the NM_SECRET_AGENT_GET_SECRETS_FLAG_REQUEST_NEW flag anyway, but this could be avoided)

### DIFF
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -1109,7 +1109,7 @@ secrets_error:
 	cleanup_association_attempt (self, TRUE);
 }
 
-static bool
+static void
 wifi_secrets_get_secrets (NMDeviceIwd *self,
                           const char *setting_name,
                           NMSecretAgentGetSecretsFlags flags)
@@ -1120,7 +1120,7 @@ wifi_secrets_get_secrets (NMDeviceIwd *self,
 	wifi_secrets_cancel (self);
 
 	req = nm_device_get_act_request (NM_DEVICE (self));
-	g_return_val_if_fail (NM_IS_ACT_REQUEST (req), FALSE);
+	g_return_if_fail (NM_IS_ACT_REQUEST (req));
 
 	priv->wifi_secrets_id = nm_act_request_get_secrets (req,
 	                                                    TRUE,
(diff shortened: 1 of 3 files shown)


vhumpa:

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1881,31 +1753,44 @@ nm_device_iwd_set_dbus_object (NMDeviceIwd *self, GDBusObject *object)
...
 
-	return nm_setting_wireless_security_get_psk (s_wireless_sec);
+	nm_device_state_changed (NM_DEVICE (self), NM_DEVICE_STATE_NEED_AUTH,
+	                         NM_DEVICE_STATE_REASON_NO_SECRETS);
+
+	if (!wifi_secrets_get_secrets (self,
+	                               NM_SETTING_WIRELESS_SECURITY_SETTING_NAME,
+	                               NM_SECRET_AGENT_GET_SECRETS_FLAG_ALLOW_INTERACTION
+		                         | NM_SECRET_AGENT_GET_SECRETS_FLAG_REQUEST_NEW))
+		return FALSE;
```
thom311: as said above, `wifi_secrets_get_secrets` cannot fail.

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1881,31 +1753,44 @@ nm_device_iwd_set_dbus_object (NMDeviceIwd *self, GDBusObject *object)
...
+	psk = nm_setting_wireless_security_get_psk (s_wireless_sec);
+	if (psk) {
+		_LOGD (LOGD_DEVICE | LOGD_WIFI,
+		       "Returning the PSK to the IWD Agent");
 
-	s_wireless_sec = nm_connection_get_setting_wireless_security (connection);
-	if (!s_wireless_sec)
-		return NULL;
+		g_dbus_method_invocation_return_value (invocation,
+	                                               g_variant_new ("(s)", psk));
```
thom311: indentation (tab vs. space)

balrog-kun:
Thanks for the check, updated the patch.

balrog-kun:
Updated once again with a minor change to account for the new NEED_AUTH state meaning.

thom311:
Obsoleted by #73

PR #73, opened 2018-03-01
iwd: secrets work and Powered changes

balrog-kun:
This includes and supersedes #72 as github PRs are a bit of a mess, I'm just adding two changes to the PR.

### DIFF
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -334,7 +334,8 @@ get_ordered_networks_cb (GObject *source, GAsyncResult *res, gpointer user_data)
 		props = g_variant_new ("a{sv}", &builder);
 
 		ap = nm_wifi_ap_new_from_properties (path, props);
-		nm_wifi_ap_set_ssid (ap, (const guint8 *) name, strlen (name));
+		if (name[0] != '\0')
+			nm_wifi_ap_set_ssid (ap, (const guint8 *) name, strlen (name));
 		nm_wifi_ap_set_strength (ap, nm_wifi_utils_level_to_quality (signal / 100));
 		nm_wifi_ap_set_freq (ap, 2417);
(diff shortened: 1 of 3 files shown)


vhumpa:

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1109,7 +1109,7 @@ wifi_secrets_cb (NMActRequest *req,
 	cleanup_association_attempt (self, TRUE);
 }
 
-static bool
+static void
```
thom311: you did the fixups on the second commit, but they really belong to the first one. Could you rebase the branch so that the second commit has the quit simple change for -1 to G_MAXINT?

While at it, could you rebase the branch on master?

The rest looks good to me.

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1040,31 +1063,53 @@ wifi_secrets_cb (NMActRequest *req,
...
+	/* Change state back to what it was before NEED_AUTH */
+	nm_device_state_changed (device, NM_DEVICE_STATE_CONFIG, NM_DEVICE_STATE_REASON_NONE);
+	return;
+
+secrets_error:
+	nm_device_state_changed (device,
+	                         NM_DEVICE_STATE_FAILED,
+	                         NM_DEVICE_STATE_REASON_NO_SECRETS);
+
+	if (priv->secrets_request) {
```
thom311: maybe first reply the secrets request before changing device state. It's also done in that order in the success-case above.

src/devices/wifi/nm-device-iwd.c
balrog-kun: Oops, yes, fixing.

balrog-kun:
I added one more commit here.

thom311:
Merged as 29e9d206aa8319c5875749cf3c3bd7b96dc0fc4e

Thanks!

PR #74, opened 2018-03-02
th/dbus rework D-Bus implementation in NetworkManager daemon

thom311:
$SUBJECT


bengal:
Now the introspection XMLs are used only to generate the documentation, and there is the possibility that they become out of sync with the actual D-Bus API, right?

The branch looks very good to me!

thom311:
> Now the introspection XMLs are used only to generate the documentation, and there is the possibility
> that they become out of sync with the actual D-Bus API, right?

Yes, when extending the D-Bus API, you need to adjust the introspection xml and the `GDBusInterfaceInfo` instance. But in my opinion that happens seldomly enough to be a minor burden.

thom311:
Thanks Beniamino

merged as 1e535789cfd8dbc94da1e4f6393d27ccc2a209fc

PR #78, opened 2018-03-14
th/complete generic cleanup

thom311:
some cleanup of using `GSList` for `nm_utils_complete_generic`.

### DIFF
--- a/src/NetworkManagerUtils.c
+++ b/src/NetworkManagerUtils.c
@@ -67,32 +67,35 @@ nm_utils_get_shared_wifi_permission (NMConnection *connection)
 /*****************************************************************************/
 
 static char *
-get_new_connection_name (const GSList *existing,
+get_new_connection_name (const GSList *existing_connections,
                          const char *preferred,
                          const char *fallback_prefix)
 {
 	GSList *names = NULL;
 	const GSList *iter;
 	char *cname = NULL;
-	int i = 0;
-	gboolean preferred_found = FALSE;
+	guint i;
 
 	g_assert (fallback_prefix);
 
-	for (iter = existing; iter; iter = g_slist_next (iter)) {
+	for (iter = existing_connections; iter; iter = g_slist_next (iter)) {
 		NMConnection *candidate = NM_CONNECTION (iter->data);
 		const char *id;
 
 		id = nm_connection_get_id (candidate);
-		g_assert (id);
+		nm_assert (id);
+
 		names = g_slist_append (names, (gpointer) id);
 
-		if (preferred && !preferred_found && (strcmp (preferred, id) == 0))
-			preferred_found = TRUE;
+		if (   preferred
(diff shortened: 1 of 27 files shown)


src/devices/nm-device.c
```diff
@@ -4767,34 +4767,50 @@ nm_device_generate_connection (NMDevice *self,
 	return g_steal_pointer (&connection);
 }
 
+/**
+ * nm_device_complete_connection:
+ * @self:
+ * @connection:
+ * @specific_object:
+ * @existing_connection:
+ * @error:
```
bengal: Either document parameters or don't add them :)

src/devices/wwan/nm-modem.c
```diff
@@ -1094,9 +1094,17 @@ nm_modem_complete_connection (NMModem *self,
...
+
+	klass = NM_MODEM_GET_CLASS (self);
+	if (!klass->complete_connection) {
+		g_set_error (error, NM_DEVICE_ERROR, NM_DEVICE_ERROR_INVALID_CONNECTION,
+		             "Modem class %s had no complete_connection method",
+		             G_OBJECT_TYPE_NAME (self));
+		return FALSE;
+	}
+
+	return NM_MODEM_GET_CLASS (self)->complete_connection (self, connection, existing_connections, error);
```
bengal: return klass->complete_connection (...)

src/NetworkManagerUtils.c
```diff
@@ -67,85 +67,85 @@ nm_utils_get_shared_wifi_permission (NMConnection *connection)
...
+			if (nm_streq (iter->data, temp)) {
+				cname = temp;
+				goto found;
 			}
 		}
-		if (!found)
-			cname = temp;
-		else
-			g_free (temp);
+		g_free (temp);
```
bengal: The search logic looks wrong here, after we find a match we should skip the generated name and try another.

src/NetworkManagerUtils.c
```diff
@@ -112,28 +116,25 @@ get_new_connection_name (const GSList *existing_connections,
...
 		temp = g_strdup_printf (C_("connection id fallback", "%s %u"),
 		                        fallback_prefix, i);
-		for (iter = names; iter; iter = g_slist_next (iter)) {
-			if (nm_streq (iter->data, temp)) {
-				cname = temp;
-				goto found;
-			}
-		}
+
+		if (nm_utils_strv_find_first (NM_UNCONST_PPTR (char, existing_names),
```
bengal: Could we use just: "(char **) existing names" instead of the macro? It seems more readable to me and I don't have to look up what the macro does.

Ok, this fixes the search.

thom311:
Addressed all and merged to master: c96c374d14be91db7a0ba2307b3d1f35bcd45707

Thanks!

PR #79, opened 2018-03-15
th/dbus-auth-external-bgo793116

thom311:
dbus: disable all but EXTERNAL D-Bus authentication for private sockets

We use a private D-Bus socket for example for DHCP clients to report back
at unix:path=/var/run/NetworkManager/private-dhcp.

By default, gdbus will enable the authentication mechanisms EXTERNAL
and DBUS_COOKIE_SHA1. However, DBUS_COOKIE_SHA1 requires a /root/.dbus-keyrings
directory, which is not available to NetworkManager as it is started with
ProtectHome=read-only. And writing to /root would be a bad idea anyway.

This leads to a warning

  NetworkManager[10962]: Error adding entry to keyring: Error creating directory “/root/.dbus-keyrings”: Read-only file system

Disable all but the EXTERNAL mechanism.

See-also: https://dbus.freedesktop.org/doc/dbus-specification.html#auth-mechanisms

https://bugzilla.gnome.org/show_bug.cgi?id=793116

### DIFF
--- a/src/nm-dbus-manager.c
+++ b/src/nm-dbus-manager.c
@@ -305,6 +305,14 @@ private_server_authorize (GDBusAuthObserver *observer,
 	return g_credentials_get_unix_user (credentials, NULL) == 0;
 }
 
+static gboolean
+private_server_allow_mechanism (GDBusAuthObserver *observer,
+                                const char *mechanism,
+                                gpointer user_data)
+{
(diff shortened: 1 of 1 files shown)


bengal:
LGTM

thom311:
merged as 5cd9340f831c5fab1e8749fcc3de4722d1701439

PR #75, opened 2018-03-08
th/ipv6_ll_reapply_rh1552069

thom311:
Attempt to fix handling of IPv6 LL address, especially in light of reapply.

This shall fix test `ipv6_route_set_route_with_tables_reapply`

See also https://bugzilla.redhat.com/show_bug.cgi?id=1552069

PR #80, opened 2018-03-20
th/ipv6_ll_reapply_rh1552069-pt2

thom311:
This is a spin-off from #75. It contains all of #75's patches + something on top.

It's not for review, but to track the patches and CI.


thom311:
this approach of retrying to add the addresses in `route_temporary_not_available_x` for 20 or 60 seconds, is of course flawed. Any better idea?

bengal:
> this approach of retrying to add the addresses in route_temporary_not_available_x for 20 or 60 seconds, is of course flawed. Any better idea?

I don't have any other ideas, the approach looks ok to me.

thom311:
Ok, reworked the second part entirely, now "platform: adding onlink gateway route for manual addresses".
(previous version is at `th/ipv6-ll-reapply-rh1552069-pt2-v1`)

How is this?

bengal:
LGTM

thom311:
merged to master.

Unfortunately, despite the goal of this branch was to fix CI, now more test are broken than before. I still merged it, because I think that the changes are going into the right direction and when the changes are on master, it hopefully becomes easier to tackle the remaining failures.

PR #83, opened 2018-03-28
th/checkpoint

thom311:
Some fixes, refactoring, and new features for checkpoints

### DIFF
--- a/introspection/org.freedesktop.NetworkManager.xml
+++ b/introspection/org.freedesktop.NetworkManager.xml
@@ -252,6 +252,27 @@
     </method>
 
     <!--
+        CheckpointAdjustRollbackTimeout:
+        @add_timeout: number of seconds from ~now~ in which the
+          timeout will expire. Set to 0 to disable the timeout.
+          Note that the added seconds start counting from now,
+          not "Created" timestamp or the previous expiration
+          time. Note that the "Created" property of the checkpoint
+          will stay unchanged by this call. However, the "RollbackTimeout"
+          will be recalculated to give the approximate new expiration time.
+          The new "RollbackTimeout" property will be approximate up to
+          one second precision, which is the accuracy of the property.
+
+        Reset the timeout for rollback for the checkpoint.
+
+        Since: 1.12
+    -->
+    <method name="CheckpointAdjustRollbackTimeout">
+      <arg name="checkpoint" type="o" direction="in"/>
+      <arg name="add_timeout" type="u" direction="in"/>
+    </method>
+
+    <!--
         Devices:
 
         The list of realized network devices. Realized devices are those which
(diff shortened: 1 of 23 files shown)


src/nm-dbus-utils.c
```diff
@@ -147,6 +147,36 @@ nm_dbus_utils_g_value_set_object_path_array (GValue *value,
 	g_value_take_boxed (value, paths);
 }
 
-/*****************************************************************************/
+void
+nm_dbus_utils_g_value_set_object_path_from_hash (GValue *value,
+                                                 GHashTable *hash /* is has keys of NMDBusObject type. */,
```
bengal:
```
/* is has keys of NMDBusObject type. */,
   ^^^^^

File: introspection/org.freedesktop.NetworkManager.xml [hunk 2]
```diff
@@ -251,6 +251,27 @@
       <arg name="result" type="a{su}" direction="out" />
     </method>
 
+    <!--
+        CheckpointAdjustRollbackTimeout:
+        @add_timeout: number of seconds from ~now~ in which the
+          timeout will expire. Set to 0 to disable the timeout.
+          Note that the added seconds start counting from now,
+          not Created timestamp or the previous expiriation
```
bengal: expiration

libnm-core/nm-dbus-interface.h
```diff
@@ -842,6 +842,18 @@ typedef enum {
  *   delete any new connection added after the checkpoint (Since: 1.6)
  * @NM_CHECKPOINT_CREATE_FLAG_DISCONNECT_NEW_DEVICES: upon rollback,
  *   disconnect any new device appeared after the checkpoint (Since: 1.6)
+ * @NM_CHECKPOINT_CREATE_FLAG_ALLOW_OVERLAPPING: by default, creating
+ *   a checkpoint fails if there are already existing checkoints that
+ *   reference the same devices. With this flag, creation of such
+ *   checkpoints is allowed, however, if an older checkpoint
+ *   that references overlapping devices gets rolled back, it will
+ *   automatically destroy this checkpoint during rollback. This
+ *   allows to create several overlappig checkpoints in parallel,
```
bengal: overlapping

bengal:
> checkpoint: allow resetting the rollback timeout via D-Bus

> The libnm API only implements the async form (at least for now).
> Sync methods are fundamentally wrong with D-Bus, and it's probably
> not needed. Also, follow glib convenction, where the async form
> doesn't have the _async name suffix.

If so, can you also rename the existing checkpoint functions to follow
the same pattern? They are not released yet and so it's not an API
breakage.

thom311:
fixed, repushed.

Note the FIXME comments added by the last commit. There is still follow-up work to do (maybe not as part of this branch).

bengal:
LGTM

thom311:
Thanks. merged

PR #85, opened 2018-03-30
th/settings_cleanup

thom311:
Various unrelated refactoring

### DIFF
--- a/src/settings/nm-settings.c
+++ b/src/settings/nm-settings.c
@@ -166,6 +166,11 @@ static void connection_ready_changed (NMSettingsConnection *conn,
                                       GParamSpec *pspec,
                                       gpointer user_data);
 
+static void default_wired_clear_tag (NMSettings *self,
+                                     NMDevice *device,
+                                     NMSettingsConnection *connection,
+                                     gboolean add_to_no_auto_default);
+
 /*****************************************************************************/
 
 static void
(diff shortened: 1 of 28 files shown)


vbenes:
Rebuild

vbenes:
I've seen a crash here https://bugzilla.redhat.com/show_bug.cgi?id=1565030

Does it need fixing here?

thom311:
> I've seen a crash here https://bugzilla.redhat.com/show_bug.cgi?id=1565030
>
> Does it need fixing here?

Yes, it's a bug. The bug is not only in this branch, but becomes visible by changes of this branch. Needs additional fixes on top of it...

thom311:
merged as commit f69dbd71fe1f714250cc865b4c2e50fcc95579d2

src/settings/nm-settings.c
```diff
@@ -896,16 +880,27 @@ connection_removed (NMSettingsConnection *connection, gpointer user_data)
...
+		                            "(o)",
+		                            nm_dbus_object_get_path (NM_DBUS_OBJECT (connection)));
+	}
+
+	nm_dbus_object_unexport (NM_DBUS_OBJECT (connection));
+
+	if (priv->connections_loaded)
+		g_signal_emit (self, signals[CONNECTION_REMOVED], 0, connection);
+
+	g_object_ref (connection);
```
bengal: Can you add a comment explaining why you change the order? It's not clear to me.

thom311: Hm... the order is anyway something we define, so that it makes the most sense. For example, it makes (more) sense (to me) emitting `PropertyChanged` signal on `NMSettings:Connections`, before emitting `NMSettings:ConnectionRemoved` signal on D-Bus. I changed it, because I think the order as it is now, makes the most sense considering all the alternatives.

(note that before the D-Bus rework, the `NMSettings:Connections` property changed signal would always be sent on an idle handler, so it wasn't even (realistically) possible to emitting the D-Bus signal before `NMSettings:ConnectionRemoved` signal. Now it is, and that was changed).

Let's not look at this intermediate commit (because it was changed afterwards again, sorry). But for example, previously we would emit the internal signal `NM_SETTINGS_SIGNAL_CONNECTION_REMOVED` while the connection is still exported on D-Bus. I think, the other order is better, because AFAIS, the signal handlers care about doing additional cleanup after the connection is mostly already gone.

I don't have exact reasons to comment why the order should be exactly like that, except, it feels more right to me. So, don't know what to comment beyond explaining what the order of events *is*. That is, I cannot easily explain *why* the order is that way, only a trival comment saying that it *is* that way. So, I'd rather not add a comment.

Does that answer the question? Do you disagree with the *is* and have a stronger *why* then I can provide?

bengal: Okay

PR #86, opened 2018-04-05
th/auth_chain

thom311:
We use NMAuthChain frequently, and I always felt it to be heavier than necessary. Some refactoring, with the overall goal to require less memory allocations, both to start the auth-request, and to track the additional data.


bengal:
LGTM

thom311:
Thanks.

Merged

PR #87, opened 2018-04-05
th/platform_tun

thom311:
Now that tun support via netlink is [merged](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1ec010e705934c8acbe7dbf31afc81e60e3d828b), enable the code in NetworkManager.

Also, add a test, and various fixes.

See also: commit 39ab38a04d580b3d9eeae0d242b7ac9b4fb53378

### DIFF
--- a/src/devices/tests/test-lldp.c
+++ b/src/devices/tests/test-lldp.c
@@ -347,8 +347,7 @@ static void
 _test_recv_fixture_setup (TestRecvFixture *fixture, gconstpointer user_data)
 {
 	const NMPlatformLink *link;
-	struct ifreq ifr = { };
-	int fd, s;
+	nm_auto_close int fd = -1;
 
 	fd = open ("/dev/net/tun", O_RDWR | O_CLOEXEC);
 	if (fd == -1) {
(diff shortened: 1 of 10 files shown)


src/devices/tests/test-lldp.c
```diff
@@ -357,20 +356,45 @@ _test_recv_fixture_setup (TestRecvFixture *fixture, gconstpointer user_data)
...
+
+		/* Bring the interface up */
+		s = socket (AF_INET, SOCK_DGRAM | SOCK_CLOEXEC, 0);
+		g_assert (s >= 0);
+		ifr.ifr_flags |= IFF_UP;
+		g_assert (ioctl (s, SIOCSIFFLAGS, &ifr) >= 0);
+		nm_close (s);
+
+		link = nmtstp_assert_wait_for_link (NM_PLATFORM_GET, TEST_IFNAME, NM_LINK_TYPE_TUN, 100);
+	}
```
bengal: Maybe, just use nmtstp_link_tun_add() to create the tun device, so that we can remove the manual creation from the test.

thom311:
merged

PR #88, opened 2018-04-05
th/settings_flags_rh1460295

thom311:
Add flags for settings-connection on D-Bus.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1460295

I am not entirely sure about the name `NMSettingsConnectionFlags`. In libnm API, the profile is of type `NMRemoteConnection`, so it's not clear what the "settings" part here means. However, we have the term "connection" used over and over. I think it's good to call this thing "settings-connection". That also follows the D-Bus API where it's called `org.freedesktop.NetworkManager.Settings.Connection` interface.


thom311:
merged

PR #90, opened 2018-04-11
th/polkit-cleanup

thom311:
Eventually, I'd like to drop the dependency of nmcli to `libpolkit-agent-1` and `libpolkit-gobject-1`.

This is a small set of patches of refactoring, towards that goal.

Please review


thom311:
merged

PR #91, opened 2018-04-12
th/manager-cleanup

thom311:
A branch with various cleanups, mostly related to NMManager and activation handling.

This is some independent patches that come from implementing [rh#1555012](https://bugzilla.redhat.com/show_bug.cgi?id=1555012), and I'd like to merge early them before the entire branch gets too large.

Thanks


bengal:
> core: cleanup code by using nm_auth_is_subject_in_acl_set_error()

Does it make sense to pass the error code? Couldn't the function just
always set NM_MANAGER_ERROR_PERMISSION_DENIED?

The rest LGTM.

thom311:
> Does it make sense to pass the error code? 

one caller passes a different error code. The error code is (AFAIS) exported on D-Bus and thus public API.

PR #93, opened 2018-04-12
dhcp: don't enforce broadcast flag

richardschuetz:
Requesting broadcast replies from the DHCP server can be problematic in
filtered environments like some wireless networks. Don't override the
default of using unicast. This matches the behaviour of the external DHCP
clients.

### DIFF
--- a/src/dhcp/nm-dhcp-systemd.c
+++ b/src/dhcp/nm-dhcp-systemd.c
@@ -630,12 +630,6 @@ ip4_start (NMDhcpClient *client, const char *dhcp_anycast_addr, const char *last
 		goto error;
 	}
 
-	r = sd_dhcp_client_set_request_broadcast (priv->client4, true);
-	if (r < 0) {
-		_LOGW ("failed to enable broadcast mode (%d)", r);
-		goto error;
-	}
-
 	dhcp_lease_load (&lease, priv->lease_file);
 
 	if (last_ip4_address)


vhumpa:

thom311:
> This matches the behaviour of the external DHCP clients.

Do you mean `dhclient`? Yes, seems `bootp-broadcast-always` is disabled. Do you mean that?


networkd's manpage has:
```
RequestBroadcast=
  Request the server to use broadcast messages before the IP address has been configured.
  This is necessary for devices that cannot receive RAW packets, or that cannot receive
  packets at all before an IP address has been configured. On the other hand, this must
  not be enabled on networks where broadcasts are filtered out.
```
also [[1]](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/systemd/src/libsystemd-network/sd-dhcp-client.c?id=69b7a76dc27f71981b7eb68bcce134654eb6fe7d#n583)


Maybe we need to make this configurable?

richardschuetz:
> Do you mean `dhclient`? Yes, seems `bootp-broadcast-always` is disabled. Do you mean that?

Yes. Same holds true for `dhcpcd`. 

> Maybe we need to make this configurable?

Maybe. At least don't override the default behaviour for no apparent reason.

bengal:
Patch looks good to me.

thom311:
merged. Thanks!!

PR #95, opened 2018-04-16
Add ifcfg-rh test for TC configuration and fix a parsing bug

bengal:


src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-reader.c
```diff
@@ -1995,11 +1995,15 @@ make_tc_setting (shvarFile *ifcfg)
...
-		if (!qdisc)
-			PARSE_WARNING ("ignoring bad qdisc: '%s': %s", value, local->message);
+		if (!qdisc) {
+			PARSE_WARNING ("ignoring bad tc qdisc: '%s': %s", value, local->message);
+			continue;
+		}
 
 		if (!nm_setting_tc_config_add_qdisc (s_tc, qdisc))
-			PARSE_WARNING ("duplicate qdisc");
+			PARSE_WARNING ("duplicate tc qdisc");
```
thom311: another pre-existing bug here is:

Neither `nm-setting-tc-config.c:set_property()` nor `nm-setting-tc-config.c:verify()` has the behavior of rejecting duplicate entries. That means, the API `nm_setting_tc_config_add_qdisc()` does not allow to create such configurations, that are otherwise creatable and valid. Also, ifcfg-writer would write them, ifcfg-reader could not read them back. I think the fix is simply to not let the add function reject duplicates.

bengal: I don't think a configuration with duplicate qdiscs or filters can be valid. @lkundrak ?

If so, the TC setting should implement verify()	and reject duplicate entries.

libnm-core/nm-utils.c
```diff
@@ -600,22 +600,38 @@ _nm_utils_copy_array_to_slist (const GPtrArray *array,
...
-                      NMUtilsCopyFunc copy_func,
-                      GDestroyNotify free_func)
+_nm_utils_copy_array_dedup (const GPtrArray *array,
+                            NMUtilsCopyFunc copy_func,
+                            NMUtilsEqualFunc equal_func,
+                            GDestroyNotify free_func)
 {
 	GPtrArray *copy;
-	int i;
+	int i, j;
```
thom311: guint ?

libnm-core/nm-utils.c
```diff
@@ -600,22 +600,38 @@ _nm_utils_copy_array_to_slist (const GPtrArray *array,
...
 
 	if (!array)
 		return g_ptr_array_new_with_free_func (free_func);
 
 	copy = g_ptr_array_new_full (array->len, free_func);
-	for (i = 0; i < array->len; i++)
+	for (i = 0; i < array->len; i++) {
+		for (j = 0; equal_func && j < copy->len; j++) {
```
thom311: Could we move the loop inside a `if (equal_func) {` block? I know, likely the compiler can hoist the check outside the loop, but these forms of `for(;;)` look so unfamiliar to me.

libnm-core/nm-utils.c
```diff
@@ -600,22 +600,38 @@ _nm_utils_copy_array_to_slist (const GPtrArray *array,
...
+			if (equal_func (array->pdata[i], copy->pdata[j]))
+				goto next;
+		}
 		g_ptr_array_add (copy, copy_func (array->pdata[i]));
```
thom311: how about, making `copy_func` and optional argument too? Now that there is the dedup functionality, it's not clear that `copy_func` is a mandator argument.

libnm-core/nm-setting-tc-config.c
```diff
@@ -1087,15 +1087,17 @@ set_property (GObject *object, guint prop_id,
 	switch (prop_id) {
 	case PROP_QDISCS:
 		g_ptr_array_unref (self->qdiscs);
-		self->qdiscs = _nm_utils_copy_array (g_value_get_boxed (value),
-		                                     (NMUtilsCopyFunc) nm_tc_qdisc_dup,
-		                                     (GDestroyNotify) nm_tc_qdisc_unref);
+		self->qdiscs = _nm_utils_copy_array_dedup (g_value_get_boxed (value),
```
thom311: the problem is, when NetworkManager receives a setting on D-Bus, it will call this property getter, which silently drops duplicates. Wouln't it be better, if setting such properties is allowed in the first moment (because, you don't know what you will receive when parsing the D-Bus variant). Only in the next step, `nm_connection_verify()` either rejects duplicates with error, or `nm_connection_normalize()` normalizes them. So, one of `NM_SETTING_VERIFY_NORMALIZABLE`, `NM_SETTING_VERIFY_NORMALIZABLE_ERROR`, or `NM_SETTING_VERIFY_ERROR`

At this point (where OVS support is almost new API), I would just flag them as error. Not trying to normalize anything. The user did something wrong, don't try to fix it silently.

thom311:
"libnm-core: reject tc configurations with duplicate elements" lgtm

bengal:
Applied to master:

https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=2576e3a8e886c6ac216392efc84ac18d1b148df8

PR #98, opened 2018-04-19
th/core-fixes-for-shutdown

thom311:
Some more fixes in preparation of doing a clean shutdown when terminating NetworkManager.


bengal:
LGTM

PR #103, opened 2018-04-22
th/connectivity-check-on-activated

thom311:
When the device reaches state `activated`, trigger a connectivity check right away, and also reset the probe interval.


thom311:
merged

PR #104, opened 2018-04-23
[lr/version-dir] Qualify plugin dir name with a version string

lkundrak:
$subj plus two fixes.


thom311:
`d end up beoing of a different`
 ^^ typo

`build: qualify plugin dir name with a version string`
Did you encounter an actual issue due to this?

Patches lgtm

bengal:
LGTM

PR #105, opened 2018-04-23
[lr/nmcli-colors] Add terminal-colors.d(5) support

lkundrak:
$subj plus some cleanups.

The WIP tag is there because there's some unclear things in the spec. Otherwise the code is ready from my point of view.


clients/cli/nmcli.c
```diff
@@ -676,6 +890,8 @@ nmc_cleanup (NmCli *nmc)
 		nmc->pager_pid = 0;
 	}
 
+	g_free (nmc->color_scheme);
```
thom311: let's not leave dangling pointers in the global struct. E.g. use `nm_clear_g_free()`.

clients/cli/nmcli.c
```diff
@@ -331,21 +331,226 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
+		if (!isgraph (*p)) {
+			g_set_error (error, NMCLI_ERROR, 0,
+			             _("Expected a value for '%s'\n"), name);
+			return FALSE;
+		}
+		while (isgraph (*p))
+			p++;
+
+		/* Trailing whitespace. */
+		while (isblank (*p)) {
```
thom311: could we avoid these functions, like `isblank()`, etc. They are all locale dependent, use instead `g_ascii_is*()`.

clients/cli/nmcli.c
```diff
@@ -331,21 +331,226 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
+                                 const char *base_dir)
+{
+	check_colors_files_for_name (nmc, color_option, base_dir, "nmcli");
+	check_colors_files_for_name (nmc, color_option, base_dir, NULL);
+}
+
+static gboolean
+parse_color_scheme (NmCli *nmc, GError **error)
+{
+	char *p = nmc->color_scheme;
```
thom311: It's a bit ugly, that the `color_scheme` field gets mangled. If you just want to have this field so that the pointers in `palette` don't need to clone the strings, then call it something like `palette_buffer`. Clearly, after `parse_color_scheme()` this field is no longer the (original) color scheme.

Likewise, the comment of the field `/* The color scheme in terminal-colors.d(5) format. */` is not correct. It's only the scheme *before* parsing.

man/nmcli.xml
```diff
@@ -2035,6 +2039,269 @@ It's equivalent to the <literal>+bond.options 'option=value'</literal> syntax.</
 
   </refsect1>
 
+  <refsect1 id='colors'><title id='colors.title'>Colors</title>
+    <para>Implicit coloring can be disabled by an empty file
+    <filename>/etc/terminal-colors.d/fdisk.disable</filename>.</para>
```
thom311: fdisk?

clients/cli/nmcli.c
```diff
@@ -355,8 +560,17 @@ set_colors (NmCli *nmc, NmcColorOption *color_option)
 		nmc->nmc_config_mutable.use_colors = FALSE;
 		break;
 	}
+
+	if (nmc->nmc_config_mutable.use_colors && nmc->color_scheme) {
+		if (!parse_color_scheme (nmc, &error)) {
+			g_printerr (_("Error parsing color scheme: %s"), error->message);
```
thom311: Does this really warrant a warning?

clients/cli/nmcli.c
```diff
@@ -331,21 +331,226 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
+			p++;
+		if (*p == '\0') {
+			g_set_error (error, NMCLI_ERROR, 0,
+			             _("Unexpected end of file following '%s'\n"), name);
+			return FALSE;
+		}
+
+		/* Separating whitespace. */
+		if (!isblank (*p)) {
+			*p = '\0';
```
thom311: this line seems unnecessary. Just bail out with error.

clients/cli/nmcli.c
```diff
@@ -331,21 +331,226 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
+			g_set_error (error, NMCLI_ERROR, 0,
+			             _("Expected whitespace following '%s'\n"), name);
+			return FALSE;
+		}
+		while (isblank (*p)) {
+			*p = '\0';
```
thom311: the `*p = '\0';` could go before the loop. Why do you want to clear all spaces, just jump over them.

Btw, there is `nm_str_skip_leading_spaces()`...

clients/cli/nmcli.c
```diff
@@ -331,21 +331,226 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
+
+		/* All good, set the palette entry. */
+		for (i = NM_META_COLOR_NONE + 1; i < _NM_META_COLOR_NUM; i++) {
+			if (strcmp (map[i], name) == 0) {
+				nmc->nmc_config_mutable.palette[i] = sequence;
+				break;
+			}
+		}
+		if (i == _NM_META_COLOR_NUM)
+			g_printerr (_("Ignoring an unrecognized color: '%s'\n"), name);
```
thom311: is this really helpful? Do you expect that NetworkManager package ships such scheme files? Otherwise, if they are packaged not along nmcli, the party that creates the scheme files has problems to target a certain version. Say, nmcli version 1.16.0 adds a new color "something-fancy", then the packager could not add it to the scheme, unless he is really sure that every user of nmcli uses 1.16.0 or newer. I would just ignore unknown names.

In other words, the parsing of scheme should be backward *and* forward compatible and not complain.

clients/cli/nmcli.c
lkundrak: It is useful for the error message that follows.

clients/cli/nmcli.c
lkundrak: Yes, the p++ would also need to be duplicated there. It seems more concise to me not to special-case the first space, and just nul them all instead.

man/nmcli.xml
lkundrak: Whoopise, copypasta exposed.

clients/cli/nmcli.c
lkundrak: Ok.

clients/cli/nmcli.c
lkundrak: I think so. Not sure. In general, when you mess up your configuration file and it doesn't work, this could help you find out what's wrong.

clients/cli/nmcli.c
```diff
@@ -296,9 +295,36 @@ matches_arg (NmCli *nmc, int *argc, char ***argv, const char *pattern, char **ar
...
 }
 
+typedef enum {
+        NMC_USE_COLOR_AUTO,
+        NMC_USE_COLOR_YES,
+        NMC_USE_COLOR_NO,
+} NmcColorOption;
+
+static void
+set_colors (NmCli *nmc, NmcColorOption *color_option)
```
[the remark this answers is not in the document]
lkundrak: Nope. Turned into passing by value.

clients/cli/nmcli.c
lkundrak: > Do you expect that NetworkManager package ships such scheme files?

No. The terminal-colors.d(5) specification only looks in user configuration files, not in /lib.The sensible defaults are hardcoded.

If a file has an error, it should be treated as such. On the other hand user is unlikely to see the warning as soon as the pager is invoked. I only did this because I couldn't think of a better way to communicate an error.

clients/cli/nmcli.c
lkundrak: Well, ideally. There's no suitable g_ascii_is* function for this :(

Added one, because there seem to be multiple spots in the code base where we detect horizontal whitespace.

vhumpa:
rebuild

bengal:
I get the following error when building with meson:


```
@exe/nmcli.c.o' -c ../clients/cli/nmcli.c
../clients/cli/nmcli.c: In function ‘set_colors’:
../clients/cli/nmcli.c:599:55: error: ‘SYSCONFDIR’ undeclared (first use in this function); did you mean ‘MSG_CONFIRM’?
  check_colors_files_for_base_dir (nmc, &color_option, SYSCONFDIR);
                                                       ^~~~~~~~~~
                                                       MSG_CONFIRM
../clients/cli/nmcli.c:599:55: note: each undeclared identifier is reported only once for each function it appears in
[10/178] Compiling C object 'clients/tui/newt/nmt-newt@sta/nmt-newt-section.c.o'.
ninja: build stopped: subcommand failed.
```

Otherwise, LGTM.

clients/cli/nmcli.c
```diff
@@ -51,7 +51,6 @@
 # define NMCLI_VERSION VERSION
```
fgiudici: double "it" in the commit msg... drop a useless 'it' in "drop a useless comment" ;-)

lkundrak:
rebuild

PR #109, opened 2018-04-30
manager: fix assertion in nm_manager_activate_connection()

bengal:
nm_manager_activate_connection() should not require a device to be
passed in for VPN connections because when the argument is NULL the
actual device will be determined by _new_active_connection().

Fixes: 10753c36168a82cd658df8a7da800960fddd78ed
https://bugzilla.redhat.com/show_bug.cgi?id=1570545


thom311:
lgtm

PR #111, opened 2018-05-14
platform: clean up wifi_data for interface when it goes away

lkundrak:
Otherwise we just leak WifiData structures and even pile up open genlink
sockets.

### DIFF
--- a/src/platform/nm-linux-platform.c
+++ b/src/platform/nm-linux-platform.c
@@ -7163,6 +7163,29 @@ finalize (GObject *object)
 }
 
 static void
+link_changed_cb (NMPlatform *platform,
+                 int obj_type_i,
+                 int ifindex,
+                 NMPlatformLink *info,
+                 int change_type_i,
+                 gpointer user_data)
+{
+	const NMPlatformSignalChangeType change_type = change_type_i;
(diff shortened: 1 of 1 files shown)


lkundrak:
I tend to think that how we handle WifiData instances could use some improvement. In particular:

 * A generic netlink socket is open per-device. Perhaps the socket (single one) should be managed by the platform directly.
* wifi-util/WifiData (that is, platform/wifi) is always compiled in. Maybe it should be moved to device/wifi and the nm_platform_wifi_*() abstractions should be just removed. I think there's little potential value in abstracting those in platform and no actual value since the PlatformFake implementations are not used by the tests anyway.

What do you think?

thom311:
Note how the signal is *only* emitted at one place: `nm_platform_cache_update_emit_signal()`
I think a better place to hook the deletion would be around `cache_on_change()` or `nm_platform_cache_update_emit_signal()`.

Also, looking at the code, it seems that a call to `cache_on_change()` and `nm_platform_cache_update_emit_signal()` always goes together. That makes me think, 
 - `cache_on_change()` should call `nm_platform_cache_update_emit_signal()`
 - `cache_on_change()` should handle the cleanup task of WifiData


We could also cache the wifi-data in the private parts in the `NMPObjectLink` structure. It's not unlike the `udev.device` instance.

lkundrak:
Superseded by the lr/6lowpan branch

PR #112, opened 2018-05-18
dispatcher: rework creating dispatcher environment

thom311:
- don't use GSList as intermediate data type to construct
  the environment, especially when all we want is to get
  a strv array at the end. This reverts the order of elements
  compared to previously.

- add and use helper methods _items_add_*() which assert against sensible
  input.

- merge IPv4 and IPv6 implementations for creating the environment.


thom311:
merged as 591cc60a5b456e2f01ae5b5589a59b300b8e1c73

PR #113, opened 2018-05-18
th/ppp-stop

thom311:
Rework stopping of pppd process in NMPPPManager.

Especially regarding how I think we should one day (in the future) handle shutdown of NetworkManager.


bengal:
LGTM

thom311:
Thanks. merged

PR #114, opened 2018-05-21
lr/alternatives: provide alternative ifup/ifdown implementations

lkundrak:
Needs the initscripts part to go in first: https://github.com/fedora-sysv/initscripts/pull/197

https://bugzilla.redhat.com/show_bug.cgi?id=1578436


lkundrak:
> This is not really the same what initscripts do.
...
> Your alternatives implementation does something significantly different. You could argue, it's good enough. But I'd rather do it right.

Yes, you're right.

I'm going to fold this with another change I've been pondering for some time; exposing the filename of file-backed connections to the clients.

I'd really prefer that to using the `com.redhat.ifcfgrh1` undocumented interface. I'd argue that one should not have existed in the first place. Perhaps we could even remove it at some point.

> Also, I don't really like this extending the nmcli binary this way. Can we not instead add a shell script, which gets called as alternative to `ifup` / `ifdown`, calls `get_uuid_by_config`, and then execs `nmcli`? It's not clear to me, why nmcli would be bothered with this Fedora/RHEL specific hack.

I'd not giving up on this easily.

This thing is going to stay there for some time, and folding it to nmcli seems less hacky to me than falling back to shell trickery. Also this allows for a very straightforward implementation with fairly minimal bloat to nmcli (the line noise is negligible and if we care about the few bytes this adds everywhere then we could ifdef it when the ifcfg plugin is enabled).

I'm following up a new version that addresses your first point.

thom311:
>  folding it to nmcli seems less hacky to me than falling back to shell trickery.

"trickery" sounds like there would be a complicated or large shell script. I think this would be doable with a very simple script.

I dislike adding Fedora/RHEL specific code into nmcli. It doesn't belong there.

> I'd really prefer that to using the com.redhat.ifcfgrh1 undocumented interface

it is undocumented, because it only has only one user. I don't think there is a problem with it (given how simple this API is)

Exposing a filename for the profile maps well to keyfile and ifcfg-rh. It works bad with ifupdown (where /etc/network/interfaces contains multiple profiles) and ibft (where there is no filename).
Sure, it might still be a worthy extension to the API, because this is often needed. But it doesn't seem necessary to solve this bug. Also, `com.redhat.ifcfgrh1` is so simple, you can just call it with dbus-send. If you expose the filename in the NMRemoteSetting, the user needs to load all profiles, and search them for the right path. That essentially means, you can no longer write a simple shell-script, but need NMClient and libnm. Then indeed, you can no longer write a simple ifup/ifdown alternatives script.

libnm/nm-remote-connection.c
```diff
@@ -669,6 +671,22 @@ nm_remote_connection_get_flags (NMRemoteConnection *connection)
...
+ * @connection: the #NMRemoteConnection
+ *
+ * Returns: file that stores the connection in case the connection is file-backed.
+ *
+ * Since: 1.12
+ **/
+const char *
+nm_remote_connection_get_filename (NMRemoteConnection *connection)
+{
+	g_return_val_if_fail (NM_IS_REMOTE_CONNECTION (connection), FALSE);
```
bengal: It should return NULL on error.

bengal:
> Exposing a filename for the profile maps well to keyfile and ifcfg-rh. It works bad with ifupdown (where /etc/network/interfaces contains multiple profiles) and ibft (where there is no filename).
> Sure, it might still be a worthy extension to the API, because this is often needed. But it doesn't seem necessary to solve this bug.

I think the filename property is useful (actually, I wondered many times why such property didn't exist yet).

Regarding "cli: allow being called as ifup/ifdown", while I like the idea of avoiding wrapper scripts, I agree with Thomas that it should not belong to nmcli. If the code was more generic and didn't have the hardcoded paths:

 "%s/ifcfg-%s", SYSCONFDIR "/sysconfig/network-scripts"

I think it would have been ok. But since those are required to emulate ifup/ifdown, I prefer if we implement this in a script.

libnm/nm-remote-connection.c
lkundrak: Good catch. Fixed.

lkundrak:
> I think it would have been ok. But since those are required to emulate ifup/ifdown, I prefer if we implement this in a script.

Fair enough, thanks. Pushed an updated implementation.

Verified to work fine on Rawhide.

bengal:
> Initscripts issue first the equivalent to nmcli connection load (nm_con_load). The wrapper script should do that as well. Note that it just calls D-Bus directly, no need to use nmcli for that.

+1.

Even if it is ugly to call dbus-send in a script, it is faster than nmcli, which needs to load the whole object tree on startup and thus it could take some seconds on systems with many connections/devices, slowing down scripts where ifup/ifdown are called multiple times. If we replace dbus-send with nmcli, there will be a regression in performances. 

To fix this, we could modify nmcli to not instantiate a NMClient for the 'connection load/reload' commands and only issue a direct D-Bus call. Or just use dbus-send.

bengal:
The rest LGTM.

PR #116, opened 2018-05-21
lr/initrd: support running from initrd

lkundrak:
:poodle:


thom311:
> [WIP] keyfile: write in-memory connections to /run

aside what I already said, it's also that currently, if you restart NetworkManager, in-memory connections are lost. With this patch, they wouldn't be lost until reboot. I don't say that one or the other behavior is better, but at least it's a significant change in behavior, something that should be addressed very carefully.

lkundrak:
Note to self: This needs a `NEWS` file entry.

vbenes:
Crash was caught here:
https://faf.lab.eng.brq.redhat.com/faf/problems/bthash/?bth=fda9404a54486bc14101b28058b59c0e941a0332

src/dhcp/nm-dhcp-systemd.c
```diff
@@ -432,6 +433,13 @@ lease_to_ip4_config (NMDedupMultiIndex *multi_idx,
 		add_option (options, dhcp4_requests, SD_DHCP_OPTION_NTP_SERVER, str->str);
 	}
 
+	/* Root path */
+	r = sd_dhcp_lease_get_root_path (lease, &s);
+	if (r == 0) {
```
thom311: systemd convention is to check for error/success via `r < 0` / `r >= 0`.

src/nm-config.c
```diff
@@ -2341,7 +2341,11 @@ nm_config_reload (NMConfig *self, NMConfigChangeFlags reload_flags)
 		                     (const char *const*) priv->atomic_section_prefixes, NULL);
 	}
 
-	new_data = nm_config_data_new (config_main_file, config_description, (const char *const*) no_auto_default, keyfile, keyfile_intern);
+	new_data = nm_config_data_new (config_main_file,
+	                               config_description,
+	                               (const char *const*) no_auto_default,
+	                               keyfile,
+	                               keyfile_intern);
```
thom311: I like this!! Recently, I pretty much prefer to have just one argument per-line. It seems easier to read to me.

src/dns/nm-dns-dnsmasq.c
```diff
@@ -356,22 +354,16 @@ start_dnsmasq (NMDnsDnsmasq *self)
...
-	g_dbus_proxy_new (connection,
-	                  G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START,
-	                  NULL,
-	                  DNSMASQ_DBUS_SERVICE,
-	                  DNSMASQ_DBUS_PATH,
-	                  DNSMASQ_DBUS_SERVICE,
-	                  priv->dnsmasq_cancellable,
-	                  dnsmasq_proxy_cb,
-	                  self);
+	g_dbus_proxy_new_for_bus (G_BUS_TYPE_SYSTEM,
```
thom311: why is this better? Previously, at least `NMDBusManager` was the authority regarding our D-Bus session.

Ok, we use `G_BUS_TYPE_SYSTEM` **all** over the place. So, in that spirit, the patch is certainly correct, and the commit message could just say: "like we do at ~30 other places."

I still think `NMDBusManager` should have a say which D-Bus connection to use, but anyway, the patch doesn't change that.

src/settings/plugins/keyfile/nms-keyfile-plugin.c
```diff
@@ -175,13 +175,30 @@ update_connection (NMSKeyfilePlugin *self,
...
 	if (full_path)
 		_LOGD ("loading from file \"%s\"...", full_path);
 
+	if (g_str_has_prefix (full_path, nms_keyfile_utils_get_path ())) {
+		dir_len = strlen (nms_keyfile_utils_get_path ());
+	} else if (g_str_has_prefix (full_path, NM_CONFIG_KEYFILE_PATH_IN_MEMORY)) {
+		dir_len = NM_STRLEN (NM_CONFIG_KEYFILE_PATH_IN_MEMORY);
+	} else {
+		/* Just make sure the file name is not going go pass the following check. */
+		dir_len = strlen (full_path);
```
thom311: This condition seems so complicated to me, and does't cover special cases like filenames ".", "..".

Can you not add a helper function 

```C
gboolean
nm_utils_path_is_in_dir (const char *path, const char *dir)
{
    gsize dir_len;

    dir_len = strlen (dir);
    while (   dir_len > 0
           && dir[dir_len - 1] == '/')
        dir_len--;
    if (strncmp (path, dir, dir_len) != 0)
        return FALSE;
    path += dir_len;
    if (path != '/')
        return FALSE;
    if (strchr (path, '/'))
        return FALSE;
    if (   path[0] == '.'
        && (   path[1] == '\0'
            || (path[1] == '.' && path[2] == '\0')))
        return FALSE;
    return TRUE; 
}
```

Actually, this also does not properly handle repated "//" in `dir`. But it's a start.

src/settings/plugins/keyfile/nms-keyfile-plugin.c
```diff
@@ -412,39 +429,49 @@ _sort_paths (const char **f1, const char **f2, GHashTable *paths)
...
-	GPtrArray *filenames;
-	GHashTable *paths;
+	GError *error = NULL;
 
-	dir = g_dir_open (nms_keyfile_utils_get_path (), 0, &error);
+	dir = g_dir_open (path, 0, &error);
 	if (!dir) {
 		_LOGW ("cannot read directory '%s': %s",
-		             nms_keyfile_utils_get_path (),
+		             path,
```
thom311: indentation

thom311: Is this warning at all helpful? Now that we have multiple directories, it's not at all clear, that they all *must* exist.

src/settings/plugins/keyfile/nms-keyfile-plugin.c
```diff
@@ -412,39 +429,49 @@ _sort_paths (const char **f1, const char **f2, GHashTable *paths)
...
 	}
 
-	alive_connections = g_hash_table_new (nm_direct_hash, NULL);
-
-	filenames = g_ptr_array_new_with_free_func (g_free);
 	while ((item = g_dir_read_name (dir))) {
 		if (nms_keyfile_utils_should_ignore_file (item))
 			continue;
-		g_ptr_array_add (filenames, g_build_filename (nms_keyfile_utils_get_path (), item, NULL));
+		g_ptr_array_add (filenames, g_build_filename (path, item, NULL));
```
thom311: in /etc we accept almost all files (sans same extensions via `nms_keyfile_utils_should_ignore_file()`. I always thought that is a mistake, and we should only load files with a particular extension (and write them as such). We cannot change that for /etc, but can we do that for /run? (read and write).

src/settings/plugins/keyfile/nms-keyfile-plugin.c
```diff
@@ -533,16 +554,16 @@ add_connection (NMSettingsPlugin *config,
...
+	if (!nms_keyfile_writer_connection (connection,
+	                                    save_to_disk,
+	                                    NULL,
+	                                    FALSE,
+	                                    &path,
+	                                    &reread,
+	                                    NULL,
+	                                    error))
+		return NULL;
+
```
bengal: I wonder, this works only if the user doesn't have another settings plugin enabled, right?
If, for example, user has 'plugins=ifcfg-rh' in the configuration, the keyfile add_connection() is not called and the connection file is not written into /run. Is this by design?

src/nm-config.c
```diff
@@ -509,34 +517,72 @@ nm_config_cmd_line_options_free (NMConfigCmdLineOptions *cli)
...
+		{ "configure-and-quit", 0, G_OPTION_FLAG_OPTIONAL_ARG, G_OPTION_ARG_CALLBACK, parse_configure_and_quit, N_("Quit after initial configuration"), NULL },
+		{ "debug", 'd', 0, G_OPTION_ARG_NONE, &cli->is_debug, N_("Don't become a daemon, and log to stderr"), NULL },
+
+			/* These three are hidden for now, and should eventually just go away. */
+		{ "connectivity-uri", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_STRING, &cli->connectivity_uri, N_("An http(s) address for checking internet connectivity"), "http://example.com" },
+		{ "connectivity-interval", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_INT, &cli->connectivity_interval, N_("The interval between connectivity checks (in seconds)"), G_STRINGIFY (NM_CONFIG_DEFAULT_CONNECTIVITY_INTERVAL) },
+		{ "connectivity-response", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_STRING, &cli->connectivity_response, N_("The expected start of the response"), NM_CONFIG_DEFAULT_CONNECTIVITY_RESPONSE },
+		{ 0 },
+	};
+
```
bengal: Can you also update the man page?

src/settings/plugins/keyfile/nms-keyfile-plugin.c
thom311: No, ifcfg-rh does too. Though, that maybe should be avoided: https://bugzilla.gnome.org/show_bug.cgi?id=744711

lkundrak: In any case, I think the current behavior is good. I don't think we want the ifcfg plugin to read or write connections in /run.

PR #117, opened 2018-05-22
th/per-host-stable-id

thom311:
- extend stable-id to support a per-host specifier (`${HOST}`) and a per-device specifier (`${DEVICE}`)
- fix `ipv4.dhcp-client-id=stable` to hash also the per-host secret key. Otherwise, different hosts generate the same client-id, which is not desirable.


src/nm-core-utils.c
```diff
@@ -2797,57 +2797,97 @@ nm_utils_file_get_contents (int dirfd,
```
fgiudici: [...] "In case of error, *the we* won't be persisted" [...]
The commit msg probably has a typo.

src/nm-core-utils.c
```diff
@@ -2797,57 +2797,97 @@ nm_utils_file_get_contents (int dirfd,
...
+nm_utils_secret_key_get (const guint8 **out_secret_key,
+                         gsize *out_key_len)
+{
+	static volatile const SecretKeyData *p_secret_key;
+	const SecretKeyData *secret_key;
+
+	secret_key = g_atomic_pointer_get (&p_secret_key);
+	if (G_UNLIKELY (!secret_key)) {
+		static gsize init_value = 0;
+		static SecretKeyData p_secret_key_data;
```
fgiudici: Just a question: what those "p_" and "d_" prefixes to "secret_data" vars stands for?

src/nm-core-utils.h
```diff
@@ -283,7 +283,10 @@ gboolean nm_utils_file_set_contents (const gchar *filename,
 char *nm_utils_machine_id_read (void);
 gboolean nm_utils_machine_id_parse (const char *id_str, /*uuid_t*/ guchar *out_uuid);
 
-guint8 *nm_utils_secret_key_read (gsize *out_key_len, GError **error);
+gboolean nm_utils_secret_key_read (guint8 **out_secret_key,
```
fgiudici: Now that we have "nm_utils_secret_key_get", do we still need to expose in the cure header file "nm_utils_secret_key_read" or can we just make it a static function?

bengal:
> core: let nm_utils_secret_key_read() handle failures internally

persistant -> persistent (2 times)

Can you split long lines?

If we fail to read the key or obtain enough random data, we never
retry later as the previous implementation did. Do you think is ok?

> all: add stable-id specifiers "${HOST}" and "${DEVICE}"
 >  ... "While arguably "${HOST}" is (soon) anyway implicitly used everywhere," ...

Isn't it misleading to add a ${HOST} key when it's is already used anyway?

thom311:
> If we fail to read the key or obtain enough random data, we never retry later as the previous implementation did. Do you think is ok?

I think that is OK, because it seems worse to change the key later on. NM cannot guarantee that the key can be persisted or has good randomness. In that case, something is very wrong to start with, and possibly the admin needs to fix it (mount /var rw, seed kernel's random generator before starting NM). However, in the spirit of handling the unsolvable problem as best as possible, it seems safer to stick to the bad key, instead of re-loading a different one.

bengal:
> I think that is OK, because it seems worse to change the key later on

Makes sense.

LGTM now.

thom311:
mergd

PR #118, opened 2018-05-23
rpm: prefer python3

lkundrak:
From Fedora 28 on we can build without Python 2. That is good,
because it's eventually going to be removed.

Based on a change in Fedora dist-git by Iryna Shcherbina.

### DIFF
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -169,8 +169,15 @@ BuildRequires: ModemManager-glib-devel >= 1.0
 BuildRequires: newt-devel
 %endif
 BuildRequires: /usr/bin/dbus-launch
+%if 0%{?fedora} > 27 || 0%{?rhel} > 7
+BuildRequires: python3
+BuildRequires: python3-gobject-base
+BuildRequires: python3-dbus
+%else
+BuildRequires: python2
 BuildRequires: pygobject3-base
 BuildRequires: dbus-python
+%endif
 BuildRequires: libselinux-devel
 BuildRequires: polkit-devel
 BuildRequires: jansson-devel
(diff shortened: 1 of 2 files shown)


thom311:
```
+BuildRequires: python3
+BuildRequires: python3-gobject-base
+BuildRequires: python3-dbus
```

Maybe add these packages to `contrib/fedora/REQUIRED_PACKAGES`? I actually use this file, whenever I have a new VM to easily all package for NM development I need. Yes, it installs more then strictly necessary, but I don't mind.

lkundrak:
> +BuildRequires: python3
> +BuildRequires: python3-gobject-base
> +BuildRequires: python3-dbus
>
>> Maybe add these packages to contrib/fedora/REQUIRED_PACKAGES? I actually use this file, whenever I have a new VM to easily all package for NM development I need. Yes, it installs more then strictly necessary, but I don't mind.

Well, the flat package list in REQUIRED_PACKAGES can't be reasonably expected to install Python 3 packages in some environments while falling back to Python 2 versions elsewhere. No big deal, it's still going to work with Python 2 versions (until it's actually dropped).

thom311:
I assume and I think that this especially covers [the change to rawhide](https://src.fedoraproject.org/rpms/NetworkManager/c/4797640ddf916f33de1af333796d6d8a007893cc?branch=master), so that when we re-import contrib/rpm's spec file into rawhide, those changes are not effectively reverted.

Agree?

thom311:
> rpm/build.sh: suggest dnf builddep after BUILDTYPE=SRPM 
very nice!

lkundrak:
> I assume and I think that this especially covers the change to rawhide

Well, the last line of commit message is supposed to imply that.

lkundrak:
Merged.

PR #119, opened 2018-05-24
lr/setting-plugin-props-cleanup: drop properties from setting plugin

lkundrak:
The initial motivation was to get rid of the ugly logging on a plugin load. Turns out all of the properties are straight out useless. This saves quite some line noise.

### DIFF
--- a/src/settings/nm-settings.c
+++ b/src/settings/nm-settings.c
@@ -632,29 +632,22 @@ add_plugin (NMSettings *self, NMSettingsPlugin *plugin)
 	return TRUE;
 }
 
-static GObject *
-find_plugin (GSList *list, const char *pname)
+static gboolean
+plugin_loaded (GSList *list, const char *path)
 {
 	GSList *iter;
-	GObject *obj = NULL;
 
-	g_return_val_if_fail (pname != NULL, NULL);
+	g_return_val_if_fail (path != NULL, TRUE);
 
-	for (iter = list; iter && !obj; iter = g_slist_next (iter)) {
-		NMSettingsPlugin *plugin = NM_SETTINGS_PLUGIN (iter->data);
-		char *list_pname = NULL;
+	for (iter = list; iter; iter = g_slist_next (iter)) {
+		const char *list_path = g_object_get_qdata (G_OBJECT (iter->data),
+		                                            plugin_module_path_quark ());
 
-		g_object_get (G_OBJECT (plugin),
-		              NM_SETTINGS_PLUGIN_NAME,
-		              &list_pname,
-		              NULL);
-		if (list_pname && !strcmp (pname, list_pname))
-			obj = G_OBJECT (plugin);
-
-		g_free (list_pname);
+		if (g_strcmp0 (path, list_path) == 0)
+			return TRUE;
 	}
 
-	return obj;
+	return FALSE;
 }
 
 static gboolean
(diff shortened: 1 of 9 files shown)


src/settings/nm-settings.c
```diff
@@ -721,21 +716,19 @@ load_plugins (NMSettings *self, const char **plugins, GError **error)
...
 		}
 
-		if (find_plugin (list, pname))
+		full_name = g_strdup_printf ("nm-settings-plugin-%s", pname);
+		path = g_module_build_path (NMPLUGINDIR, full_name);
+
+		if (plugin_loaded (list, path))
 			continue;
 
 load_plugin:
```
thom311: this label should be moved before
```
full_name = g_strdup_printf ("nm-settings-plugin-%s", pname);
```
Also, make sure that we don't leak `full_name` nor `path`. That means, use `nm_clear_g_free()`, either immediately after `load_plugin:` label, or immediately before `goto load_plugin;`.

If you dislike that this is handled via goto, you can also refactor the code to solve it somehow differently. But I doubt that would be much nicer...

src/settings/nm-settings.c
```diff
@@ -535,11 +535,13 @@ get_plugin (NMSettings *self, NMSettingsPluginCapabilities capability)
 
 	/* Do any of the plugins support the given capability? */
 	for (iter = priv->plugins; iter; iter = iter->next) {
-		NMSettingsPluginCapabilities caps = NM_SETTINGS_PLUGIN_CAP_NONE;
+		NMSettingsPlugin *plugin = NM_SETTINGS_PLUGIN (iter->data);
+
+		if (has_add_connection != TRUE)
```
thom311: can we avoid comparing boolean values `!=` and `==` with `TRUE|FALSE`?

Also, at several places the branch compares pointers against NULL, which I think should be avoided. But I care less about pointer comparisons than about booleans. On the contrary, I think integers should be compared explicitly against `0`, not just `if (value)`. Yes, we are not very consistent about all this.

lkundrak:
> this label should be moved before
> ```
> full_name = g_strdup_printf ("nm-settings-plugin-%s", pname);
> ```
> Also, make sure that we don't leak `full_name` nor `path`. That means, use `nm_clear_g_free()`, either immediately after `load_plugin:` label, or immediately before `goto load_plugin;`.
>
> If you dislike that this is handled via goto, you can also refactor the code to solve it somehow differently. But I doubt that would be much nicer...

Yes, I think the `goto` makes it rather difficult to follow that's going on.

I made an attempt to streamline things. The odd thing is that `load_plugin()` returns an error only on fatal-errors, logs a warning and happily returns success on a non-fatal one. I guess it's still better than it used to

> can we avoid comparing boolean values != and == with TRUE|FALSE?

yes.

> Also, at several places the branch compares pointers against NULL, which I think should be avoided.

I disagree hysterically and violently. The explicit `== NULL` is actually a documentation that helps inferring what type are we dealing with.

src/settings/nm-settings.c
```diff
@@ -666,6 +666,77 @@ find_plugin (GSList *list, const char *pname)
...
+	/* after accessing the plugin we cannot unload it anymore, because the glib
+	 * types cannot be properly unregistered. */
+	g_module_make_resident (plugin);
+
+	obj = (*factory_func) ();
+	if (!obj || !NM_IS_SETTINGS_PLUGIN (obj)) {
+		g_set_error (error, NM_SETTINGS_ERROR, NM_SETTINGS_ERROR_FAILED,
+		             "Plugin '%s' returned invalid system config object.",
+		             pname);
+		return FALSE;
```
thom311: leaks `obj`. I think, `gs_unref_object` in combination with `g_steal_pointer (&obj)` below is nicer, because that way, ownership is always explicitly somewhere (at `obj` auto variable), and explicitly transferred (stolen).

lkundrak: Agreed.

lkundrak:
Merged, with the leak fix as mentioned above.

PR #120, opened 2018-05-24
lr/6lowpan: add IEEE 802.15.4 and 6LoWPAN support

lkundrak:
### Overall picture

* IEEE 802.15.4 ("WPAN" from now on) defines the bottom 2 layers (PHY and MAC) for a Wireless Personal Area Network. IoT devices use it with various open or proprietary protocols layered on top. Because of its small MTU (of 123) it is not capable of directly transporting IP. The devices are grouped in a network using a 16-bit network ID ("PAN ID"). They are identified using a globally unique 64-bit hardware address or optionally using a 16-bit "Short Address" unique to the network.
* 6LoWPAN encapsulates IPv6 into smaller fragments suitable for transmission over IEEE 802.15.4 (and perhaps Bluetooth? Haven't looked into that)

In Linux both are separate links. WPAN links are configured via a Generic Netlink protocol nl802154 that is obviously modelled is a fashion similar to Wi-Fi. The 6LoWPAN devices layer on top of them much like VLANs of MACSec.

### Caveats

* Changes to WPAN properties are not sent to any of the nl802154 multicast groups (there's just "config" one that basically just announces new links, perhaps redundantly to rtnetlink). Thus we don't learn about external changes to the PAN ID, or Short Address. nl80211 seems to have the same problem. Perhaps this could be easily addressed.
* 6LoWPAN NEWLINK messages seem to lack IFLA_LINK thus we use a rather kludgy way of determining the parent. This may need a kernel fix in the long run.
* Both the WPAN and the 6LoWPAN devices need to be down for changes to WPAN properties to be allowed. This is inconvenient, perhaps WPAN down should be sufficient while 6LoWPAN should just go !IFF_LOWER_UP without us doing anything.
* Kernel doesn't install the relevant headers, despite they're part of uapi and even suggest they're public in comments. An easyfix i guess.

### Testing

The `fakelb` driver can emulate a pair (or more) of WPAN devices:

```
# nmcli c add type wpan ifname wpan0 pan-id 0x8086 short-addr 0x1337
Connection 'wpan-wpan0' (58dc04f8-0ff9-4ec5-94a7-099743560fd9) successfully added.
# nmcli c add type 6lowpan ifname lowpan0 dev wpan0 ipv6.method link-local
Connection '6lowpan-lowpan0' (1f3433d9-afbc-451e-80f9-7ee7be217b90) successfully added.
# modprobe fakelb
# iwpan dev wpan0 info
Interface wpan0
        ifindex 109
        wpan_dev 0x3000000001
        extended_addr 0x8ec07badf2d01bc2
        short_addr 0x1337
        pan_id 0x8086
        type node
        max_frame_retries 3
        min_be 3
        max_be 5
        max_csma_backoffs 4
        lbt 0
        ackreq_default 0
# 
```

Openlabs sells $10 radios that plugs nicely into a Raspberry Pi: http://openlabs.co/store/Raspberry-Pi-802.15.4-radio
Making that work is basically just a matter of ```echo 'dtoverlay=at86rf233' >>/boot/config.txt && reboot```


src/nm-core-utils.c
```diff
@@ -2962,7 +2962,12 @@ nm_utils_get_ipv6_interface_identifier (NMLinkType link_type,
 		memcpy (out_iid->id_u8 + 4, &addr, 4);
 		return TRUE;
 	default:
-		if (hwaddr_len == ETH_ALEN) {
+		if (hwaddr_len == G_N_ELEMENTS (out_iid->id_u8)) {
```
thom311: this essentially detects the MAC address type, based on the length. We already do that sometimes, to distinguish ethernet vs. infiniband. I find that really ugly, because we should always explicitly know which layer2 type we have when we operate on a MAC, not detect it based on the MAC's length. The function `nm_utils_get_ipv6_interface_identifier()` already makes some decisions based on the link-type argument. Can we not instead add a `case NM_LINK_TYPE_WPAN:`?

src/devices/nm-device.c
```diff
@@ -1743,9 +1746,36 @@ get_ip_iface_identifier (NMDevice *self, NMUtilsIPv6IfaceId *out_iid)
...
+	if (pllink->type == NM_LINK_TYPE_6LOWPAN) {
+		/* If the underlying IEEE 802.15.4 device has a short address we generate
+		 * a "pseudo 48-bit address" that's to be used in the same fashion as a
+		 * wired Ethernet address. The mechanism is specified in Section 6. of
+		 * RFC 4944 */
+		guint16 pan_id;
+		guint16 short_addr;
+		guint8 pseudo_hwaddr[ETH_ALEN];
+
+		short_addr = nm_platform_wpan_get_short_addr (platform, pllink->parent);
```
thom311: this not only looks into the platform cache, it make a genl request. I dislike the idea, that in order to generate the interface-identifyer, we need to ask kernel. I guess, that cannot easily be avoided, can it? In the pull-request summary, you alluded to the fact that we cannot cache the value, because we don't get change notifications via rtnetlink.

src/devices/nm-device.c
```diff
@@ -1743,9 +1746,36 @@ get_ip_iface_identifier (NMDevice *self, NMUtilsIPv6IfaceId *out_iid)
...
+		if (short_addr != G_MAXUINT16) {
+			pan_id = nm_platform_wpan_get_pan_id (platform, pllink->parent);
+			pseudo_hwaddr[0] = short_addr & 0xff;
+			pseudo_hwaddr[1] = (short_addr >> 8) & 0xff;
+			pseudo_hwaddr[2] = 0;
+			pseudo_hwaddr[3] = 0;
+			pseudo_hwaddr[4] = pan_id & 0xff;
+			pseudo_hwaddr[5] = (pan_id >> 8) & 0xff;
+
+			hwaddr = pseudo_hwaddr;
```
thom311: this is not strictly speaking safe, because `pseudo_hwaddr` is declared inside a block, and you use it later (although, compile would usually not reclaim nor reuse the stack memory at that point). Just declare pseudo_hwaddr at beginning of function.

src/platform/nm-platform.c
```diff
@@ -2088,6 +2088,71 @@ nm_platform_link_tun_add (NMPlatform *self,
...
+	/* As of 4.16 kernel does not expose the peer_ifindex as IFA_LINK.
+	 * Find the WPAN device with the same MAC address. */
+	if (out_parent) {
+		const NMPlatformLink *parent_plink;
+
+		parent_plink = nm_platform_link_get_by_address (self, NM_LINK_TYPE_WPAN,
+		                                                plink->addr.data,
+		                                                plink->addr.len);
+		if (parent_plink)
+			NM_SET_OUT (out_parent, parent_plink->ifindex);
```
thom311: might leave "out_parent" un-initialized. At least, when returning TRUE, we should always init the out-argument, so that callers are free to pass an uninitialized out-argument.

src/platform/nm-linux-platform.c
```diff
@@ -5751,6 +5751,43 @@ link_vxlan_add (NMPlatform *platform,
 	g_return_val_if_reached (FALSE);
 }
 
+static gboolean
+link_6lowpan_add (NMPlatform *platform,
```
thom311: commit could also be squashed in parent commit...

src/platform/nm-linux-platform.c
```diff
@@ -5751,6 +5751,43 @@ link_vxlan_add (NMPlatform *platform,
...
+		goto nla_put_failure;
+
+	NLA_PUT_STRING (nlmsg, IFLA_INFO_KIND, "lowpan");
+
+	nla_nest_end (nlmsg, info);
+
+	return do_add_link_with_lookup (platform,
+	                                NM_LINK_TYPE_6LOWPAN,
+	                                name, nlmsg, out_link);
+nla_put_failure:
```
thom311: now, that we have our own `nm-netlink.c` implementation, there is really no reason why `NLA_PUT*()` can fail. TODO: make them non-failable, and get rid of these `nla_put_failure` labels.

src/devices/nm-device-6lowpan.c
```diff
@@ -0,0 +1,371 @@
...
+/*****************************************************************************/
+
+static void
+nm_device_6lowpan_init (NMDevice6Lowpan *self)
+{
+}
+
+static const NMDBusInterfaceInfoExtended interface_info_device_6lowpan = {
+	.parent = NM_DEFINE_GDBUS_INTERFACE_INFO_INIT (
+		NM_DBUS_INTERFACE_DEVICE_6LOWPAN,
```
thom311: Lacks D-Bus Property "Parent". Don't make it a legacy property (meaning: don't use `NM_DEFINE_DBUS_PROPERTY_INFO_EXTENDED_READABLE_L`).

libnm-core/nm-keyfile-utils.c
```diff
@@ -38,6 +38,7 @@ static const SettingAlias alias_list[] = {
 	{ NM_SETTING_WIRED_SETTING_NAME, "ethernet" },
 	{ NM_SETTING_WIRELESS_SETTING_NAME, "wifi" },
 	{ NM_SETTING_WIRELESS_SECURITY_SETTING_NAME, "wifi-security" },
+	{ NM_SETTING_WPAN_SETTING_NAME, "wpan" },
```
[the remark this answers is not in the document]
thom311: Or maybe just `#define NM_SETTING_WPAN_SETTING_NAME "wpan"`? Is "802-15-4-wpan" really a good name at all? If we feel the alias is worth it, then it seems actually the be the better name. E.g. NM_SETTING_VLAN_SETTING_NAME is also just "vlan", not "802-1q-vlan"

lkundrak:
Note to myself: This needs a `NEWS` file entry.

libnm-core/nm-setting-wpan.c
```diff
@@ -0,0 +1,241 @@
...
+	NMSettingWpanPrivate *priv = NM_SETTING_WPAN_GET_PRIVATE (setting);
+
+	priv->short_address = G_MAXUINT16;
+	priv->pan_id = G_MAXUINT16;
+}
+
+/* Ideally we'll be able to get this from a public header. */
+#ifndef IEEE802154_ADDR_LEN
+#define IEEE802154_ADDR_LEN 8
+#endif
```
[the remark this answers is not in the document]
lkundrak: Well, I'd prefer not. I hope this to be get available from a kernel header at some point.

libnm-core/nm-setting-wpan.c
```diff
@@ -0,0 +1,241 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ * Copyright 2018 Lubomir Rintel <lkundrak@v3.sk>
```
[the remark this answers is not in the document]
lkundrak: Yes, no need to do anything about wpan connections in ifcfg-rh.

libnm-core/nm-keyfile-utils.c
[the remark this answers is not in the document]
lkundrak: At the risk of getting into name bikeshedding -- it seems good enough to me. Consistent with that Wi-Fi does.

clients/common/nm-meta-setting-desc.c
```diff
@@ -1200,6 +1200,11 @@ _set_fcn_gobject_mtu (ARGS_SET_FCN)
 	return TRUE;
 }
 
+/* Ideally we'll be able to get this from a public header. */
+#ifndef IEEE802154_ADDR_LEN
+#define IEEE802154_ADDR_LEN 8
```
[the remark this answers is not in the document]
lkundrak: No, it has no place in a public header.

src/devices/nm-device.c
lkundrak: I think fixing this would require kernel change.

src/platform/wpan/nm-wpan-utils.c
```diff
@@ -0,0 +1,286 @@
...
+	NLA_PUT_U16 (msg, NL802154_ATTR_PAN_ID, htole16 (pan_id));
+	err = nl802154_send_and_recv (self, msg, NULL, NULL);
+	return err >= 0;
+
+nla_put_failure:
+	return FALSE;
+}
+
+guint16
+nm_wpan_utils_get_short_addr (NMWpanUtils *self)
```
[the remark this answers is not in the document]
thom311: Note that most functions named `nm_platform_${TYPE}_get_${SOMETHING}()` only look into the platform cache -

-- yes, `nm_platform_wifi_get_*()` and `nm_platform_mesh_get_*()` are inconsistent.

Wouldn't a naming scheme `nm_platform_${TYPE}_fetch_${SOMETHING}()` be better to indicate functions that get the information fresh from kernel? I think these two kinds are fundamentally different and the name should reflect that. A get-function that only looks into the cache, will always return the same results (unless the cache changes). But multiple calls to fetch can yield different results and imply a larger overhead.

Compare that also to `nm_platform_link_infiniband_get_properties()`, `nm_platform_link_veth_get_properties()`, and `nm_platform_link_tun_get_properties()`. All these functions first try to look into the platform cache, and if kernel has no support, it uses alternative kernel API to fetch the properties (e.g. sysctl). For that reason,  `nm_platform_link_6lowpan_get_properties()` seems misnamed to me. Well, sure, it also has a fallback, but it still relies entirely on the platform cache to perform the lookup. Maybe just call it `nm_platform_link_6lowpan_get_parent()` and return the ifindex (not boolean)?

src/devices/nm-device-6lowpan.c
```diff
@@ -0,0 +1,375 @@
...
+	      const NMPlatformLink *pllink)
+{
+	NMDevice6Lowpan *self = NM_DEVICE_6LOWPAN (device);
+	int parent = 0;
+	int ifindex;
+
+	NM_DEVICE_CLASS (nm_device_6lowpan_parent_class)->link_changed (device, pllink);
+
+	ifindex = nm_device_get_ifindex (device);
+	if (!nm_platform_link_6lowpan_get_properties (nm_device_get_platform (device), ifindex, &parent)) {
```
thom311: Why would this fail? Is it worth a warning?


I think this should be just:
```
     nm_device_parent_set_ifindex (device, 
                                   nm_platform_link_6lowpan_get_parent (nm_device_get_platform (device),
                                                                        nm_device_get_ifindex (device)));
```
Note that `nm_device_parent_set_ifindex()` is fine with unsetting the parent ifindex (in case, get_parent() cannot find the information).

src/platform/nm-linux-platform.c
```diff
@@ -5783,6 +5783,43 @@ link_vxlan_add (NMPlatform *platform,
...
+static gboolean
+link_6lowpan_add (NMPlatform *platform,
+                  const char *name,
+                  int parent,
+                  const NMPlatformLink **out_link)
+{
+	nm_auto_nlmsg struct nl_msg *nlmsg = NULL;
+	struct nlattr *info;
+
+	_LOGD ("adding 6lowpan '%s' parent %u", name, parent);
```
thom311: this results in duplicate logging lines. See "platform: avoid double log messages" commit on bg/sriov-vf-rh1555013 branch.

libnm-core/nm-setting-6lowpan.h
```diff
@@ -0,0 +1,68 @@
...
+struct _NMSetting6Lowpan {
+	NMSetting parent;
+};
+
+typedef struct {
+	NMSettingClass parent;
+
+	/*< private >*/
+	gpointer padding[4];
+} NMSetting6LowpanClass;
```
thom311: these structs should be private (opaque types in the source file). No need for padding.

NEWS
```diff
@@ -1,3 +1,10 @@
+=============================================
+NetworkManager-1.?? (not released yet)
+Overview of changes since NetworkManager-1.12
+=============================================
```
thom311: the branch still contains some "Since: 1.12" and `NM_AVAILABLE_IN_1_12` tags that need updating.

thom311:
there are CI failures, maybe due to not dist-ing the linux header.

src/platform/tests/test-general.c
```diff
@@ -57,7 +57,7 @@ NMTST_DEFINE ();
 int
 main (int argc, char **argv)
 {
-	nmtst_init_assert_logging (&argc, &argv, "INFO", "DEFAULT");
+	nmtst_init_assert_logging (&argc, &argv, "WARN", "DEFAULT");
```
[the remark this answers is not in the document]
lkundrak: That is because we now instantiate the NMWifiUtils a bit earlier, and it logs an INFO line when created. This seems like a reasonable thing to happen and just bumping the test log levels (as other tests do) seems like a more sensible thing to do than asserting on the log. I just can't figure out a good way to document why I changed this.

src/platform/wpan/nm-wpan-utils.c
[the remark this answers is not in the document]
lkundrak: > note how this code is only called at one place:
> ```
>    short_addr = nm_platform_wpan_get_short_addr (platform, pllink->parent);
>    if (short_addr != G_MAXUINT16) {
>        pan_id = nm_platform_wpan_get_pan_id (platform, pllink->parent);
> ```
> so, it always fetches both. If you would add API like
> ```
> gboolean nm_platform_wpan_fetch_details (NMPlatform *, int ifindex *, guint16 *out_id, guint16 *out_short_addr);
> ```
> it would be less code, but also avoid fetching the same information twice.
 
Yes. On the other hand, the Wi-Fi code has sort of the same problem (there it is for the sake of the legacy WEXT implementation). Also, the separation makes sense for the setters, and having getters consistent with setters makes things looks lest confusing overall.
 
If there's a good reason to address this, then the appropriate fix belongs to the kernel -- it should multicast the property changes, we'd just cache them the way we cache most properties and the getters would just fetch things from the cache. (This also applies to Wi-Fi.)
 
> Note that most functions named `nm_platform_${TYPE}_get_${SOMETHING}()` only look into the platform cache -
>
> -- yes, `nm_platform_wifi_get_*()` and `nm_platform_mesh_get_*()` are inconsistent.
>
> Wouldn't a naming scheme `nm_platform_${TYPE}_fetch_${SOMETHING}()` be better to indicate functions that get the information fresh from kernel? I think these two kinds are fundamentally different and the name should reflect that. A get-function that only looks into the cache, will always return the same results (unless the cache changes). But multiple calls to fetch can yield different results and imply a larger overhead.
 
Perhaps. Given this is not specific to this branch, I won't act on it here. Also, the above note about fixing this in kernel still applies (and would allow us to avoid inventing a weird-ish naming scheme by keeping our get() semantics everywhere).

> Compare that also to `nm_platform_link_infiniband_get_properties()`, `nm_platform_link_veth_get_properties()`, and `nm_platform_link_tun_get_properties()`. All these functions first try to look into the platform cache, and if kernel has no support, it uses alternative kernel API to fetch the properties (e.g. sysctl). For that reason,  `nm_platform_link_6lowpan_get_properties()` seems misnamed to me. Well, sure, it also has a fallback, but it still relies entirely on the platform cache to perform the lookup. Maybe just call it `nm_platform_link_6lowpan_get_parent()` and return the ifindex (not boolean)?
 
No. We do look into the platform cache first too. The whole routinge is basically the same as `nm_platform_link_veth_get_properties`.

src/devices/nm-device-6lowpan.c
lkundrak: Apart from failing on a bug, this would blow up when the device is removed at that very moment (which will be dealt with with subsequently). Logging a warning seems sensible to me.

src/platform/nm-linux-platform.c
lkundrak: Fixed.

NEWS
lkundrak: Fixed.

libnm-core/nm-setting-6lowpan.h
lkundrak: Fixed.

lkundrak:
-----

Pushed an updated branch

------

PR #121, opened 2018-05-25
policy: don't block connection if device is gone

aleksander0m:
If the active connection is deactivated because the device is gone,
don't block autoconnection. Otherwise, whenever the device comes
back (e.g. maybe it was reset in the middle of a connection attempt),
the autoconnection logic won't be triggered, as the settings are still
blocked.

I'm able to reproduce this by performing a WWAN modem reset in the
middle of a connection attempt.


thom311:
ok to test

bengal:
Looks good to me. CI failures are due to other issues now fixed on master.

thom311:
merged.

Thanks Aleksander!!

PR #122, opened 2018-05-25
th/dhcp-cleanup

thom311:
Some minor refactoring to DHCP.

- mostly: use cleanup-attribute at a few places
- use `NMIPConfig` type, which better unifies IPv4 vs IPv6 implementations.


bengal:
Ack.

thom311:
thanks. Merged.

PR #123, opened 2018-05-25
cli: fix string matching

bengal:


thom311:
the "expected" files must also be dist-ed in Makefile.am. That is rather inconvenient indeed... maybe that should be changed, that the tests only work from git...

bengal:
> the "expected" files must also be dist-ed in Makefile.am. That is rather inconvenient indeed... maybe that should be changed, that the tests only work from

Fixed, thanks.

thom311:
lgtm

lkundrak:
looks good (though a rebase & conflict resolution is needed before merge)

bengal:
Merged.

PR #125, opened 2018-05-29
th/cli-connection-handling-2

thom311:
This is a continuation of PR #100.

It reworks how nmcli handles connections. Long term goal is that eventually `nmc_print()` should be used everywhere. `nmc_print()` separates the printing/layout of the text, from accessing and getting properties. Once that's happening, `nmc_print()` itself should be simplified.

Also, a major part of this branch, is to improve nmcli handling profiles that are active more than once. This is related to wildcard profiles (rh#1555012).


bengal:
Looks good to me.

thom311:
merged to master

PR #126, opened 2018-05-30
th/build-cleanup-defines

thom311:
Cleanup defines.

- avoid using different names for things that are similar.
- consistently, only define the source-root and build-root tree. Each test should itself define locally the relative paths it is interested in.


thom311:
merged

PR #128, opened 2018-06-01
lr/cli-wifi-list-rescan: ensure the AP list is fresh on "nmcli d wifi list"

lkundrak:
Following up from https://bugzilla.gnome.org/show_bug.cgi?id=793845

@thom311 wrote:

> ```
> +        if (--nmc->should_wait == 0) {
> +         if (nmc->return_value == NMC_RESULT_ERROR_NOT_FOUND) {
> +              g_string_printf (nmc->return_text, _("Error: Access point with bssid '%s' not found."),
> +                               data->bssid_user);
> +         }
> +                g_main_loop_quit (loop);
> +    }
> ```
>
> intention

Fixed

> ```
> +wifi_list_rescan_cb (GObject *source_object, GAsyncResult *res, gpointer user_data)
> +{
> +    NMDeviceWifi *wifi = NM_DEVICE_WIFI (source_object);
> +    WifiListData *data = user_data;
> +    gs_free_error GError *error = NULL;
> +
> +    if (!nm_device_wifi_request_scan_finish (wifi, res, &error)) {
> +         if (   g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED)
> +             || g_error_matches (error, NM_DEVICE_ERROR, NM_DEVICE_ERROR_NOT_ALLOWED))
> +              return;
> ```
>
> why does it not do anything in case NM_DEVICE_ERROR_NOT_ALLOWED?

Good point. Added a comment.

> Also, in the cancelled case, will wifi_list_finish() ever be called to cleanup the data?

Yes. There's also a timeout set up to safeguard the case the scan couldn't finish in reasonable period that cleans this up.

In fact, the timeout firing is the only way the scan could be cancelled.

> ```
> time_t rescan_cutoff
> ```
>
> The choice of the D-Bus type for LastScan and nm_device_wifi_get_last_scan() already makes it clear that the program is not gonna work with anything outside the range of int32. Let's not compare or do arithmetic with intergers of different (even unknown) types like time_t. Just convert time_t early on to int/gint32 and use a consistent integer type.

Done

> > cli/devices: parse "nmcli d wifi list" arguments with next_arg()
>
> ```
> -         if (strcmp (*argv, "ifname") == 0) {
> +         case 1: /* ifname */
> ```
>
> is this really better then before? Instead of explicitly having the option name, it's now a number with a comment.
>
> Ok, whatever. I don't really mind. But there is also matches_arg() which has an odd overlap with argument parsing in next_arg(). Especially, matches_arg() can support forms like --rescan=auto.
> Also, next_arg() explicitly completes
>    nmc_complete_strings (**argv, "--ask", "--show-secrets", NULL);
> which is wrong for most subcommands
>
>   $ nmcli --complete-args networking connectivity -
>   -help
>   --help
>   --ask
>   --show-secrets
>
> TODO.

Done nothing about this, seems fine enough to me. A matter of opinion.

> regardless of how fres the access point list is.</para>
>                   ^^^^

Fixed.

> ```
> -    if (ifname) {
> +    if (rescan == NULL || strcmp (rescan, "auto") == 0) {
> +         struct timespec tp;
>
> +         if (clock_gettime (CLOCK_BOOTTIME, &tp) == -1) {
> +              g_printerr (_("Failed to get CLOCK_BOOTTIME: %s\n"), strerror (errno));
> +              rescan_cutoff = 0;
> +         } else {
> +              rescan_cutoff = MAX (tp.tv_sec - 30, 0);
> +         }
> +    } else if (strcmp (rescan, "no") == 0) {
> +         rescan_cutoff = 0;
> +    } else if (strcmp (rescan, "yes") == 0) {
> +         rescan_cutoff = -1;
> +    } else {
> +         g_string_printf (nmc->return_text, _("Error: invalid rescan argument: '%s' not among [auto, no, yes]"), rescan);
> +         return NMC_RESULT_ERROR_USER_INPUT;
> ```
>
> optimally, the options would be validated left to right. While certain validation is not possible right away, just validating for auto|no|yes is possible right away. That should be done inside the loop above.

I don't understand clearly what are you talking about here. If you're suggesting that I add a redundant check for yes/no/auto somewhere around the `rescan = *argv;` line then that sounds completely superfluous to me.

I think that this is already structured rather reasonably -- first the arguments are picked off argv[] and assigned into variables by their meaning and the if more work is required (be it finding the device instance for ifname, or figuring out the rescan timeout byt the rescan argument value). If an error is encountered along the way then it's always handled wherever it arises -- there doesn't seem to be a need for any advance validation.

Is there a case where you suppose this wouldn't work, or does the code just seem difficult to follow?

> Maybe expose the timestamp with milliseconds precision? (and type gint64).
>
> I know, AccessPoint's LastSeen is in the same scale as you added LastScan, which is good for consistency. On the other hand, CheckPoint's "Created" timestamp is also in milliseconds.
>
> Milliseconds precision is plenty for LastScan, whole seconds barely.

Well, the reason was indeed consistency with `LastSeen`. I think it's good and would need to be convinced otherwise. The seconds are all that matters here, the user is very unlikely to be in a situation where they would need to discover an access point that appeared within a fraction of a second.

(Pushed an updated branch)


clients/cli/devices.c
```diff
@@ -2640,24 +2815,20 @@ do_device_wifi_list (NmCli *nmc, int argc, char **argv)
...
 	NMDevice *device = NULL;
-	NMAccessPoint *ap = NULL;
 	const char *ifname = NULL;
 	const char *bssid_user = NULL;
+	const char *rescan = NULL;
 	gs_free NMDevice **devices = NULL;
-	const GPtrArray *aps;
-	APInfo *info;
-	int i, j;
+	int i; //, j;
```
thom311: ^^^

thom311: and lets not use signed integers for loop variables that shall not be negative?

clients/cli/devices.c
```diff
@@ -2709,55 +2893,33 @@ do_device_wifi_list (NmCli *nmc, int argc, char **argv)
 	if (argc)
 		g_printerr (_("Unknown parameter: %s\n"), *argv);
 
-	if (ifname) {
+	if (rescan == NULL || strcmp (rescan, "auto") == 0) {
+		struct timespec tp;
+
+		if (clock_gettime (CLOCK_BOOTTIME, &tp) == -1) {
```
thom311: NetworkManager falls back to CLOCK_MONOTONIC, if CLOCK_BOOTTIME is not supported. Client-side we should behave identically. Possibly there should be a public API function in libnm which returns a suitable "now" timestamp.

src/devices/wifi/nm-device-wifi.c
```diff
@@ -3300,6 +3308,11 @@ nm_device_wifi_class_init (NMDeviceWifiClass *klass)
 	                          G_PARAM_READABLE |
 	                          G_PARAM_STATIC_STRINGS);
 
+	obj_properties[PROP_LAST_SCAN] =
```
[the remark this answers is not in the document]
lkundrak: Ok.

clients/cli/devices.c
lkundrak: Ok.

clients/cli/devices.c
lkundrak: CLOCK_BOOTTIME is there since 2.6.39, we can safely assume it's always there.
The CLOCK_MONOTONIC is there just so that unit tests succeed on RHEL-6 kernel.

thom311:
Commit "wifi: expose the LastScan property" already contains changes to the clients tests, but it's too early and it breaks the test on that commit. Could you rebase the commit, so that the tests pass, and that it's clear which commit fixes nmcli (and requires adjustment of the tests)?

---

```
              self.call_nmcli_l(mode + ['-f', 'ALL', 'device', 'wifi', 'list', 'bssid', 'C0:E2:BE:E8:EF:B6'],
-                               replace_stdout = replace_stdout, expected_stderr = _UNSTABLE_OUTPUT)
+                               replace_stdout = replace_stdout)
```
now, that you fix the assertion failure, the stderr output is no longer unstable. Let's remove `_UNSTABLE_OUTPUT`?

lkundrak:
Done. Pushed an updated branch.

thom311:
lgtm

PR #130, opened 2018-06-05
th/clients-tests

thom311:
Extend unit tests to verify connection using libnm.

This war rather cumbersome to do, because our `test-networkmanager-service.py` uses python's D-Bus library, while libnm uses GDBus. Hence, I first had to convert the dbus structure to a GVariant, which is surprisingly complicated.


bengal:
LGTM

thom311:
Merged to master

PR #132, opened 2018-06-11
lr/wwan-ppp-route-rh1585611:  wwan: set the route parameters at the beginning of ip4 config 

lkundrak:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1585611

### DIFF
--- a/src/devices/bluetooth/nm-device-bt.c
+++ b/src/devices/bluetooth/nm-device-bt.c
@@ -903,11 +903,8 @@ act_stage3_ip6_config_start (NMDevice *device,
 {
 	NMDeviceBtPrivate *priv = NM_DEVICE_BT_GET_PRIVATE ((NMDeviceBt *) device);
 
-	if (priv->bt_type == NM_BT_CAPABILITY_DUN) {
-		return nm_modem_stage3_ip6_config_start (priv->modem,
-		                                         nm_device_get_act_request (device),
-		                                         out_failure_reason);
-	}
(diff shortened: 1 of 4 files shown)


lkundrak:
Actually, not. Let's do that in stage2, so that it works for both IPv4 and IPv6.

thom311:
nm_modem_set_route_parameters_from_device() takes the device's route-metric and route-table.

Note, that reapply does not permit changes to route-table setting ( https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/devices/nm-device.c?id=c2932dd7dbb8997b7c46b49506efbd16cd3f56ca#n10399 ), however, it allows to reapply changes to the route-metric.

By calling it only in state2, I think reapplying the route-metric won't work with routes obtained from NMModem, will it? Usually, during reapply we restart DHCP/autoconf6, so we use the new route-metric.

bengal:
LGTM

bengal:
> Usually, during reapply we restart DHCP/autoconf6, so we use the new route-metric.

Do we? I think not. Currently, route-metrics don't seem to be reapplied for DHCP. That is a missing feature.

thom311:
ah, it doesn't work for the default-route, but works for DHCP.

Missing feature is ok, but then https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/devices/nm-device.c?id=c2932dd7dbb8997b7c46b49506efbd16cd3f56ca#n10399 should reject reapplying the route-metric.

PR #133, opened 2018-06-11
th/dhcp-client

thom311:
Some fixes for ipv4.dhcp-client-id and ipv6.dhcp-duid

### DIFF
--- a/libnm-core/nm-utils.c
+++ b/libnm-core/nm-utils.c
@@ -4448,11 +4448,10 @@ _nm_utils_inet6_is_token (const struct in6_addr *in6addr)
 gboolean
 _nm_utils_dhcp_duid_valid (const char *duid, GBytes **out_duid_bin)
 {
+	guint8 duid_arr[128 + 2];
 	gsize duid_len;
-	gs_unref_bytes GBytes *duid_bin = NULL;
 
-	if (out_duid_bin)
-		*out_duid_bin = NULL;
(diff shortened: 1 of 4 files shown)


libnm-core/nm-utils.c
```diff
@@ -4448,11 +4448,10 @@ _nm_utils_inet6_is_token (const struct in6_addr *in6addr)
 gboolean
```
fgiudici: Typos in the commit msg:
"Also, [is/it] is called from [...]"
[...]
"hug[e] settings, which could trick NetworkManage[r] to allocate [...]"

src/devices/nm-device.c
```diff
@@ -7881,51 +7900,42 @@ dhcp6_get_duid (NMDevice *self, NMConnection *connection, GBytes *hwaddr, NMDhcp
...
 
-	} else if (nm_streq0 (duid, "stable-uuid")) {
-		duid_out = generate_duid_uuid (sha256_digest, len);
+		goto out_good;
 	}
 
-	duid_error = "generation failed";
-end:
-	if (!duid_out) {
+	g_return_val_if_reached (NULL);
```
fgiudici: We want to log the error but provide the backup random DUID here: wouldn't be better have just an nm_assert here?

thom311:
All comments by Francesco addresses (via IRC)

bengal:
LGTM

PR #134, opened 2018-06-12
cli: Reuse connections in nmcli dev wifi con

balrog-kun:


vhumpa:

thom311:
Hi,

this obviously changes behaviour. Could you elaborate, why you think this behaviour is better?

The commit message says, that it allows to connect to WPA-Enterprise connections. But that only works if the user already has a profile. In that case, why should the user not just call `nmcli con up`?

thom311:
The first commit is obviously right. Merged to master as e8a197506cf61161397d8f67304d89187ebbd26d. Thanks!

balrog-kun:
Thanks.  WRT the second commit it seems like creating a new connection when one already exists is an obvious shortcoming so I'm trying to fix that.  You could perform the same check for whether a connection already exists in the calling script, or somewhere else, and then decide the command sequence to use but that's more work.  So if you can use a single mechanism like you'd do in a GUI by clicking on the connection name, that should be an improvement.

thom311:
maybe there could be a command line option to select whether it should try activating an existing? (leaving the default behavior as it currently is).

balrog-kun:
If the current behavior is there on purpose then I don't mind dropping this patch.  I couldn't see why you'd want the new connections created.  I guess if you had a connection for the given AP that was misconfigured then you might want a fresh one with default settings although you could just fix the saved connection too.

thom311:
I do think the patch makes a lot of sense.

Note, how also `nmcli device connect "$IFNAME"` tries to re-use an existing connection.


I think, the ability to let NetworkManager create a new profile is valueble. Both
```
nmcli device connect $IFNAME`
```
and
```
nmcli device wifi connect $SSID
```
achieve that.

The problem is, the latter always creates a now profile, and the former only creates a new profile if none exists yet.


For consistency reasons, I think the patch makes sense in general -- even if it is a change in behaviour. I would think, people would not commonly encounter this, or not use this command from scripts.


Regardless, it would be interesting to have an `--always-create-new` option for both cases. Independent of this patch.


@balrog-kun , care to re-open the pull request? :) Thanks.

balrog-kun:
Ok, reopened as PR #137.  Add --always-create-new for `device wifi connect` is easy enough, for `device connect $IFNAME` it might be more difficult unless you specifically mean creating a copy of an existing connection, specially for wifi devices.
`--new-connection` may be a better name because `--always-create-new` can't be abbreviate to `-a` as that means `--ask`.

PR #135, opened 2018-06-13
cli: Actually add passwords to the AddAndActivateConnection arguments

balrog-kun:
Check the wpa_flags and rsn_flags values to see if the network needs the
password added to the 802-11-wireless-security settings.  The current
ap_flags check alone would only trigger the password to be sent for WEP
networks.  Also remove unneeded initialization of the three variables.


vhumpa:

lkundrak:
+1.

Applied upstream, thanks.

PR #136, opened 2018-06-13
fg/auto neg specmodes rh1487477

fgiudici:
Allow to force  speed / duplex mode also when auto-negotiation is enabled (for 802.3 BASE-T specifications only). This will end up configuring the device to advertise only the specified mode for the link negotiation.
The change is required as the 802.3 standard mandates that 1000BASE-T and 10000BASE-T specifications have the link auto-negotiation feature enabled (required as clock synchronization is required).

https://bugzilla.redhat.com/show_bug.cgi?id=1487477


src/platform/nm-platform-utils.c
```diff
@@ -404,7 +455,18 @@ nmp_utils_ethtool_set_link_settings (int ifindex,
...
+			edata.advertising = edata.supported;
+		else {
+			unsigned long int mode;
+
+			mode = get_baset_mode (speed, duplex);
+
+			if (   !mode
+			    || !(edata.supported & mode))
+				return FALSE;
+			edata.advertising = edata.supported & ~(BASET_ALL_MODES & ~mode);
```
thom311: edata.advertising is uint32 type. There is no need to pretend, mode could be an "unsigned long int". `get_baset_mode()` should return guint32. If you want, you can add a `G_STATIC_ASSERT()` that all `ADVERTISED_` flags fit into guint32, though not necessary. I think when doing bitwise operations, all operands should have the same integer type.

src/platform/nm-platform-utils.c
```diff
@@ -384,6 +384,57 @@ nmp_utils_ethtool_get_link_settings (int ifindex,
...
+		case 1000:
+			speed_index = 2;
+			break;
+		case 10000:
+			speed_index = 3;
+			break;
+		default:
+			return ADVERTISED_INVALID;
+			break;
+	}
```
thom311: wouldn't it be simpler to just return the selected value right away?
```
if (duplex == NM_PLATFORM_LINK_DUPLEX_HALF) {
    switch (speed) {
        case 10: return ADVERTISED_10baseT_Half;
        ...
        default: return ADVERTISED_INVALID;
    }
} else {
    ...
```

A lookup array makes sense, if you can avoid the if-else search for the index. But you don't avoid that. It only adds more code.

src/platform/nm-platform-utils.c
thom311:
```
edata.supported & ~(BASET_ALL_MODES & ~mode)
```
is the same as
```
(edata.supported & ~BASET_ALL_MODES) | (edata.supported & mode)
```
but the former seems hard to understand (to me).

src/platform/nm-platform-utils.c
```diff
@@ -404,7 +455,18 @@ nmp_utils_ethtool_set_link_settings (int ifindex,
...
+		if (!speed && !duplex)
+			edata.advertising = edata.supported;
+		else {
+			unsigned long int mode;
+
+			mode = get_baset_mode (speed, duplex);
+
+			if (   !mode
+			    || !(edata.supported & mode))
+				return FALSE;
```
thom311: This returns FALSE silently, without possibility to find out why it failed in the logfile. ethool-utils already log with level trace when something important happens. It should be done in this case as well.

src/platform/nm-platform-utils.c
fgiudici: From ethtool.h:
"[...] Please do NOT define any SUPPORTED_* or ADVERTISED_ macro for bits > 31. [...]"
So, looks more than safe to use guint32 for the ADVERTISED_ flags.

As we check just above that (edata.supported & mode) is > 0, and mode is by construction just "one" mode (1 flipped bit), what about:
edata.advertising = (edata.supported & ~BASET_ALL_MODES) | mode  ?

src/platform/nm-platform-utils.c
fgiudici: Didn't liked the double switch construct... but well, I agree the alternative didn't looked really better. Maybe a little worse as less straightforward.
So, switched to plain construct as you proposed.

src/platform/nm-platform-utils.c
fgiudici: Right. Added.

src/platform/nm-platform-utils.c
thom311: ` (edata.supported & ~BASET_ALL_MODES) | mode`, yes even better. It's less about `mode` being just one bit (which is not immediately obvious), but more about mode being a subset of BASE_ALL_MODES (which seems obvious -- from the naming).

src/platform/nm-platform-utils.c
```diff
@@ -395,6 +447,7 @@ nmp_utils_ethtool_set_link_settings (int ifindex,
 	};
 
 	g_return_val_if_fail (ifindex > 0, FALSE);
+	g_return_val_if_fail ((!speed && duplex) || (speed && !duplex), FALSE);
```
bengal: Why this check? And perhaps it should compare duplex with an enum value, not assume that zero == UNKNOWN.

src/devices/nm-device-ethernet.c
```diff
@@ -805,16 +803,17 @@ link_negotiation_set (NMDevice *device)
 
 	/* If link negotiation setting are already in place do nothing and return with success */
 	if (   (!!autoneg == !!link_autoneg)
-	    && (!speed || (speed == link_speed))
-	    && (!duplex || (duplex == link_duplex))) {
+	    && (speed == link_speed)
+	    && (duplex == link_duplex)) {
```
bengal: Internal parens aren't needed anymore.

libnm-core/nm-setting-wired.c
```diff
@@ -1057,8 +1049,16 @@ nm_setting_wired_class_init (NMSettingWiredClass *setting_wired_class)
 	/**
 	 * NMSettingWired:duplex:
 	 *
-	 * Can be specified only when "auto-negotiate" is "off". In that case, statically
-	 * configures the device to use that specified duplex mode, either "half" or "full".
+	 * When a value is set, either "half" or "full", configures the device
+	 * to use the specified duplex mode. If "auto-negotiate" is "yes" the
+	 * specified duplex mode will be the only one advertized during link
```
bengal: advertized -> advertised

libnm-core/nm-setting-wired.c
```diff
@@ -1057,8 +1049,16 @@ nm_setting_wired_class_init (NMSettingWiredClass *setting_wired_class)
...
+	 * negotiation: this works only for BASE-T 802.3 specifications and is
+	 * useful for enforcing gigabits modes, as in these cases link negotiation
+	 * is mandatory.
+	 * If the value is unset (the default), the link configuration will be
+	 * either skipped ((if "auto-negotiate" is "no", the default) or will
```
bengal: Double parenthesis.

src/platform/nm-platform-utils.c
fgiudici: If a mode is specified, both speed and duplex should have a value (otherwise the mode is unspecified). While this could be anyway fine when autoneg is off, because we will keep the missing duplex or speed from the current status this does not work when autoneg is on  and we need to specify a particular mode.
We assume in the logging that the duplex is specified when the speed is.
 So, let's enforce it at the beginning of the function and check to have either both specified or both unset.
Ok, let's check duplex against NM_PLATFORM_LINK_DUPLEX_UNKNOWN

src/devices/nm-device-ethernet.c
fgiudici: ok, removed

libnm-core/nm-setting-wired.c
fgiudici: ops, fixed

libnm-core/nm-setting-wired.c
fgiudici: fixed

fgiudici:
Addressed all the reviews and added one more commit to allow read/writes of the new values in the ifcfg plugin

src/platform/nm-platform-utils.c
```diff
@@ -404,7 +440,31 @@ nmp_utils_ethtool_set_link_settings (int ifindex,
...
+				              speed,
+				              (duplex == NM_PLATFORM_LINK_DUPLEX_HALF) ? "Half" : "Full");
+				return FALSE;
+			}
+			if (!(edata.supported & mode)) {
+				nm_log_trace (LOGD_PLATFORM,
+				              "ethtool[%d]: device does not support %uBASE-T %s duplex mode",
+				              ifindex,
+				              speed,
+			                      (duplex == NM_PLATFORM_LINK_DUPLEX_HALF) ? "Half" : "Full");
```
thom311: indentation. Also, `link_duplex_to_string()` returns lower-case words (not "Half" and "Full"). Can you move `link_duplex_to_string()` to nm-platform.c, and re-use it?

src/platform/nm-platform-utils.c
```diff
@@ -395,6 +429,8 @@ nmp_utils_ethtool_set_link_settings (int ifindex,
 	};
 
 	g_return_val_if_fail (ifindex > 0, FALSE);
+	g_return_val_if_fail (   (!speed && duplex != NM_PLATFORM_LINK_DUPLEX_UNKNOWN)
+	                      || (speed && duplex == NM_PLATFORM_LINK_DUPLEX_UNKNOWN), FALSE);
```
bengal: The condition still seems inverted to me: it should allow only both values set or both unset.

src/platform/nm-platform-utils.c
fgiudici: Done

src/platform/nm-platform-utils.c
fgiudici: Right, fixed thanks

PR #137, opened 2018-06-13
cli: reuse connections in nmcli dev wifi con

balrog-kun:
Try to locate an existing connection before creating a new one when
handling "nmcli device wifi connect".  This allows WPA-Enterprise
networks to be activated this way, consistent with the comment that this
command is equivalent to clicking on an SSID in a GUI client.

### DIFF
--- a/clients/cli/devices.c
+++ b/clients/cli/devices.c
@@ -1849,7 +1849,6 @@ add_and_activate_cb (GObject *client,
 	AddAndActivateInfo *info = (AddAndActivateInfo *) user_data;
 	NmCli *nmc = info->nmc;
 	NMDevice *device = info->device;
-	NMActiveConnectionState state;
 	NMActiveConnection *active;
 	GError *error = NULL;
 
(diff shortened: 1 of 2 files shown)


vhumpa:

thom311:
this is a follow-up to #134

thom311:
I think the change in behavior is ok, because:
1) it makes it consistent with what `nmcli device connect` does
2) it is unlikely, that a user wants to use this command to repeatedly create new profiles
3) it is unlikely, that a user currently relies on the current behavior.
4) this command generally makes more sense for interactive use (and if it was called from a script/programatically, 3) applies). For interactive use, change in behaviour is less of a problem.

I think, an option `--new-connection` would be nice. But since we are going to rectify the default-behavior, an additional option can be added any time later.

clients/cli/devices.c
```diff
@@ -3203,45 +3219,111 @@ do_device_wifi_connect_network (NmCli *nmc, int argc, char **argv)
...
+		 * fill them into proper settings. */
+		if (con_name || private || bssid2_arr || password || hidden)
+			connection = nm_simple_connection_new ();
 
-		/* 'hidden' parameter is used to indicate that SSID is not broadcasted */
-		if (hidden) {
-			GBytes *ssid = g_bytes_new (param_user, strlen (param_user));
+		if (!con_name || private) {
+			s_con =  (NMSettingConnection *) nm_setting_connection_new ();
+			nm_connection_add_setting (connection, NM_SETTING (s_con));
```
bengal: It seem to me that 'connection' can be uninitialized here.

balrog-kun: Good catch, the `!con_name` is actually a typo.  Not sure how this happened as this part only has more indent added.

thom311:
Thanks, I appreciate the work for the `--new-connection` option.

however, I have to complain about it :)

nmcli (for better or worse), has two types of --options. Those that apply to all subcommands and must come first (e.g. `--terse`), and those that are for the subcommands. I find this distinction in general problematic, and maybe on day, we should rework this, so that options can be specified in any order.

Anyway, for now, `--new-connection` is really an option that only makes sense for this particular command. Hence, it should be more like `nmcli device wifi list --rescan` -- an option for the subcommand.

balrog-kun:
Makes sense, I'll change this.  I chose this option because we also talked about using it for `device connection up` but I don't see it as very practical.

lkundrak:
I like the patch as it is. This some thing, that nmcli always creates a new connection, bothered me for quite  some time too. Your implementation looks sane to me.

I'd prefer if we didn't add the `--new-connection` switch though. Whoever need to always create a new connection has `nmcli c add` at their disposal. And we already have quite too many switched that don't contribute to usability too much.

I've tested this too; works well for me. Rebased on top of master, dropped the last patch and added a minor comsetic fix:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/tree/lr/balrog-kun/cli-d-wifi-connect

@thom311 -- how strongly you feel the `--new-connection` option is necessary? Would you mind if I merged this now without it? (And perhaps add it, if someone actually has an use for it?)

thom311:
Thanks Lubomir, for picking this up...


> I'd prefer if we didn't add the --new-connection switch though. Whoever need to always create a new 
> connection has nmcli c add at their disposal.

OK. But note that `nmcli con add` is not a replacement for `nmcli device [wifi] connect`. The latter lets NetworkManager daemon create a profile based on how the server sees fit (e.g. with information from the AP). With `nmcli con add` the user has to set the Wi-Fi parameters manually.

On the other hand, if the user just wants to activate *any* existing profile on the device, then there is already a fully suitable replacement:

    nmcli connection up ifname $DEVICE

I do agree, we should add command line options carefully. Also, `nmcli device connect` and `nmcli device wifi connect` should behave consistently. So, I don't really oppose this change. nmcli only provides a subset of possibly useful options, so it's not clear to me that this change shifts the subset into the more favourable direction.


---
```
          const char *candidate_bssid = nm_access_point_get_bssid (candidate_ap);
 
+         if (bssid && candidate_bssid) {
```

isn't this wrong? If the user specified bssid, we shouldn't be satisfied wit an AP that has no BSSID (like we do with ssid).


```
+         if (complete)
+              continue;
```
this also seems not right. In complete mode, we still want to find and return the found ap.


branch lgtm.

balrog-kun:
> ```
>           const char *candidate_bssid = nm_access_point_get_bssid (candidate_ap);
>  
> +         if (bssid && candidate_bssid) {
> ```
> 
> isn't this wrong? If the user specified bssid, we shouldn't be satisfied wit an AP that has no BSSID (like we do with ssid).

You're right, this would be better if a candidate BSSID is required here.

> 
> ```
> +         if (complete)
> +              continue;
> ```
> 
> this also seems not right. In complete mode, we still want to find and return the found ap.

Oh ok... I assumed we wanted to print all matching APs SSIDs or BSSIDs in complete mode.

Best regards

thom311:
> >             if (complete)
> >                  continue;
> >
> >   this also seems not right. In complete mode, we still want to find and return the found ap.

>Oh ok... I assumed we wanted to print all matching APs SSIDs or BSSIDs in complete mode.

ah, you are right. I think the patch is correct.

balrog-kun:
Ok so I hopefully now made it require the BSSID on the AP when searching for a BSSID value.  I used the lr/balrog-kun/cli-d-wifi-connect branch and edited the first of the three commits and then pushed the result here.  I don't know if this is even helpful in github/gitlab workflow so feel free to ignore it.

lkundrak:
Thank you, merged.

PR #138, opened 2018-06-14
th/device-reapply-metric

thom311:
Follow-up to pr #132


bengal:
> shared: simplify nm_g_object_ref_set()

Can you also update the '@pdst' argument name in the documentation comment?

> device: don't allow reapplying a different route-metric

`+ /* changing the default route-metric is complicated. Don't supported at the moment. */`

"Don't supported" -> "Not supported".

Seems ok to me. Or maybe, since the problem is only with ipvX.method=auto, allow the route-metric reapply only for static IP method.

Otherwise, LGTM.

thom311:
hm, it seems too complicated.

I dropped the patches that prevent reapply of the route-metric. Let's just say, if the route-metric does not get properly reapplies, it's a bug that we should fix separately. Not, preventing it now. In some cases it does work.

Merged only the early bits.

PR #139, opened 2018-06-15
iwd secrets-related changes

balrog-kun:
I'm attempting to make 802.1x networks work with the IWD backend.  These patches enable IWD agent requests for TLS/TTLS/PEAP private key passwords to be sent to the NM clients' secret agents. To convey which secrets IWD wants specifically, I'm using the setting_name and hints parameters in the GetSecrets call although I'm not 100% clear this is the intended use but it seems to be at least one of the allowed uses. If this is Ok, I'll need to separately make sure this works with nm-applet and the Gnome Shell.
The NMSettingsConnection change (last commit) was prompted by the settings plugin I sent to the list but this change can be seen as independent from the plugin too.

### DIFF
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -1219,15 +1219,13 @@ act_stage1_prepare (NMDevice *device, NMDeviceStateReason *out_failure_reason)
 	ap = ap_path ? nm_wifi_ap_lookup_for_device (NM_DEVICE (self), ap_path) : NULL;
 	if (!ap) {
 		ap = nm_wifi_aps_find_first_compatible (&priv->aps_lst_head, connection);
-
-		/* TODO: assuming hidden networks aren't supported do we need
-		 * to consider the case of APs that are not in the scan list
-		 * yet, for which nm-device-wifi.c creates the temporary fake
-		 * AP object?
-		 */
+		if (!ap) {
+			NM_SET_OUT (out_failure_reason, NM_DEVICE_STATE_REASON_CONFIG_FAILED);
+			return NM_ACT_STAGE_RETURN_FAILURE;
+		}
 
 		nm_active_connection_set_specific_object (NM_ACTIVE_CONNECTION (req),
-	                                              nm_dbus_object_get_path (NM_DBUS_OBJECT (ap)));
+		                                          nm_dbus_object_get_path (NM_DBUS_OBJECT (ap)));
 	}
 
 	set_current_ap (self, ap, FALSE);
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -73,7 +73,6 @@ typedef struct {
 	GCancellable *  cancellable;
 	NMDeviceWifiCapabilities capabilities;
 	NMActRequestGetSecretsCallId *wifi_secrets_id;
-	GDBusMethodInvocation *secrets_request;
 	guint           periodic_scan_id;
 	bool            enabled:1;
 	bool            can_scan:1;
@@ -379,14 +378,6 @@ wifi_secrets_cancel (NMDeviceIwd *self)
 	if (priv->wifi_secrets_id)
 		nm_act_request_cancel_secrets (NULL, priv->wifi_secrets_id);
 	nm_assert (!priv->wifi_secrets_id);
-
-	if (priv->secrets_request) {
-		g_dbus_method_invocation_return_error_literal (priv->secrets_request, NM_DEVICE_ERROR,
-		                                               NM_DEVICE_ERROR_INVALID_CONNECTION,
-		                                               "NM secrets request cancelled");
-		priv->secrets_request = NULL;
-	}
-
 }
 
 static void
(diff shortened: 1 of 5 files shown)


vhumpa:

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -104,7 +104,10 @@ psk_agent_dbus_method_cb (GDBusConnection *connection,
 	if (!nm_streq0 (g_dbus_object_manager_client_get_name_owner (omc), sender))
 		goto return_error;
 
-	g_variant_get (parameters, "(&o)", &network_path);
+	if (!strcmp (method_name, "RequestUserPassword"))
+		g_variant_get (parameters, "(&os)", &network_path, &username);
```
thom311: use `&s` to not leak `username`. As you don't use the parameter, I think you could just pass `NULL` instead.

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -188,8 +203,50 @@ psk_agent_export (GDBusConnection *connection, gpointer user_data,
...
+	static const GDBusArgInfo *const request_user_password_in_args[] = {
+		&agent_request_arg_network,
+		&agent_request_arg_user,
+		NULL,
+	};
+	static const GDBusArgInfo *const request_user_password_out_args[] = {
+		&agent_request_arg_password,
+		NULL,
+	};
+	static const GDBusMethodInfo request_user_password_info = {
```
thom311: could you rework these to use our macros ([[1]](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-plugin.c?id=f4bef9a12eba69e2cd94b2428939d2a523f4fce0#n824), [[2]](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/nm-dbus-manager.c?id=f4bef9a12eba69e2cd94b2428939d2a523f4fce0#n1455))?

src/devices/wifi/nm-device-iwd.c
```diff
@@ -1746,38 +1872,34 @@ nm_device_iwd_set_dbus_object (NMDeviceIwd *self, GDBusObject *object)
...
-	                          NM_SETTING_WIRELESS_SECURITY_SETTING_NAME,
-	                          NM_SECRET_AGENT_GET_SECRETS_FLAG_ALLOW_INTERACTION
-	                            | NM_SECRET_AGENT_GET_SECRETS_FLAG_REQUEST_NEW);
+	wifi_secrets_get_one (self,
+	                      setting_name,
+	                      NM_SECRET_AGENT_GET_SECRETS_FLAG_ALLOW_INTERACTION
+	                        | NM_SECRET_AGENT_GET_SECRETS_FLAG_REQUEST_NEW,
+	                      setting_key);
 
 	priv->secrets_request = invocation;
```
thom311: it seems a bit error-prone to piggyback the invocation on `priv`. Indeed, you only set `priv->secrets_request` when also schedule a new secret request (that is: `priv->wifi_secrets_id` is initialized), and cancelling this request will (should) cleanup the invocation. It's not immediately obvious, that `wifi_secrets_get_one()` does nothing but setting `priv->wifi_secrets_id`, but it's correct. All of this is almost correct, but confusing.

A small leak is that `wifi_secrets_cancel()` will call `nm_act_request_cancel_secrets()`, which calls `wifi_secrets_cb()`, which clears `priv->secrets_request = NULL;`, thus the line in `wifi_secrets_cancel()` will never send `g_dbus_method_invocation_return_error_literal()`. The simple solution for that, is to move
```
»···if (priv->secrets_request) {
»···»···g_dbus_method_invocation_return_error_literal (priv->secrets_request, NM_DEVICE_ERROR,
»···»···                                               NM_DEVICE_ERROR_INVALID_CONNECTION,
»···»···                                               "NM secrets request cancelled");
»···»···priv->secrets_request = NULL;
»···}
```
into `wifi_secrets_cb()`.

But while at it, could you not piggyback priv->secret_requests? Instead, pass it on as async data. As you then have more async-data, you could use `nm_utils_user_data_pack()` / `nm_utils_user_data_unpack()` [[1]](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/src/nm-manager.c?id=f4bef9a12eba69e2cd94b2428939d2a523f4fce0#n4869). I prefer that, because then the invocation is always explicitly passed on and it's clearer to follow (instead of hooking it at `priv`).

src/devices/wifi/nm-iwd-manager.c
balrog-kun: Good point, fixed.

src/devices/wifi/nm-iwd-manager.c
balrog-kun: Done too, it looks better now.

src/devices/wifi/nm-device-iwd.c
balrog-kun: Oh yes, it was probably good before and I broke it with this path.  Now moved to use the suggested `nm_utils_user_data_pack()` / `nm_utils_user_data_unpack()` utilities in a separate preceding commit.
Also added one more unrelated fix as a new patch.

thom311:
Branch lgtm. I merged it to master (5e8773ee631541c4f4e39954a62e02bef6fb2154) and backported it to nm-1-12 (ae393ed9baf02c9b9baa5346944e1e90047b32df).

+ a some minor follow-up commits (if you disagree with the changes, we can still revert/adjust them).

Thank!

bengal:
commit "cli: use the hints from 802.1x secrets requests if given" broke CI tests where 802.1x secrets are asked by 'nmcli -a'. The commit looks wrong because when there is no hint, request->hints is an empty non-NULL array.

The check should probably be:

```
  -       if (request->hints) {
  +       if (request->hints && request->hints[0]) {
```

so that when NMDeviceIwd sets hints, only those are asked. What do you think?

bengal:
ok to test

vhumpa:

balrog-kun:
@bengal Oops, yes, `if (request->hints && request->hints[0])` is correct.  I copied the line from `add_vpn_secrets` assuming `request->hints` would be NULL but the context is different in that function.  I should have tested both wifi backends after this change...

thom311:
merged fix with Beniamino's suggestion (to master and nm-1-12)

PR #140, opened 2018-06-15
Use miliseconds precision for LastScan

lkundrak:
$SUBJ, and add an utility function to libnm to get the right timestamp.

The original motivation behind this is ability to hide the fallback to CLOCK_MONOTONIC, which is a hack to get our test suite run on terribly old kernels.


src/devices/wifi/nm-device-iwd.c
```diff
@@ -1535,10 +1535,10 @@ get_property (GObject *object, guint prop_id,
 		g_value_set_boolean (value, priv->scanning);
 		break;
 	case PROP_LAST_SCAN:
-		g_value_set_int (value,
-		                 priv->last_scan > 0
-		                     ? (gint) nm_utils_monotonic_timestamp_as_boottime (priv->last_scan, NM_UTILS_NS_PER_SECOND)
-		                     : -1);
+		g_value_set_int64 (value,
+		                   priv->last_scan > 0
+		                       ? (gint) nm_utils_monotonic_timestamp_as_boottime (priv->last_scan, NM_UTILS_NS_PER_MSEC)
```
thom311: (gint64)

src/devices/wifi/nm-device-wifi.c
```diff
@@ -3187,7 +3187,7 @@ get_property (GObject *object, guint prop_id,
 	case PROP_LAST_SCAN:
 		g_value_set_int (value,
 		                 priv->last_scan > 0
-		                     ? (gint) nm_utils_monotonic_timestamp_as_boottime (priv->last_scan, NM_UTILS_NS_PER_SECOND)
+		                     ? (gint) nm_utils_monotonic_timestamp_as_boottime (priv->last_scan, NM_UTILS_NS_PER_MSEC)
```
thom311: (gint64)

libnm-core/nm-utils.c
```diff
@@ -6223,6 +6223,33 @@ nm_utils_format_variant_attributes (GHashTable *attributes,
 
 /*****************************************************************************/
 
+/*
+ * nm_utils_get_timestamp_msec():
+ *
+ * Gets current time in milliseconds of CLOCK_BOOTTIME.
```
thom311: That this is CLOCK_BOOTTIME is an implementation detail. It's not even necessary to document it (but fine, if you want).

More important however is, that the returned time is the same source and scale, as various timestamps that are exposed by NetworkManager on D-Bus.

Also say, that the value only has meaning on the same host, and that it is increasing monotonically.

libnm-core/nm-utils.c
```diff
@@ -6223,6 +6223,33 @@ nm_utils_format_variant_attributes (GHashTable *attributes,
...
+nm_utils_get_timestamp_msec (void)
+{
+	struct timespec ts;
+
+	if (clock_gettime (CLOCK_BOOTTIME, &ts) == -1) {
+		/* The fallback to CLOCK_MONOTONIC is taken only if we're running on a
+		 * criminally old kernel, prior to 2.6.39 (released on 18 May, 2011).
+		 * That happens during buildcheck on old builders, we don't expect to
+		 * be actually runs on kernels that old. */
+		return g_get_monotonic_time () / 1000;
```
thom311: should you not check `errno == EINVAL` and only perform the fallback then?
And on all other errors, `g_return_val_if_reached (-1);`
^^ though, of course, kernel really never should fail in such a manner (hence the assertion).

Also, is it nice to use g_get_monotonic_time()? GLib leaves it undefined that this is actually CLOCK_MONOTONIC (for good reasons), but libnm -- as an implementation -- should explicitly fallback to CLOCK_MONOTONIC -- despite itself leaving the actually used clock undefined (from a user's point of view)

lkundrak: Okay. I'm not too comfortable about returning -1 -- the function should not be expected to ever fail, since even GLib itself can't work without CLOCK_MONOTONIC. But I don't really mind -- it's never going to be reached anyway.

thom311: You are right. glib even crashes with `g_error()`, if it cannot get `CLOCK_MONOTONIC`. It's good to have functions that cannot fail -- especially, if it's not clear how to handle the error.

thom311:
All patches merged as 9c3e52b5321cdec266fe9c36e35196c45bd0b936 (with minor changes).
Except `examples/python: drop nmex.py`, which was excluded from merging.

I think, nmex.py should not be dropped, because:

- nmex.py is intended to be a utility-toolbox for the python examples. With the purpose to write more complex examples, while avoiding copy & paste between them. Even if the timestamp functions are no longer necessary, nmex.py should stay (so it is there for use in the future). Yes, there is a conflict: examples should be as simple as possible without extra dependencies. So, nmex.py should be used with care, in particular it should be simple and self-contained itself. The alternative is copy&paste, and having lots of examples that are unmaintained and no longer preferred style.
- nmex.py would also be useful for the python examples that are not based on libnm/gi. Adding a new utility function to libnm does not help the examples under `python/dbus` directory. Arguably, the code is not used there at the moment.

lkundrak:
That doesn't sound reasonable to me at all. The examples should be as self-contained as possible, turning them into modules just impairs their methodical value.

Apart from that, there's not a single bit of useful code in nmex.py. Why not add it in an unlikely event it's actually needed?

PR #142, opened 2018-06-18
dhcp: drop NMDhcpDuidEnforce type

fgiudici:
A gboolean is enough: make code easier.


src/devices/nm-device.c
```diff
@@ -7972,7 +7972,7 @@ dhcp6_get_duid (NMDevice *self, NMConnection *connection, GBytes *hwaddr, NMDhcp
 	       "ipv6.dhcp-duid: generate %s DUID '%s' (%s)",
 	       duid,
 	       (logstr1 = nm_dhcp_utils_duid_to_string (duid_out)),
-	       (duid_enforce == NM_DHCP_DUID_ENFORCE_ALWAYS) ? "enforcing" : "fallback");
+	       duid_enforce ? "enforcing" : "fallback");
```
thom311: let's change "fallback" to "prefer lease".

src/dhcp/nm-dhcp-client.c
```diff
@@ -541,10 +541,7 @@ nm_dhcp_client_start_ip6 (NMDhcpClient *self,
 	nm_assert (!priv->duid);
 	nm_assert (client_id);
 
-	if (enforce_duid == NM_DHCP_DUID_ENFORCE_NEVER)
-		priv->duid = NM_DHCP_CLIENT_GET_CLASS (self)->get_duid (self, TRUE);
-	else if (enforce_duid == NM_DHCP_DUID_ENFORCE_LEASE_FALLBACK)
-		priv->duid = NM_DHCP_CLIENT_GET_CLASS (self)->get_duid (self, FALSE);
+	priv->duid = NM_DHCP_CLIENT_GET_CLASS (self)->get_duid (self, !enforce_duid);
```
thom311: should be `if (!enforce_duid) ... get_duid (self, TRUE);`

fgiudici: oh, right... fixed thanks.
I realized also that we don't use nor need anymore the gboolean option in get_duid: reverting the commit that brought that in.

src/devices/nm-device.c
fgiudici: Done

thom311:
lgtm

PR #144, opened 2018-06-20
th/device-dad-handling-rh1593210

thom311:
There is a race, where NetworkManager would touch externally configured devices. See CI failure at https://bugzilla.redhat.com/show_bug.cgi?id=1593210 .

Please review


bengal:
>     But it's worse. If the device is in state disconnected, and the user
>     externally adds IPv6 addresses, we will end up in queued_ip_config_change().
>     It is easily possible that need_ipv6ll evaluates to TRUE, which results
>     in a call to check_and_add_ipv6ll_addr() and possibly
>     
>       ip_config_merge_and_apply (self, AF_INET6, TRUE);
>     
>     This in turn will modify the IP configuration on the device, although
>     the device may be externally managed and NM shouldn't touch it.

ip_config_merge_and_apply() has:

```
	if (nm_device_sys_iface_state_is_external (self))
		commit = FALSE;
```
so it doesn't seem possible that we end up changing an externally managed device in this way. But perhaps what happens is that we call merge_and_apply() when the device is still disconnected (sys-iface-state:managed), before the in-memory connection is generated.

bengal:
The branch LGTM.

thom311:
Merged to master (8076a6f00355710f3d0706a981aefdbb25b33e3b) and backported to nm-1-12 (f3d1cf8100dbe75f04cac3787b2c1a593190cdcb), so that it will be part of 1.12.0 relese.

Note that it fixes test `ipv6_keep_external_routes` ([rh#1593210](https://bugzilla.redhat.com/show_bug.cgi?id=1593210#c4)), but breaks test `ipv6_add_static_address_manually_not_active` ([rh#1585078](https://bugzilla.redhat.com/show_bug.cgi?id=1585078#c8)).

PR #147, opened 2018-06-25
th/libnm-register-settings

thom311:
Drop some `__attribute__((constructor))` uses in libnm. I think it should be avoided to run code before `main()`.


thom311:
merged to master

PR #152, opened 2018-06-29
nm-online: print "[started|start-pending|failure]" result for --wait-for-startup

thom311:
`nm-online --wait-for-startup` isn't really intended to be called directly.
It is mainly for implementing NetworkManager-wait-online. service.

Anyway, at the end, the result does not indicate the connectivty status
of the host. Hence, printing

  Connecting...............   30s [online]

is misleading. It merely means, that startup is complete. Likewise,
printing "[offline]" would not mean that there is no connectivity.
Instead, it means that startup is still in progress.

In this form the distinction between whether to print "start-pending"
and "failure" is not very clear. Not that it matters much though. At
least is corresponds to the exit status of the program. If we sometimes
confuse EXIT_FAILURE_OFFLINE with other failure reasons, the exit could
should be adjusted first.

### DIFF
--- a/clients/nm-online.c
+++ b/clients/nm-online.c
@@ -40,6 +40,7 @@
 
 #define PROGRESS_STEPS 15
 
+#define EXIT_NONE               -1
 #define EXIT_FAILURE_OFFLINE     1
 #define EXIT_FAILURE_ERROR       2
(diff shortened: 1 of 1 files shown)


thom311:
merged to master

PR #153, opened 2018-06-29
dhcp/dhclient-utils: skip over dhclient.conf blocks

lkundrak:
Extend the lame-ass dhclient.conf parser to ignore the blocks we can't
do anything useful about: alias{}, pseudo{} and even lease{}.

Note that there's still a lot of cases we can't handle without a
full-fledged dhclient.conf parser -- notably the files that don't use
line breaks to separate the statements.

That is probably okay -- the whole thing is probably mostly useless and
we shall ever bother only about cases that actually cause trouble.

### DIFF
--- a/src/dhcp/nm-dhcp-dhclient-utils.c
+++ b/src/dhcp/nm-dhcp-dhclient-utils.c
@@ -316,6 +316,7 @@ nm_dhcp_dhclient_create_config (const char *interface,
 
 	if (orig_contents) {
 		char **lines, **line;
+		int nest = 0;
 		gboolean in_alsoreq = FALSE;
 		gboolean in_req = FALSE;
 		char intf[IFNAMSIZ];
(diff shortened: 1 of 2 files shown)


thom311:
Could you explain what prompted you to do this patch or what brought this issue to your attention? Was there a bugreport (is there an URL)? (if it makes sense, not explain here, but in the commit message).

lkundrak:
> Could you explain what prompted you to do this patch or what brought this issue to your attention? Was there a bugreport (is there an URL)? (if it makes sense, not explain here, but in the commit message).

Not really. This was a over a private conversation, without a record kept. But I guess it's not a problem, it's sort of obvious that the reason was that someone was actually had dhclient.conf with one of the blocks that we don't grok ("alias {}").

thom311:
the entire (pre-existing) parsing is not nice, because it does not justice to the complexity of the file format. It's easy to write configurations that confuse the parser.

Anyway, the patch itself lgtm.

PR #154, opened 2018-06-29
lr/hotspot hostname: default hotspot SSID to hostname

lkundrak:
Both Cinnamon and GNOME always forcibly set the hotspot SSID to a hostname on activation. They override what the user has set, which is not too nice. However they have no way of telling whether the actual wifi.ssid is what the user has set, or what the hostname was previously.

Make their lives easier -- allow keeping the SSID blank on hotspots and just tell supplicant to use a current (pretty) hostname.

### DIFF
--- a/libnm-core/nm-setting-wireless.c
+++ b/libnm-core/nm-setting-wireless.c
@@ -752,25 +752,6 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
 	gsize length;
 	GError *local = NULL;
 
-	if (!priv->ssid) {
-		g_set_error_literal (error,
-		                     NM_CONNECTION_ERROR,
-		                     NM_CONNECTION_ERROR_MISSING_PROPERTY,
-		                     _("property is missing"));
-		g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
-		return FALSE;
-	}
-
-	length = g_bytes_get_size (priv->ssid);
-	if (length == 0 || length > 32) {
-		g_set_error_literal (error,
-		                     NM_CONNECTION_ERROR,
-		                     NM_CONNECTION_ERROR_INVALID_PROPERTY,
-		                     _("SSID length is out of range <1-32> bytes"));
-		g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
-		return FALSE;
-	}
-
 	if (priv->mode && !g_strv_contains (valid_modes, priv->mode)) {
 		g_set_error (error,
 		             NM_CONNECTION_ERROR,
@@ -779,6 +760,29 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
 		             priv->mode);
 		g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_MODE);
 		return FALSE;
+	}
+
+	if (!priv->mode || strcmp (priv->mode, NM_SETTING_WIRELESS_MODE_INFRA) == 0) {
+		if (!priv->ssid) {
+			g_set_error_literal (error,
+			                     NM_CONNECTION_ERROR,
+			                     NM_CONNECTION_ERROR_MISSING_PROPERTY,
+			                     _("property is missing"));
+			g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
+			return FALSE;
+		}
+	}
+
+	if (priv->ssid) {
+		length = g_bytes_get_size (priv->ssid);
+		if (length == 0 || length > 32) {
+			g_set_error_literal (error,
+			                     NM_CONNECTION_ERROR,
+			                     NM_CONNECTION_ERROR_INVALID_PROPERTY,
+			                     _("SSID length is out of range <1-32> bytes"));
+			g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
+			return FALSE;
+		}
 	}
 
 	if (priv->band && !g_strv_contains (valid_bands, priv->band)) {
--- a/src/devices/wifi/nm-wifi-ap.c
+++ b/src/devices/wifi/nm-wifi-ap.c
@@ -1228,7 +1228,7 @@ nm_wifi_ap_new_from_properties (const char *supplicant_path, GVariant *propertie
 }
 
 NMWifiAP *
-nm_wifi_ap_new_fake_from_connection (NMConnection *connection)
+nm_wifi_ap_new_fake (NMConnection *connection, GBytes *ssid)
 {
 	NMWifiAP *ap;
 	NMWifiAPPrivate *priv;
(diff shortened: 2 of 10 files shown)


src/supplicant/nm-supplicant-config.c
```diff
@@ -452,13 +452,13 @@ gboolean
 nm_supplicant_config_add_setting_wireless (NMSupplicantConfig * self,
                                            NMSettingWireless * setting,
                                            guint32 fixed_freq,
+                                           const GByteArray *ssid,
```
thom311: `GByteArray` is a mutable array. This should be a `GBytes` type, which is immutable. I guess, you did this, because of `nm_wifi_ap_get_ssid()`. But it's really NMWifiAP who does it wrong already and should be fixed. See how even `nm_wifi_utils_complete_connection()` is wrong, and at first need to convert the `GByteArray` argument to a `GBytes`. Let's not proliferate the use of the ill suited type, but rather fix it.

thom311: Another reason, `GBytes` being immutable, there is little reason to ever clone it, just take another ref. Avoiding pointless cloning at
```C
nm_wifi_ap_set_ssid (ap, g_bytes_get_data (ssid, NULL), g_bytes_get_size (ssid));
```

src/devices/wifi/nm-wifi-ap.c
```diff
@@ -1213,32 +1212,29 @@ nm_wifi_ap_new_fake_from_connection (NMConnection *connection)
...
 
 	mode = nm_setting_wireless_get_mode (s_wireless);
-	if (mode) {
-		if (!strcmp (mode, "infrastructure"))
-			nm_wifi_ap_set_mode (ap, NM_802_11_MODE_INFRA);
-		else if (!strcmp (mode, "adhoc"))
+	if (!mode || !strcmp (mode, "infrastructure")) {
+		nm_wifi_ap_set_mode (ap, NM_802_11_MODE_INFRA);
+	} else {
+		if (!strcmp (mode, "adhoc"))
```
thom311: let's do `if-else-if` without increasing the indentation level... this is a classical selection out of multiple values, it's visually nicer, if they are all presented at the same level.

libnm-core/nm-setting-wireless.c
```diff
@@ -745,25 +745,6 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
...
-	if (length == 0 || length > 32) {
-		g_set_error_literal (error,
-		                     NM_CONNECTION_ERROR,
-		                     NM_CONNECTION_ERROR_INVALID_PROPERTY,
-		                     _("SSID length is out of range <1-32> bytes"));
-		g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
-		return FALSE;
-	}
-
 	if (priv->mode && !g_strv_contains (valid_modes, priv->mode)) {
```
thom311: could we replace `valid_modes` with `NM_IN_STRSET()`?

libnm-core/nm-setting-wireless.c
```diff
@@ -774,6 +755,27 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
 		return FALSE;
 	}
 
+	if (!priv->mode || strcmp (priv->mode, NM_SETTING_WIRELESS_MODE_INFRA) == 0) {
```
thom311: why does it allow missing SSID for `NM_SETTING_WIRELESS_MODE_ADHOC`? That seems wrong and it should not check for INFRA mode, but instead explicitly check for not AP.

libnm-core/nm-setting-wireless.c
```diff
@@ -774,6 +755,27 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
...
+		if (!priv->ssid) {
+			g_set_error_literal (error,
+			                     NM_CONNECTION_ERROR,
+			                     NM_CONNECTION_ERROR_MISSING_PROPERTY,
+			                     _("property is missing"));
+			g_prefix_error (error, "%s.%s: ", NM_SETTING_WIRELESS_SETTING_NAME, NM_SETTING_WIRELESS_SSID);
+			return FALSE;
+		}
+
+		length = g_bytes_get_size (priv->ssid);
```
thom311: the length validation is now no longer done for AP/ADHOC. That is wrong. If a SSID is present, it must always be checked for being valid.

thom311:
> Both Cinnamon and GNOME always forcibly set the hotspot SSID to a hostname on activation. They override what the user has set, which is not too nice. However they have no way of telling whether the actual wifi.ssid is what the user has set, or what the hostname was previously.

Don't they anyway do the wrong thing? Why would they change the SSID of an existing profile? It means, if the user previously already configured the hotspot on his phone, he has to re-configure it. That doesn't seem desirable behaviour -- regardless, whether the user intentionally adjusted the SSID, or whether the hostname changed since last time.

Also, is the hostname really the best choice for the SSID? IMO creating a hotspot is anyway not zero-config. Meaning, the user must at least read the generated PSK (likely, he even wants to configure even a more memorable PSK, like "12345678"). So, user clicks on hotspot button to create a new profile, and he also needs to note down the randomly generated PSK. At that point, he can just as well not a randomly generated SSID. And SSID that does not make the user so easily recognizable. I question, whether by default using the hostname as SSID is preferable. I certainly wouldn't like it, I'd rather have a randomly generated SSID (and hopefully, the city is full of Gnome users, that have similarly generated hotspots and I blend nicely in).

Does this branch really solve the right problem? Is it really valuable, to allow the tool/user that creates the profile, to omit selecting an SSID, and let it instead be autogenerated by NM?

What might make some sense, to generate a SSID based on the stable-id (or connection ID). Then, if the city is full of NetworkManager hotspots, I will again blend in nicely. The advantage here is, if Gnome shell uses a different scheme to generate randomized SSIDs than Plasma, I only blend with Gnome users. If NM generates a stable SSID, then I look like most other NM users.

thom311:
Maybe I didn't make that so clear: the advantage of the SSID being generated based on the connection-id/stable-id, is that it doesn't change. When I create a hotspot, it's already cumbersome to configure it in the client. I don't want the SSID to change. It's a misfeature to have it based on the hostname and subject to change.
That the generated SSD looks less recognizable than a hostname, is an added value.

Yes, a less recognizable SSID might be a slight inconvenience, when I need to configure the network on my phone. However, I already need to type in the PSK, which an order of magnitude more inconvenient already, than having to read the SSID from the Gnome display and search network "wifi-382132897" on my phone. Or do you plan to implement WPS for Wi-Fi APs?

src/supplicant/nm-supplicant-config.c
lkundrak: I suppose this is what #182 does?

libnm-core/nm-setting-wireless.c
lkundrak: Perhaps. I'm not touching those lines.

libnm-core/nm-setting-wireless.c
lkundrak: I can't see why the Ad-Hoc networks shouldn't default to the hostname. On the contrary, it seems useful. At least that's what the Android handsets used to do iirc.

libnm-core/nm-setting-wireless.c
lkundrak: Good catch. Will fix.

src/devices/wifi/nm-wifi-ap.c
lkundrak: Okay.

lkundrak:
>> Both Cinnamon and GNOME always forcibly set the hotspot SSID to a hostname on activation.
> Don't they anyway do the wrong thing?

They do the best they can I suppose.

Whether using the hostname at all is the "right" thing might be a matter of everyone's taste (I think it's all right), but as a matter of fact virtually *everyone* (that is, Android, Windows, iPhone, GNOME, Cinnamon) is doing this. The pretty hostname is actually a pretty good identifier of a computer.

I prefer to leave your follow-up comment without a detailed response, because I think it diverts the discussion too far in a direction away from the original problem.

thom311:
> but as a matter of fact virtually everyone (that is, Android, Windows, iPhone, GNOME, Cinnamon) is doing this

It's not clear to me what you mean with **this**. The main point is that the SSID of an existing profile gets changed (or activating the hotspot multiple times, might yield different SSIDs). Whether the hostname is a good default choice when creating the hotspot profile is mostly a separate question. 

AFAICS, my Android phone does not dynamically change the SSID for my hotspot when I activate the hotspot. I don't remember what the original default choice was when I first created the hotspot. Can you provide more information for your claim?

> I prefer to leave your follow-up comment without a detailed response

Good. But could you better explain why the new feature of this branch is desirable? "They do the best they can I suppose" does not seem like a good rationale -- I previously tried to elaborate previously why it is undesirable.

bengal:
> > but as a matter of fact virtually everyone (that is, Android, Windows, iPhone, GNOME, Cinnamon) is doing this
> 
> It's not clear to me what you mean with **this**. The main point is that the SSID of an existing profile gets changed (or activating the hotspot multiple times, might yield different SSIDs). Whether the hostname is a good default choice when creating the hotspot profile is mostly a separate question.

I tend to agree with Thomas; if the SSID is not specified, a hostname change causes a different SSID to be used, which prevents clients configured with the old SSID to associate. This seems highly unexpected to me; if gnome-control-center wants to enforce this behavior, fine but I don't think we should hardcode it in NM.

PR #157, opened 2018-07-03
lr/conn-check-af: reliably check connectivity per device, separately for IPv4 and IPv6

lkundrak:


vbenes:
there was a crash when testing this PR 
http://faf.lab.eng.brq.redhat.com/faf/problems/136390/

man/NetworkManager.conf.xml
```diff
@@ -306,15 +306,13 @@ no-auto-default=*
...
-        <filename>/usr/lib/systemd/resolv.conf</filename>.
-        In that case, <literal>systemd-resolved</literal> is chosen automatically.
-        </para>
+        <listitem><para>Configures the processing mode of DNS information known to
+        NetworkManager. Takes a value in
+        <arg choice='plain'><replaceable>mode</replaceable></arg><arg>,systemd-resolved</arg>
+        form. The primary <replaceable>mode</replaceable> influences how is the system's
+        <filename>resolv.conf</filename> managed, while the supplementary systemd-resolved
+        mode allows passing the DNS configuration to systemd-resolved even if it's not
+        configured as a primary resolver. The following modes are available:</para>
```
thom311: it seems not optimal to configure it this way. the only setting that makes sense, is either having
```
dns=$ONE_VALUE
```
or
```
dns=$ONE_VALUE_NOT_SDRD,systemd-resolved
```
meaning: you either select one mode; or you select a mode != `systemd-resolved` and followed by `systemd-resolved`.

Wouldn't it be clearer to add separate option like:
```
dns-internal=systemd-resolved
```
?

src/nm-connectivity.c
```diff
@@ -483,35 +622,30 @@ nm_connectivity_check_start (NMConnectivity *self,
...
+		                          G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS,
+		                          NULL,
+		                          "org.freedesktop.resolve1",
+		                          "/org/freedesktop/resolve1",
+		                          "org.freedesktop.resolve1.Manager",
+		                          cb_data->concheck.resolve_cancellable,
+		                          resolved_proxy_created,
+		                          cb_data);
+
+		_LOG2D ("start request to '%s'", priv->uri);
```
thom311: Let's just `return cb_data;` at this point. Then the dangling `else` in the `#if` block is not needed. It's error-prone.

src/nm-connectivity.c
```diff
@@ -483,35 +622,30 @@ nm_connectivity_check_start (NMConnectivity *self,
...
-			cb_data->timeout_id = g_timeout_add_seconds (20, _timeout_cb, cb_data);
-
-			_LOG2D ("start request to '%s'", priv->uri);
-			return cb_data;
-		}
+	if (iface && ifindex > 0 && priv->enabled && priv->host) {
+		cb_data->concheck.ifindex = ifindex;
+		cb_data->concheck.resolve_cancellable = g_cancellable_new ();
+
+		g_dbus_proxy_new_for_bus (G_BUS_TYPE_SYSTEM,
```
thom311: usually, we don't create new D-Bus proxies for each request. Also, we usually watch `name-owner` property, to know whether name resolving is available or not.

src/nm-connectivity.c
```diff
@@ -458,8 +466,139 @@ _idle_cb (gpointer user_data)
...
+	proxy = g_dbus_proxy_new_for_bus_finish (res, &error);
+	if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+		return;
+
+	self = cb_data->self;
+	priv = NM_CONNECTIVITY_GET_PRIVATE (self);
+
+	if (!proxy) {
+		/* Log a warning, but still proceed without systemd-resolved */
+		_LOG2W ("failed to connect to resolved via DBus: %s", error->message);
```
thom311: I think logging warnings/error at such low layers is not nice. But the problem is, that we might make *a lot* of connectivity checks, so if you mask sd-resolved service, a lot of warnings will be logged. `<debug>` should be enough.

src/nm-connectivity.c
```diff
@@ -458,8 +466,139 @@ _idle_cb (gpointer user_data)
...
+		_LOG2D ("can't resolve a name via systemd-resolved: %s", error->message);
+		do_curl_request (cb_data);
+		return;
+	}
+
+	addresses = g_variant_get_child_value (result, 0);
+	no_addresses = g_variant_n_children (addresses);
+	g_variant_unref (result);
+
+	for (i = 0; i < no_addresses; i++) {
```
thom311: differing data types: `int` vs `gsize`.

src/nm-connectivity.c
```diff
@@ -458,8 +466,139 @@ _idle_cb (gpointer user_data)
...
+		g_variant_get_child (addresses, i, "(ii@ay)", &ifindex, &family, &address);
+		address_buf = g_variant_get_fixed_array (address, &len, 1);
+
+		if (   (family == AF_INET && len == sizeof (struct in_addr))
+		    || (family == AF_INET6 && len == sizeof (struct in6_addr))) {
+			inet_ntop (family, address_buf, str, sizeof (str));
+			host = g_strdup_printf ("%s:%s:%s", priv->host,
+						priv->port ? priv->port : "80", str);
```
thom311: `priv->port ?: "80"`

src/nm-connectivity.c
```diff
@@ -458,8 +466,139 @@ _idle_cb (gpointer user_data)
...
+	no_addresses = g_variant_n_children (addresses);
+	g_variant_unref (result);
+
+	for (i = 0; i < no_addresses; i++) {
+		g_variant_get_child (addresses, i, "(ii@ay)", &ifindex, &family, &address);
+		address_buf = g_variant_get_fixed_array (address, &len, 1);
+
+		if (   (family == AF_INET && len == sizeof (struct in_addr))
+		    || (family == AF_INET6 && len == sizeof (struct in6_addr))) {
+			inet_ntop (family, address_buf, str, sizeof (str));
```
thom311: FYI, there is also `nm_utils_inet_ntop()` and `NM_UTILS_INET_ADDRSTRLEN` instead of `INET6_ADDRSTRLEN`.

src/nm-connectivity.c
```diff
@@ -458,8 +466,139 @@ _idle_cb (gpointer user_data)
...
+		return;
+	}
+
+	addresses = g_variant_get_child_value (result, 0);
+	no_addresses = g_variant_n_children (addresses);
+	g_variant_unref (result);
+
+	for (i = 0; i < no_addresses; i++) {
+		g_variant_get_child (addresses, i, "(ii@ay)", &ifindex, &family, &address);
+		address_buf = g_variant_get_fixed_array (address, &len, 1);
```
thom311: does this ensure, that `address_buf` is properly aligned? Does POSIX' `inet_ntop` allow pointers to `in_addr_t` / `struct in6_addr` not being properlty aligned? I don't know. Let's just use the functions from `shared/nm-utils/unalinged.h` to be sure.

thom311:
> Could you rebase to resolve the conflicts and pass CI?

Ah. CI passed. Ignore me. Still, rebase would be nice, because there are conflicts.

man/NetworkManager.conf.xml
lkundrak: I don't know. Reusing `dns` for both seems more future-proof to me.

There doesn't seem to be a reason (apart from current implementation limitations) why you wouldn't keep an instances of dnsmasq, sd-resolved and perhaps some other dns forwarders all configured at the same time while perhaps defaulting to configuring the system resolver not to use them.

src/nm-connectivity.c
thom311: I didn't check, but I would expect that the returned pointer lies inside a larger block of memory that belongs to the GVariant. Why do you think that would be aligned to something particular? Especially, because on D-Bus this is received as a block of mere bytes ("ay") and because you just request an array with `element_size` 1. I wouldn't expect stronger alignment guarantees than for the `element_size` argument (that is, the "y" in "ay").

But ok, may you be right. Is there a problem with playing safe and copy it once with

```C
    NMIPAddr addr_aligned;

    memcpy (&addr_aligned, address_buf, len);
```

src/nm-connectivity.c
thom311: > And I believe it's the wrong thing to do. Due to the asynchronous nature of DBusProxy creation, we'd need to make NMConnectivityCheck it async initable or hold back all 

The connectivity check is async already and tracked via that a `NMConnectivityCheckHandle` instance. If you create one proxy for all requests, that means, you need queuing handles while creating the proxy. Sure, that's mildly complicated, but seems very doable. For example, `src/nm-firewall-manager.c` does exactly that, in a very similar manner.

Not sure what you mean with `async initiable`. `NMConnectivityCheckHandle` is not a `GObject` and wouldn't implement `GAsyncInitable`.

lkundrak: Yeah, I think NMFirewallManager is pretty terrible. It needs a global/singleton to keep track of in-flight requests. That I'd prefer to avoid.

thom311: It does not need a global/singleton at all, it needs a NMFirewallManager instance -- which, since we only create one of such instance, is indeed a singleton. But there is nothing in code preventing you to create arbitrary many NMFirewallManager instances, all making their own requests.

I also don't see where it differs from NMConnectivity (as it is already today). Granted, firewall-manager makes an async D-Bus request, while nm-connectivity makes and async wget (+ creating a D-Bus proxy asynchronously). Apart from that their asynchronous operations do, they work fundamentally similar. They are also both "singleton" instances, and keep both track of their in-flight async handles.

I'm curious why you think NMFirewallManager is terrible and how it could be better. Could you elaborate?

lkundrak: > I'm curious why you think NMFirewallManager is terrible and how it could be better. Could you elaborate?

It needs a logic to keep track of whether the proxy construction of the proxy was successful and track of in-flight requests. That adds extra moving parts at questionable benefit.

thom311: With this reply I don't understand what you exactly suggest to improve it. Seem it will stay "terrible" and I will continue adding such code.

The main point of components like `NMFirewallManager` is to provide a *simple*, *correct* and *efficient* API to the upper layers. I clam it succeeds at doing that with mere 600 LOC, which is in the realm of being understandable. That is different from it's callers (`NMManager`(*) and `NMDevice`) which are complex beyond words. Looking at `NMFirewallManager` and saying that it's too complicated misses an important point: `NMFirewallManager` takes over a reasonable(?) amount of complexity, to make `NMManager` and `NMDevice` simpler. 

As can be seen with:
```
  $ git grep -30 nm_firewall_manager_ -- ':(exclude)src/nm-firewall-manager.?'
```

(*) `NMManager` does not call firewall, but `NMPolicy` does. However, `NMPolicy` *is* essentially a part of `NMManager`. There is no clear separation of responsibilities or API between them and they are strongly intermingled.

man/NetworkManager.conf.xml
bengal: I don't see so useful to allow multiple DNS plugins at the same time in
an arbitrary way. The only special exception should be systemd-resolved,
which is the only one that can be used to resolve names for connectivity
check. So, I think we should use it by default for connectivity check
(along with the one specified by the dns= key to be used system-wide).

If we want to allow disabling sd-resolved for connectivity check, we
could add a new key (like Thomas said) as 'dns-internal', or
'concheck-dns' (but perhaps this is too specific and in the future we'll
use it also for something else). The allowed values would be 'none' and
'systemd-resolved', with the latter being the default.

src/nm-connectivity.c
bengal: An implementation using a single proxy is probably a bit more efficient
(how much, I can't say) but for sure it adds more complexity. In my
opinion, creating a proxy for each request is good enough and keeps the
code simple.

man/NetworkManager.conf.xml
lkundrak: Okay. I'm going with systemd-resolved=yes/no, because the setting is not specific to connectivity checking.

lkundrak:
Pushed an updated branch.

vbenes:
1. Fail in NM CI was caused by new test fixed in another branch.
2. Was this crash addressed? http://faf.lab.eng.brq.redhat.com/faf/reports/7034/ It looks like we saw it again last week.

PR #159, opened 2018-07-06
manager: accept non-null device for VPN activations

bengal:
Commit 10753c36168a ("manager: merge VPN handling into
_new_active_connection()") added a check to fail the activation of
VPNs when a device is passed to ActivateConnection(), since the device
argument is ignored for VPNs.

This broke activating VPNs from nm-applet as nm-applet sets both the
specific_object (parent-connection) and device arguments in the
activation request.

Note that we already check in _new_active_connection() that when a
device is supplied, it matches the device of the parent
connection. Therefore, the check can be dropped.

Reported-by: Michael Biebl <biebl@debian.org>
Fixes: 10753c36168a82cd658df8a7da800960fddd78ed

### DIFF
--- a/src/nm-manager.c
+++ b/src/nm-manager.c
@@ -4611,21 +4611,7 @@ validate_activation_request (NMManager *self,
 		}
 	}
 
-	if (is_vpn && device) {
-		/* VPN's are treated specially. Maybe the should accept a device as well,
-		 * however, later on during activation, we don't handle the device.
-		 *
-		 * Maybe we should, and maybe it makes sense to specify a device
-		 * when activating a VPN. But for now, just error out.  */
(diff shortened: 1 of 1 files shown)


vbenes:
any idea how to cover with test?

lamby:
(This is https://bugs.debian.org/903109 in Debian)

bengal:
> any idea how to cover with test?

Activate a VPN connection specifying the base device:

```
 $ nmcli connection up vpn1 ifname eth0
 Error: Connection activation failed: Cannot specify device when activating VPN

```

thom311:
lgtm

bengal:
Patch applied as e205664ba8c25939f1678d1b078a67989c180046.

PR #160, opened 2018-07-09
device: disable IPv6 on links when they're not ready for autoconfiguration

lkundrak:
Otherwise kernel logs a line every time we up a device that is not ready
for IPv6 autoconfiguration (that is every 5 minutes when a random MAC
address is genreated for Wi-Fi scanning), flooding its message buffer:

  IPv6: ADDRCONF(NETDEV_UP): wlan0: link is not ready

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -9192,14 +9192,6 @@ restore_ip6_properties (NMDevice *self)
 }
 
 static inline void
-set_disable_ipv6 (NMDevice *self, const char *value)
-{
-	/* We only touch disable_ipv6 when NM is not managing the IPv6LL address */
-	if (!NM_DEVICE_GET_PRIVATE (self)->ipv6ll_handle)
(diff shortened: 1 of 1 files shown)


thom311:
In principle this looks right to me...

but it seems so complicated to me.

E.g. remember [rh#1585078](https://bugzilla.redhat.com/show_bug.cgi?id=1585078#c5) and [pr#144](https://github.com/NetworkManager/NetworkManager/pull/144#issuecomment-401384028). There -- if kernel supports addrgenmode -- we would set `addrgenmode=none` for disconnected devices. Later, if the user starts configuring the device (and NM assumes and external connection), the question is, whether NM should stay away from the device, or reset the `addrgenmode=eui64` for the user. This goes even a step further: it does not only set `addrgenmode=none`, it even disables IPv6 on disconnected devices. Is that right? What are your thoughts about that?

btw, could you rebase the patch, because CI failed.

lkundrak:
Fixed in kernel instead.

PR #161, opened 2018-07-09
platform: put wireguard_family_id in NMPObjectLink

jbeta:
Limit the lifetime of the cached genl family ID to the lifetime of the
interface so we correctly handle module reloads.

Suggested by @thom311 in #143. This is short and simple - but if it's not desirable to query often, then we should be able to copy the ID across link data refetches, e.g. around:
https://github.com/NetworkManager/NetworkManager/blob/73fa1b54ef98d70402ea91e88fd15298eb151fb7/src/platform/nm-linux-platform.c#L2275

### DIFF
--- a/src/platform/nm-linux-platform.c
+++ b/src/platform/nm-linux-platform.c
@@ -3402,8 +3402,6 @@ typedef struct {
 
 		gint is_handling;
 	} delayed_action;
-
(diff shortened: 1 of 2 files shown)


vhumpa:

thom311:
Merged, thanks!!!

PR #162, opened 2018-07-09
core,libnm: introduce NMDeviceWireguard

jbeta:
For now, the device only exposes partial link status (not including
peers). It cannot create new links.

This PR is mostly sanity check / request for comments. I'm also looking for guidance on how to model Peers/AllowedIPs on the D-Bus interface: would these be object paths to separate instances, a complex `a(ss...qta(...))` D-Bus signature that maps to a `NMWireguardPeer` array, neither...?


vhumpa:

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,48 @@
...
+
+    <!--
+        ListenPort:
+
+        Local UDP listening port.
+    -->
+    <property name="ListenPort" type="u" access="read"/>
+
+    <!--
+        Fwmark:
```
lkundrak: Wouldn't "FirewallMark" be a better name?

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,48 @@
...
+
+        Firewall mark.
```
lkundrak: Please add a sentence or two explaining what a Firewall mark is.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,48 @@
...
+    -->
+    <property name="Fwmark" type="u" access="read"/>
+
+    <!--
+        PrivateKey:
+
+        32-byte private WireGuard key.
+    -->
+    <property name="PrivateKey" type="v" access="read"/>
```
lkundrak: "v" signature ? Wouldn't "ay" make more sense for a byte array?

Also, the PrivateKey probably shouldn't be visible on the D-Bus API, since it has not access control and is visible to all local users. Is it necessary to have this?

libnm/libnm.ver
```diff
@@ -1383,6 +1383,11 @@ global:
 	nm_connection_get_setting_6lowpan;
 	nm_connection_get_setting_wpan;
 	nm_device_6lowpan_get_type;
+	nm_device_wireguard_get_public_key;
```
lkundrak: Please keep this sorted alphabetically.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+ * Gets the private key for this interface
+ *
+ * Returns: pointer to the 32-byte private key
+ *
+ * Since: 1.14
+ **/
+const guchar*
+nm_device_wireguard_get_private_key (NMDeviceWireguard *device)
+{
+	g_return_val_if_fail (NM_IS_DEVICE_WIREGUARD (device), 0);
```
lkundrak: You probably meant NULL instead of 0.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+ * Gets the public key for this interface
+ *
+ * Returns: pointer to the 32-byte public key
+ *
+ * Since: 1.14
+ **/
+const guchar*
+nm_device_wireguard_get_public_key (NMDeviceWireguard *device)
+{
+	g_return_val_if_fail (NM_IS_DEVICE_WIREGUARD (device), 0);
```
lkundrak: Ditto here, please s/0/NULL/.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+ * nm_device_wireguard_get_private_key:
+ * @device: a #NMDeviceWireguard
+ *
+ * Gets the private key for this interface
+ *
+ * Returns: pointer to the 32-byte private key
+ *
+ * Since: 1.14
+ **/
+const guchar*
```
lkundrak: I'm wondering if GBytes would perhaps be better than guchar here (esp. if the key can contain a NUL byte).
@thom311 ?

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+ * nm_device_wireguard_get_fwmark:
+ * @device: a #NMDeviceWireguard
+ *
+ * Gets the firewall mark (fwmark) byte, if this interface sets one
+ *
+ * Returns: 0 if fwmark not in use, fwmark byte value otherwise
+ *
+ * Since: 1.14
+ **/
+guint
```
lkundrak: guchar, if it's really a byte?

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+
+	NM_OBJECT_CLASS (nm_device_wireguard_parent_class)->init_dbus (object);
+
+	_nm_object_register_properties (object,
+	                                NM_DBUS_INTERFACE_DEVICE_WIREGUARD,
+	                                property_info);
+}
+
+static void
+finalize (GObject *object)
```
lkundrak: You don't need to implement finalize if you just chain it to the parent class.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+              GValue *value,
+              GParamSpec *pspec)
+{
+	NMDeviceWireguard *device = NM_DEVICE_WIREGUARD (object);
+	NMDeviceWireguardPrivate *priv = NM_DEVICE_WIREGUARD_GET_PRIVATE (object);
+	gchar *b64;
+
+	switch (prop_id) {
+	case PROP_PRIVATE_KEY:
+		b64 = g_base64_encode (priv->private_key, sizeof (priv->private_key));
```
lkundrak: Is there a particular reason to encode this in base64? Perhaps just exposing the raw bytes (as suggested above -- GBytes and "ay" on D-Bus) would make more sense.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,254 @@
...
+	 * NMDeviceWireguard:listen-port:
+	 *
+	 * Local UDP listen port.
+	 * Set to 0 to allow a random port to be chosen (default).
+	 *
+	 * Since: 1.14
+	 **/
+	obj_properties[PROP_LISTEN_PORT] =
+	    g_param_spec_uint (NM_DEVICE_WIREGUARD_LISTEN_PORT,
+	                       _("Listen port"),
```
lkundrak: In general, we leave these blank. We don't use them anywhere really and a translatable string just adds work for translators.

libnm/nm-device-wireguard.h
```diff
@@ -0,0 +1,72 @@
...
+
+#define NM_DEVICE_WIREGUARD_PRIVATE_KEY  "private-key"
+#define NM_DEVICE_WIREGUARD_PUBLIC_KEY   "public-key"
+#define NM_DEVICE_WIREGUARD_LISTEN_PORT  "listen-port"
+#define NM_DEVICE_WIREGUARD_FWMARK       "fwmark"
+
+/**
+ * NMDeviceWireguard:
+ */
+struct _NMDeviceWireguard {
```
lkundrak: For new devices, we like to keep these in the .c file, be cause they don't need to be part of the public API. See (nm-device-wpan for an example).

Also, when the object structure itself is not public, there's no need for the NMDeviceWireguardPrivate structure, since all NMDeviceWireguard fields would be private. (This is left for your consideration -- you may still want to keep the private structure for clarity.)

src/devices/nm-device-wireguard.c
```diff
@@ -0,0 +1,234 @@
+/* -*- Mode: C; tab-width: 4; indent-tabs-mode: t; c-basic-offset: 4 -*- */
+/* NetworkManager -- Network link manager
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
```
lkundrak: I'm wondering if I could ask you to license this under LGPL, the same as the libnm part? We sometimes move things around between libnm and core, and the inconsistent licensing has caused us trouble.

This also applies to some other GPL-licensed files.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,48 @@
...
+
+    <!--
+        PublicKey:
+
+        32-byte public WireGuard key.
+    -->
+    <property name="PublicKey" type="v" access="read"/>
+
+    <!--
```
lkundrak: Please just drop this, none of your properties emit the legacy signal.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,48 @@
...
+    -->
+    <property name="PrivateKey" type="v" access="read"/>
+
+    <!--
+        PublicKey:
+
+        32-byte public WireGuard key.
+    -->
+    <property name="PublicKey" type="v" access="read"/>
+
```
lkundrak: Please include a "Parent" property too.

src/devices/nm-device-wireguard.h
```diff
@@ -0,0 +1,46 @@
...
+#include "nm-device.h"
+
+#define NM_TYPE_DEVICE_WIREGUARD            (nm_device_wireguard_get_type ())
+#define NM_DEVICE_WIREGUARD(obj)            (G_TYPE_CHECK_INSTANCE_CAST ((obj), NM_TYPE_DEVICE_WIREGUARD, NMDeviceWireguard))
+#define NM_DEVICE_WIREGUARD_CLASS(klass)    (G_TYPE_CHECK_CLASS_CAST ((klass),  NM_TYPE_DEVICE_WIREGUARD, NMDeviceWireguardClass))
+#define NM_IS_DEVICE_WIREGUARD(obj)         (G_TYPE_CHECK_INSTANCE_TYPE ((obj), NM_TYPE_DEVICE_WIREGUARD))
+#define NM_IS_DEVICE_WIREGUARD_CLASS(klass) (G_TYPE_CHECK_CLASS_TYPE ((klass),  NM_TYPE_DEVICE_WIREGUARD))
+#define NM_DEVICE_WIREGUARD_GET_CLASS(obj)  (G_TYPE_INSTANCE_GET_CLASS ((obj),  NM_TYPE_DEVICE_WIREGUARD, NMDeviceWireguardClass))
+
+/* defined in the parent class, but exposed on D-Bus by the subclass. */
```
lkundrak: Well, actually not exposed. But I guess it should be.

src/devices/nm-device-wireguard.c
jbeta: Ah - don't worry, it's just a case of lazy copy-paste licensing. Happy to mirror the licenses for similar files.

libnm/nm-device-wireguard.h
jbeta: Understood. `NMDeviceWireguardPrivate` I was using while playing around with a `PROP_PEERS` implementation - but yeah I can drop it for now.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
jbeta: OK.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
lkundrak: Well, it still doesn't have to be base64 on D-Bus API -- it may make more sense to encode it in nmcli and possibly the keyfile plugin. @thom311, what do you think?

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
jbeta: This is a question I meant to ask too - what is the purpose of the parent device?

src/devices/nm-device-wireguard.h
jbeta: True.

libnm/nm-device-wireguard.c
jbeta: Remember key is fixed-size though. So it's down to the NM team's preference :)

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
thom311: I also would think that the private key is either a binary array (ay) or a string (s).

Maybe string is nicer, because then we can have there the base64 key directly. And if we some day in the future a different format (but why would we?), then we possibly could even extend the string with a "prefix:" (since ':' cannot be in plain base64).

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
lkundrak: On the second thought -- I have no idea if this is relevant for Wireguard. Maybe not.

Links that are associated with some other interface have this. E.g. VLANs, Tunneling devices or Veth pairs. In general it mirrors the IFLA_LINK property of the rtnetlink link. (In other words, if it says "link1@link0:" in "ip link", then link0 is the parent device.)

libnm/nm-device-wireguard.c
jbeta: Gah - no, it's not a byte, it's 32-bit. The explanation is wrong (in a few places too).

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
lkundrak: If we settle on base64, then it the documentation comment should state that it's base64-encoded.

libnm/nm-device-wireguard.c
jbeta: I see - thanks!

libnm/nm-device-wireguard.c
jbeta: Removing these then.

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
jbeta: > If we settle on base64, then it the documentation comment should state that it's base64-encoded.
That's fine.

Also, I hadn't looked into the access control for the D-Bus API yet. If indeed it's readable by all non-root users then of course we can't have the private key there :)

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
jbeta: In that case, I don't think it's relevant, no. Wireguard is layer 3, it doesn't care about another particular link.

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,256 @@
...
+	PROP_LISTEN_PORT,
+	PROP_FWMARK,
+);
+
+/**
+ * nm_device_wireguard_get_private_key:
+ * @device: a #NMDeviceWireguard
+ *
+ * Gets the private key for this interface
+ *
```
lkundrak: Please document that it's a base64-encoded here too. I still find it somewhat odd.
Also, the "32-byte" length might be somewhat impractical -- it would be better if it was NUL terminated. g_base64_encode() doesn't actually guarantee any particular length, since base64 allows using whitespace pretty arbitrarily.

jbeta: > Please document that it's a base64-encoded here too.

Well, for this particular function it isn't base64-encoded.

> Also, the "32-byte" length might be somewhat impractical -- it would be better if it was NUL terminated. g_base64_encode() doesn't actually guarantee any particular length, since base64 allows using whitespace pretty arbitrarily.

I probably didn't make this clear - WG keys are fixed-size 32-byte binary data (NULs allowed). base64 is only used to portably move them around and we're not really interested in the length of the base64 strings at this level.

In case I got the _code_ wrong, my _intention_ is to use base64 only in interfaces that are expected to be directly visible to end-users, as this is the standard across WG utilities. This includes NM logs (e.g. `nm_platform_lnk_wireguard_to_string`), and any CLI components. Honestly I'm not sure about the D-Bus interface, maybe it doesn't fall under this category in which case I agree base64 would be odd.

For all the other internal and external APIs we simply need to pass around fixed-size 32-byte buffers that can hold arbitrary data.

thom311:
ok to test

libnm/nm-device-wireguard.c
lkundrak: > My intention was to use base64 only in interfaces that are expected to be directly visible to end-users, as this is the standard across WG utilities. This includes NM logs (e.g. nm_platform_lnk_wireguard_to_string), and any CLI components. Honestly I'm not sure about the D-Bus interface, maybe it doesn't fall under this category in which case I agree base64 would be odd.

That makes sense. D-Bus is an API that's not exposed to the user, and I believe it shouldn't use base48. Agreed that logs/cli should.

> I'll update this PR.

Looking forward. I think this already looks pretty well and almost ready to e merged.

Makefile.am
```diff
@@ -234,6 +234,8 @@ introspection_sources = \
 	introspection/org.freedesktop.NetworkManager.Device.Vlan.h \
 	introspection/org.freedesktop.NetworkManager.Device.Vxlan.c \
 	introspection/org.freedesktop.NetworkManager.Device.Vxlan.h \
+	introspection/org.freedesktop.NetworkManager.Device.Wireguard.c \
+	introspection/org.freedesktop.NetworkManager.Device.Wireguard.h \
```
zx2c4: Wireguard -> WireGuard

(similar to WiMax.h below)

docs/api/Makefile.am
```diff
@@ -74,6 +74,7 @@ content_files = \
 	dbus-org.freedesktop.NetworkManager.Device.Veth.xml \
 	dbus-org.freedesktop.NetworkManager.Settings.xml \
 	dbus-org.freedesktop.NetworkManager.Device.Wired.xml \
+	dbus-org.freedesktop.NetworkManager.Device.Wireguard.xml \
```
zx2c4: WireGuard

docs/api/network-manager-docs.xml
```diff
@@ -203,8 +203,10 @@
       <xi:include href="dbus-org.freedesktop.NetworkManager.Device.Veth.xml"/>
       <xi:include href="dbus-org.freedesktop.NetworkManager.Device.Vlan.xml"/>
       <xi:include href="dbus-org.freedesktop.NetworkManager.Device.Vxlan.xml"/>
+      <xi:include href="dbus-org.freedesktop.NetworkManager.Device.Wireguard.xml"/>
```
zx2c4: WireGuard

introspection/meson.build
```diff
@@ -34,6 +34,7 @@ ifaces = [
   'org.freedesktop.NetworkManager.Device.Wpan',
   'org.freedesktop.NetworkManager.Device.Wired',
   'org.freedesktop.NetworkManager.Device.Wireless',
+  'org.freedesktop.NetworkManager.Device.Wireguard',
```
zx2c4: Wireguard -> WireGuard

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<node name="/">
+  <!--
+      org.freedesktop.NetworkManager.Device.Wireguard:
```
zx2c4: Wireguard -> WireGuard

introspection/org.freedesktop.NetworkManager.Device.Wireguard.xml
```diff
@@ -0,0 +1,32 @@
...
+      @short_description: WireGuard Device
+
+  -->
+  <interface name="org.freedesktop.NetworkManager.Device.Wireguard">
```
zx2c4: Wireguard -> WireGuard

libnm-core/nm-dbus-interface.h
```diff
@@ -73,6 +73,7 @@
 #define NM_DBUS_INTERFACE_DEVICE_MACVLAN       NM_DBUS_INTERFACE_DEVICE ".Macvlan"
 #define NM_DBUS_INTERFACE_DEVICE_PPP           NM_DBUS_INTERFACE_DEVICE ".Ppp"
 #define NM_DBUS_INTERFACE_DEVICE_VXLAN         NM_DBUS_INTERFACE_DEVICE ".Vxlan"
+#define NM_DBUS_INTERFACE_DEVICE_WIREGUARD     NM_DBUS_INTERFACE_DEVICE ".Wireguard"
```
zx2c4: Wireguard -> WireGuard

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,256 @@
...
+ */
+struct _NMDeviceWireguard {
+	NMDevice parent;
+};
+
+typedef struct {
+	NMDeviceClass parent;
+} NMDeviceWireguardClass;
+
+G_DEFINE_TYPE (NMDeviceWireguard, nm_device_wireguard, NM_TYPE_DEVICE)
```
zx2c4: Wireguard -> WireGuard

libnm/nm-device-wireguard.c
```diff
@@ -0,0 +1,256 @@
...
+/**
+ * NMDeviceWireguard:
+ */
+struct _NMDeviceWireguard {
+	NMDevice parent;
+};
+
+typedef struct {
+	NMDeviceClass parent;
+} NMDeviceWireguardClass;
```
zx2c4: Wireguard -> WireGuard

thom311:
@zx2c4 if the spelling is "WireGuard" with camel-case, then our style requires a separator at non-camel-case places. For example

- `./nm-device-wire-guard.h`
- `nm_device_wire_guard_get_fwmark()`
- `NM_DBUS_INTERFACE_DEVICE_WIRE_GUARD`
- `NM_DEVICE_WIRE_GUARD`

thom311:
@jbeta 

I wasn't available this week, sorry for the delay.

Would you consider fixup commits from branch https://github.com/NetworkManager/NetworkManager/tree/th/review/pr/162 for inclusion? If you happen to agree, just squash the patches. Thanks.

After that, it looks good to me. The last remaining question is the naming.

thom311:
> WiMax is /org.freedesktop.NetworkManager.WiMax and nm-device-wimax.h


I am aware of that. WiMax is also deprecated. The older the code, the less it conforms to our current style (which, slightly adjusts over time).

Anyway, I think it doesn't have so strict, and 
- `WireGuard`
- `/org.freedesktop.NetworkManager.WireGuard`
- `nm-device-wireguard.h`
- `nm_device_wireguard_get_fwmark()`

is fine with me.

@lkundrak WDYT?

jbeta:
> I wasn't available this week, sorry for the delay.

No worries - this week it was my turn :)

> Would you consider fixup commits from branch https://github.com/NetworkManager/NetworkManager/tree/th/review/pr/162 for inclusion? If you happen to agree, just squash the patches. Thanks.

Thanks for this. Most fixups I agree with and I'll update the PR with them. However taking into account [the above feedback](https://github.com/NetworkManager/NetworkManager/pull/162#discussion_r203309413) from @lkundrak, I think the public key should also become an array on both libnm and D-Bus APIs.

> Anyway, I think it doesn't have so strict, and
> 
> WireGuard
> /org.freedesktop.NetworkManager.WireGuard
> nm-device-wireguard.h
> nm_device_wireguard_get_fwmark()
> is fine with me.

That looks sane to me. `Wireguard` instead of `WireGuard` was just an oversight on my part, and breaking the name into two words is wrong for many reasons including practical. So if @zx2c4 also agrees let's go for this.

thom311:
> Thanks for this. Most fixups I agree with and I'll update the PR with them. However taking into account the above feedback from @lkundrak, I think the public key should also become an array on both libnm and D-Bus APIs.

Hi, I don't fully understand this reply. My suggestion (the fixup commits) do not affect that the public key is exposed on libnm and D-Bus API. It however removes unused API for the private key. I agree with how the public key is treated.

jbeta:
> Hi, I don't fully understand this reply. My suggestion (the fixup commits) do not affect that the public key is exposed on libnm and D-Bus API. It however removes unused API for the private key. I agree with how the public key is treated.

Correct. I was just pointing out that as per previous discussion with @lkundrak we should also remove base64 from the non-user-visible APIs. That's done now.

Updated the PR with:
- extra initial commit renaming the existing code (`s/Wireguard/WireGuard`)
- review changes from @thom311
- switch to "ay" signature for D-Bus API (tested with NUL bytes in the public key, no issues)

thom311:
Hi,

> switch to "ay" signature for D-Bus API (tested with NUL bytes in the public key, no issues)

I personally don't think this change was a good idea, because a base64 string is easier to work with. Anyway, the change is fine with me.

However, 
```C
const char *
nm_device_wireguard_get_public_key (NMDeviceWireguard *device)
{
    g_return_val_if_fail (NM_IS_DEVICE_WIREGUARD (device), NULL);

    return NM_DEVICE_WIREGUARD_GET_PRIVATE (device)->public_key;
}
```
public_key is no longer a NUL terminated string. You cannot expose it with such an API. It must be instead something like:
```C
const guint8 *nm_device_wireguard_get_public_key (NMDeviceWireguard *device, guint *out_length);
```
or better:
```C
GBytes *nm_device_wireguard_get_public_key (NMDeviceWireguard *device);
```

Also,
```C
static GVariant *
get_public_key_as_variant (NMDeviceWireGuard *device)
{
     NMDeviceWireGuardPrivate *priv;

     g_return_val_if_fail (NM_IS_DEVICE_WIREGUARD (device), NULL);
     priv = NM_DEVICE_WIREGUARD_GET_PRIVATE (device);

     return g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE,
                                       priv->public_key, sizeof (priv->public_key), 1);
}
```
is wrong. `sizeof(priv->public_key)` is the size of a pointer (4 or 8 bytes).

Also, in libnm's NMDeviceWireGuard, I think that PROP_PUBLIC_KEY should be of type `g_param_spec_boxed(..., G_TYPE_BYTES, ...)` instead of GVariant.

jbeta:
> I personally don't think this change was a good idea, because a base64 string is easier to work with.

Hopefully it's the other way around and we save consumers of these APIs a base64 decoding step. I'm OK with this being reverted later if turns out not to be the case.

> public_key is no longer a NUL terminated string. You cannot expose it with such an API. It must be instead something like:

The key size is always 32 bytes, so in practice you can. But I'll follow your advice and use GBytes.

> is wrong. sizeof(priv->public_key) is the size of a pointer (4 or 8 bytes).

True. Thanks for spotting this.

> Also, in libnm's NMDeviceWireGuard, I think that PROP_PUBLIC_KEY should be of type g_param_spec_boxed(..., G_TYPE_BYTES, ...) instead of GVariant.

Ah, I didn't realize that was an option. That's easier then.

Thanks again for reviewing!

thom311:
See `NM_ACCESS_POINT_SSID` as an example for a "ay" array in libnm.

> I'm OK with this being reverted later if turns out not to be the case.

Maybe "ay" + GBytes is indeed better. In the end, there are pros+cons. Leave it as "ay", ok?


> The key size is always 32 bytes, so in practice you can.

I think it's better to not commit the API to this. Also, I think the previous API could not be nicely exposed via GObject introspection (for example, the Python bindings wouldn't be correct).

jbeta:
Switched to GBytes + boxed param for libnm now.

zx2c4:
Looks good to me. Let's merge this so @jbeta can move onto part 3?

thom311:
merged to master.

Thanks Javier!!

PR #163, opened 2018-07-10
th/not-available-reason

thom311:
The aim is to give a better failure message when the profile is not available (either a particular device or on all devices):
```
        $ nmcli connection up my-wired
```
Of course, the are major problems in that we are unable to localize error messages for the client. That is hard to do and something we should do, but an English failure reason is still better then a very vague one.


thom311:
merged to master

PR #164, opened 2018-07-11
th/style-fixes-int-types

thom311:
Some adjustments to `contrib/scripts/checkpatch.pl` to enforce a certain style.

We commonly do all these points already, so they should not be controversial or new. But of course, they are up for discussion.


bengal:
LGTM

thom311:
Merged. Thanks!

PR #166, opened 2018-07-12
lr/rhel-connectivity: Connectivity checking for RHEL

lkundrak:
Not really there yet, but the bribe has been sent the web ops way.


thom311:
lgtm

bengal:
+1

PR #167, opened 2018-07-12
th/build-flags

thom311:
Various cleanup of the build.


shared/nm-default.h
```diff
@@ -22,6 +22,12 @@
 #ifndef __NM_DEFAULT_H__
```
lkundrak: > Subject: not-to-be-commited

Really?

Makefile.am
```diff
@@ -105,6 +105,8 @@ endif
 
 endif
 
+dflt_cppflags_std = -std=gnu99
```
lkundrak: NAK; it's not a preprocessor flag; not a "cppflag".
Also, how is defining an identifier for the option better than just using the option other than harming readability of an already really terrible makefile?

thom311: > NAK; it's not a preprocessor flag; not a "cppflag".

Because of autoconf. Look how the many uses of "CPPFLAGS" in Makefile.am. It's because we don't want to conflict with `AM_CFLAGS`, not because it's a preprocessor flag. Also, it's really just a name. If you dislike the naming, please suggest a name that is better suited. I don't mind renaming, but the name is consistent with what exists on master. Search master for `"dflt_cppflags_*"` and `"*_cppflags_*"`. If we are going to rename, those likely should be renamed as well.

> Also, how is defining an identifier for the option better than just using the option other than 
> harming readability of an already really terrible makefile?

it's better, because the variable does not only represent a compiler option, it represents a concept. We don't use `-std=gnu99` everywhere (for example there is no `-std` for `shared_libcsiphash_la_CPPFLAGS`), but only at our own code (+ systemd code). It also represents the concept of "some flags, that we add to almost every compilation". If you grep for gnu99`, you'll see that there is another use already: same option, different usage/concept. By having a variable, I can grep for all the places that have a similar use/meaning/concept, not only the same option.
Sure, the name of this variable may be badly chosen to represent the concept. I am bad with finding concise names in the 4800 lines of Makefile.am. Suggestions welcome.

bengal:
LGTM

thom311:
merged

PR #169, opened 2018-07-13
th/merge-gsystem-local-alloc

thom311:
Now that we no longer care about no downstream changes for our libgsystem fork, merge it with ours.


shared/nm-utils/gsystem-local-alloc.h
```diff
@@ -98,7 +98,7 @@ GS_DEFINE_CLEANUP_FUNCTION0 (GVariant *, gs_local_variant_unref, g_variant_unref
  * scope.
  */
 #define gs_free_variant_iter __attribute__ ((cleanup(gs_local_variant_iter_free)))
-GS_DEFINE_CLEANUP_FUNCTION_STRUCT(GVariantIter, gs_local_variant_iter_free, g_variant_iter_free)
```
lkundrak: Why not just drop them.

thom311: Same reason as later commit:
```
shared/gsystem-local-alloc: rename unused gs_* cleanup macros

[...]
    Don't drop them, despite they being unused. The reason is, that we should
    make use of cleanup functions more eagerly. Dropping them now -- because
    they are momentarily unused -- hampers using them in the future. We
    often don't use the cleanup macros at places where I think we should,
    so by dropping them, we hamper future use.
```

OK?

bengal:
LGTM

PR #171, opened 2018-07-23
lr/activation-error-hint: print a hint when an activation fails

lkundrak:
Works like this:

```
$ nmcli c up Lobotoland
Error: Connection activation failed: 802.1X supplicant took too long to authenticate
Hint: use 'journalctl -xe NM_CONNECTION=7afa0957-1575-4ac2-8c60-74b31efa4bbe + NM_DEVICE=wlp4s0' to get more details.
```

Plus a couple of fixes.

### DIFF
--- a/clients/cli/connections.c
+++ b/clients/cli/connections.c
@@ -2341,6 +2341,38 @@ typedef struct {
 	NMActiveConnection *active;
 } ActivateConnectionInfo;
 
+static void
+active_connection_hint (GString *return_text, ActivateConnectionInfo *info)
+{
+	NMRemoteConnection *connection;
+	nm_auto_free_gstring GString *hint = NULL;
+	const GPtrArray *devices;
+	int i;
+
+	if (strcmp(NM_CONFIG_DEFAULT_LOGGING_BACKEND, "journal") != 0)
+		return;
+
+	connection = nm_active_connection_get_connection (info->active);
+	g_return_if_fail (connection);
+
+	hint = g_string_new ("journalctl -xe ");
+	g_string_append_printf (hint, "NM_CONNECTION=%s",
+	                        nm_connection_get_uuid (NM_CONNECTION (connection)));
+
+	if (info->device) {
+		g_string_append_printf (hint, " + NM_DEVICE=%s", nm_device_get_iface (info->device));
+	} else {
+		devices = nm_active_connection_get_devices (info->active);
+		for (i = 0; i < devices->len; i++) {
+			g_string_append_printf (hint, " + NM_DEVICE=%s",
+			                        nm_device_get_iface (NM_DEVICE (g_ptr_array_index (devices, i))));
+		}
+	}
+
+	g_string_append (return_text, "\n");
+	g_string_append_printf (return_text, _("Hint: use '%s' to get more details."), hint->str);
+}
+
 static void activate_connection_info_finish (ActivateConnectionInfo *info);
 
 static void
@@ -2369,6 +2401,7 @@ check_activated (ActivateConnectionInfo *info)
 		nm_assert (reason);
 		g_string_printf (nmc->return_text, _("Error: Connection activation failed: %s"),
 		                 reason);
+		active_connection_hint (nmc->return_text, info);
 		nmc->return_value = NMC_RESULT_ERROR_CON_ACTIVATION;
 		activate_connection_info_finish (info);
 		break;
@@ -2491,6 +2524,7 @@ activate_connection_cb (GObject *client, GAsyncResult *result, gpointer user_dat
 		g_string_printf (nmc->return_text, _("Error: Connection activation failed: %s"),
 		                 error->message);
 		g_error_free (error);
+		active_connection_hint (nmc->return_text, info);
 		nmc->return_value = NMC_RESULT_ERROR_CON_ACTIVATION;
 		activate_connection_info_finish (info);
 	} else {
--- a/src/platform/nm-platform.c
+++ b/src/platform/nm-platform.c
@@ -57,22 +57,45 @@ G_STATIC_ASSERT (G_STRUCT_OFFSET (NMPlatformIPRoute, network_ptr) == G_STRUCT_OF
 
 #define _NMLOG_DOMAIN           LOGD_PLATFORM
 #define _NMLOG_PREFIX_NAME      "platform"
+
+
+#define NMLOG_COMMON(level, name, ...) \
+        char __prefix[32]; \
+        const char *__p_prefix = _NMLOG_PREFIX_NAME; \
+        const NMPlatform *const __self = (self); \
+        \
+        if (__self && NM_PLATFORM_GET_PRIVATE (__self)->log_with_ptr) { \
+            g_snprintf (__prefix, sizeof (__prefix), "%s[%p]", _NMLOG_PREFIX_NAME, __self); \
+            __p_prefix = __prefix; \
+        } \
+        _nm_log (__level, _NMLOG_DOMAIN, 0, name, NULL, \
+                 "%s: " _NM_UTILS_MACRO_FIRST (__VA_ARGS__), \
+                 __p_prefix _NM_UTILS_MACRO_REST (__VA_ARGS__));
+
 #define _NMLOG(level, ...) \
     G_STMT_START { \
         const NMLogLevel __level = (level); \
         \
         if (nm_logging_enabled (__level, _NMLOG_DOMAIN)) { \
-            char __prefix[32]; \
-            const char *__p_prefix = _NMLOG_PREFIX_NAME; \
-            const NMPlatform *const __self = (self); \
-            \
-            if (__self && NM_PLATFORM_GET_PRIVATE (__self)->log_with_ptr) { \
-                g_snprintf (__prefix, sizeof (__prefix), "%s[%p]", _NMLOG_PREFIX_NAME, __self); \
-                __p_prefix = __prefix; \
-            } \
-            _nm_log (__level, _NMLOG_DOMAIN, 0, NULL, NULL, \
-                     "%s: " _NM_UTILS_MACRO_FIRST (__VA_ARGS__), \
-                     __p_prefix _NM_UTILS_MACRO_REST (__VA_ARGS__)); \
+            NMLOG_COMMON(level, NULL, __VA_ARGS__);  \
+        } \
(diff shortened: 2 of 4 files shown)


src/platform/wifi/nm-wifi-utils-nl80211.c
```diff
@@ -40,9 +40,13 @@
 #define _NMLOG_PREFIX_NAME      "wifi-nl80211"
 #define _NMLOG(level, domain, ...) \
 	G_STMT_START { \
-		nm_log ((level), (domain), NULL, NULL, \
-		        "%s: " _NM_UTILS_MACRO_FIRST(__VA_ARGS__), \
-		        _NMLOG_PREFIX_NAME \
+		char _ifname_buf[IFNAMSIZ]; \
+		const char *_ifname = nl80211 ? nmp_utils_if_indextoname (nl80211->parent.ifindex, _ifname_buf) : NULL; \
+		\
+		nm_log ((level), (domain), _ifname ? _ifname : NULL, NULL, \
```
thom311: `_ifname ? _ifname : NULL`

src/platform/wifi/nm-wifi-utils-nl80211.c
```diff
@@ -40,9 +40,13 @@
 #define _NMLOG_PREFIX_NAME      "wifi-nl80211"
 #define _NMLOG(level, domain, ...) \
 	G_STMT_START { \
-		nm_log ((level), (domain), NULL, NULL, \
-		        "%s: " _NM_UTILS_MACRO_FIRST(__VA_ARGS__), \
-		        _NMLOG_PREFIX_NAME \
+		char _ifname_buf[IFNAMSIZ]; \
+		const char *_ifname = nl80211 ? nmp_utils_if_indextoname (nl80211->parent.ifindex, _ifname_buf) : NULL; \
```
thom311: in all(?) other `_NMLOG()` implementations, this `nl80211` variable is commonly called `self`.

src/platform/wifi/nm-wifi-utils-nl80211.c
```diff
@@ -958,46 +947,35 @@ nm_wifi_utils_nl80211_new (int ifindex, struct nl_sock *genl)
 
 	msg = nl80211_alloc_msg (nl80211, NL80211_CMD_GET_WIPHY, 0);
 
+	device_info.nl80211 = nl80211;
 	if (nl80211_send_and_recv (nl80211, msg, nl80211_wiphy_info_handler,
 	                           &device_info) < 0) {
-		_LOGD (LOGD_PLATFORM | LOGD_WIFI,
-		       "(%s): NL80211_CMD_GET_WIPHY request failed",
-		       ifname);
+		_LOGD (LOGD_PLATFORM | LOGD_WIFI, "NL80211_CMD_GET_WIPHY request failed");
```
thom311: since all callers use the same logging domain, `_NMLOG()` should not have this parameter explicitly. See for examle `src/devices/nm-acd-manager.c`. It should be something like:

```C
#define _NMLOG_DOMAIN             (LOGD_PLATFORM | LOGD_WIFI)  
#define _NMLOG_PREFIX_NAME        "wifi-nl80211"
#define _NMLOG(level, ...) \ 
```

src/platform/nm-platform.c
```diff
@@ -57,22 +57,45 @@ G_STATIC_ASSERT (G_STRUCT_OFFSET (NMPlatformIPRoute, network_ptr) == G_STRUCT_OF
 
 #define _NMLOG_DOMAIN           LOGD_PLATFORM
 #define _NMLOG_PREFIX_NAME      "platform"
+
+
+#define NMLOG_COMMON(level, name, ...) \
+        char __prefix[32]; \
```
thom311: let's always use a `G_STMT_START` block for multiline macros.

src/platform/nm-platform.c
```diff
@@ -57,22 +57,45 @@ G_STATIC_ASSERT (G_STRUCT_OFFSET (NMPlatformIPRoute, network_ptr) == G_STRUCT_OF
...
+            NMLOG_COMMON(level, name, __VA_ARGS__);  \
+        } \
+    } G_STMT_END
+
+#define _NMLOG3(level, ...)  \
+    G_STMT_START { \
+        const NMLogLevel __level = (level); \
+        \
+        if (nm_logging_enabled (__level, _NMLOG_DOMAIN)) { \
+            NMLOG_COMMON(level, nm_platform_link_get_name (self, ifindex), __VA_ARGS__);  \
```
thom311: `nm_platform_link_get_name()` will assert that `ifindex > 0`. It's not clear that that is the case for all callers (today and in the future).

clients/cli/connections.c
```diff
@@ -2336,6 +2336,43 @@ typedef struct {
 	NMActiveConnection *active;
 } ActivateConnectionInfo;
 
+#if SYSTEMD_JOURNAL
```
thom311: the logging backend is configurable at runtime. Maybe, NetworkManager should expose on D-Bus whether journal is used or not. For example via `NM_MANAGER_CAPABILITIES` properties.

clients/cli/connections.c
```diff
@@ -2336,6 +2336,43 @@ typedef struct {
...
+		devices = nm_active_connection_get_devices (info->active);
+		for (i = 0; i < devices->len; i++) {
+			g_string_append_printf (hint, " + NM_DEVICE=%s",
+			                        nm_device_get_iface (NM_DEVICE (g_ptr_array_index (devices, i))));
+		}
+	}
+
+	g_string_append (return_text, "\n");
+	g_string_append_printf (return_text, _("Hint: use '%s' to get more details."), hint->str);
+	g_string_free (hint, TRUE);
```
thom311: `nm_auto_free_gstring`

clients/cli/connections.c
```diff
@@ -2336,6 +2336,43 @@ typedef struct {
...
+}
+#else
```
thom311: could you avoid the `#if #else` and instead move the compile time conditional inside the function?

src/platform/wifi/nm-wifi-utils-nl80211.c
```diff
@@ -40,9 +40,13 @@
...
-		nm_log ((level), (domain), NULL, NULL, \
-		        "%s: " _NM_UTILS_MACRO_FIRST(__VA_ARGS__), \
-		        _NMLOG_PREFIX_NAME \
+		char _ifname_buf[IFNAMSIZ]; \
+		const char *_ifname = nl80211 ? nmp_utils_if_indextoname (nl80211->parent.ifindex, _ifname_buf) : NULL; \
+		\
+		nm_log ((level), (domain), _ifname ? _ifname : NULL, NULL, \
+		        "%s%s%s%s: " _NM_UTILS_MACRO_FIRST(__VA_ARGS__), \
+		        _NMLOG_PREFIX_NAME, \
+                        NM_PRINT_FMT_QUOTED (_ifname, " (", _ifname, ")", "") \
```
thom311: indentation

thom311:
(and semaphoreci seems to fail)

src/devices/nm-device.c
```diff
@@ -2296,7 +2296,7 @@ concheck_periodic_timeout_cb (gpointer user_data)
 {
 	NMDevice *self = user_data;
 
-	_LOGt (LOGD_CONCHECK, "connectivity: periodic timeout");
+	_LOGT (LOGD_CONCHECK, "connectivity: periodic timeout");
```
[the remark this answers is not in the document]
lkundrak: Yes, this looks like a mistake. Dropping it.

src/platform/wifi/nm-wifi-utils-nl80211.c
lkundrak: Yes, but not here...

src/platform/wifi/nm-wifi-utils-nl80211.c
lkundrak: Note that I didn't change anything about the logging domains.

src/platform/nm-platform.c
lkundrak: Why? NMLOG_COMMON is *not* to be used mimicking a statement.

clients/cli/connections.c
lkundrak: I don't think it is worth the complexity. This is a hint. If someone doesn't use journal on a platform with systemd-journald, they're on their own.

clients/cli/connections.c
lkundrak: I prefer not to, the preprocessor macros intertwined with C code harm readability.

clients/cli/connections.c
lkundrak: Okay...

clients/cli/connections.c
thom311: Like rhel-7 (which compiles with journal support, but `logging.backend` defaults to `syslog`). So, this is hardly an unusual setup.

src/platform/wifi/nm-wifi-utils-nl80211.c
thom311: Sorry for being unclear. Could you rename the "nl80211" variables to "self", which is what we do almost everywhere else?

src/platform/wifi/nm-wifi-utils-nl80211.c
thom311: You refactor `_NMLOG()`. At that point, it would be good to also get rid of the domain argument and move it inside the macro. Like done at most other places (notable exception, NMDevice, which actually wants to log with different domains).

src/platform/wifi/nm-wifi-utils-nl80211.c
lkundrak: Okay...

src/platform/wifi/nm-wifi-utils-nl80211.c
lkundrak: Okay...

clients/cli/connections.c
lkundrak: Sigh. This is still rather odd setup. The logs end up in journal anyway, only stripped of useful metadata :( Eventually, everybody should be logging to journal.

In an effort to attempt complicating things to the point we'll have to ask the server, I've changed the condition to look at whether journal logging is the default.

PR #173, opened 2018-07-27
lr/auth-dialog: improve dealing with VPN auth-dialogs

lkundrak:
There's basically two somewhat unrelated parts to this:

* Allowing auth-dialogs to consistently deal with newlines in values. Currently the auth dialog protocol doesn't deal with this and all producers of the protocol are producing ambiguous serializations. Oh well...

* Adding auth-dialog support to nmcli. One day this may allow us to get rid of the hardcoded secret handlers.

### DIFF
--- a/clients/common/nm-secret-agent-simple.c
+++ b/clients/common/nm-secret-agent-simple.c
@@ -31,6 +31,8 @@
 
 #include "nm-default.h"
 
+#include <gio/gunixoutputstream.h>
+#include <gio/gunixinputstream.h>
 #include <string.h>
 
 #include "nm-vpn-service-plugin.h"
@@ -58,6 +60,8 @@ typedef struct {
 	char                         **hints;
 	NMSecretAgentOldGetSecretsFunc callback;
 	gpointer                       callback_data;
+	GCancellable                  *cancellable;
+	NMSecretAgentGetSecretsFlags   flags;
 } NMSecretAgentSimpleRequest;
 
 typedef struct {
@@ -73,11 +77,23 @@ nm_secret_agent_simple_request_free (gpointer data)
 {
 	NMSecretAgentSimpleRequest *request = data;
 
+	g_clear_object (&request->cancellable);
 	g_object_unref (request->self);
 	g_object_unref (request->connection);
 	g_strfreev (request->hints);
 
 	g_slice_free (NMSecretAgentSimpleRequest, request);
+}
+
+static void
+nm_secret_agent_simple_request_cancel (NMSecretAgentSimpleRequest *request,
+                                       GError *error)
+{
+	NMSecretAgentSimplePrivate *priv = NM_SECRET_AGENT_SIMPLE_GET_PRIVATE (request->self);
+
+	request->callback (NM_SECRET_AGENT_OLD (request->self), request->connection,
+	                   NULL, error, request->callback_data);
+	g_hash_table_remove (priv->requests, request->request_id);
 }
 
 static void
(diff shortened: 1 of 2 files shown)


thom311:
> Equals sign was picked as a continuation character arbitratily.

I don't think this was arbitrarily. It's an ingenious extension of the existing protocol

thom311:
oh, and the example in the commit message is wrong:
```
      DATA_KEY=some-key
      DATA_VALUE=string
      =continued after a line break
      SECRET_KEY=key names
      =can have
      =continuations too
      SECRET_VALUE=value
```
it's `VAL`, not `VALUE`. And it lacks the `DONE`

clients/common/nm-secret-agent-simple.c
```diff
@@ -73,13 +77,26 @@ nm_secret_agent_simple_request_free (gpointer data)
...
+                                       GError *error)
+{
+	NMSecretAgentSimplePrivate *priv = NM_SECRET_AGENT_SIMPLE_GET_PRIVATE (request->self);
+
+	request->callback (NM_SECRET_AGENT_OLD (request->self), request->connection,
+			   NULL, error, request->callback_data);
+	g_hash_table_remove (priv->requests, request->request_id);
+}
+
+
```
thom311: spurious newline

clients/common/nm-secret-agent-simple.c
```diff
@@ -73,13 +77,26 @@ nm_secret_agent_simple_request_free (gpointer data)
...
 }
 
+static void
+nm_secret_agent_simple_request_cancel (NMSecretAgentSimpleRequest *request,
+                                       GError *error)
+{
+	NMSecretAgentSimplePrivate *priv = NM_SECRET_AGENT_SIMPLE_GET_PRIVATE (request->self);
+
+	request->callback (NM_SECRET_AGENT_OLD (request->self), request->connection,
+			   NULL, error, request->callback_data);
```
thom311: indentation

clients/common/nm-secret-agent-simple.c
```diff
@@ -443,6 +460,290 @@ add_vpn_secrets (NMSecretAgentSimpleRequest *request,
...
+	GPtrArray *secrets;
+	guint child_watch_id;
+	gulong cancellable_id;
+} AuthDialogData;
+
+static void
+_auth_dialog_data_free (AuthDialogData *data)
+{
+	g_ptr_array_unref (data->secrets);
+	g_spawn_close_pid (data->auth_dialog_pid);
```
thom311: this is not needed on Posix systems. Here the pid is a plain integer, not a handle. We don't need to pretend to run on anything but Linux.

clients/common/nm-secret-agent-simple.c
```diff
@@ -443,6 +460,290 @@ add_vpn_secrets (NMSecretAgentSimpleRequest *request,
...
+
+	if (!g_spawn_async_with_pipes (NULL, (char **)auth_dialog_argv, NULL,
+	                               G_SPAWN_DO_NOT_REAP_CHILD,
+	                               NULL, NULL,
+	                               &auth_dialog_pid,
+	                               &auth_dialog_in_fd,
+	                               &auth_dialog_out_fd,
+	                               NULL,
+	                               &error)) {
+		g_warning ("Failed to spawn the auth dialog%s\n", error->message);
```
thom311: let's not use g_warning(), it's not nice to nmcli/nmtui? Either handle the error (reporting it back a caller who can do something about it) or be silent.

clients/common/nm-secret-agent-simple.c
```diff
@@ -595,9 +899,7 @@ request_secrets_from_ui (NMSecretAgentSimpleRequest *request)
 		                     "Cannot service a secrets request %s for a %s connection",
 		                     request->request_id,
 		                     nm_connection_get_connection_type (request->connection));
-		request->callback (NM_SECRET_AGENT_OLD (request->self), request->connection,
-		                   NULL, error, request->callback_data);
-		g_hash_table_remove (priv->requests, request->request_id);
+		nm_secret_agent_simple_request_cancel (request, error);
```
thom311: it seems ugly, that the API sometimes invokes the callback synchronously and sometimes asynchronously.

clients/common/nm-secret-agent-simple.c
```diff
@@ -443,6 +460,290 @@ add_vpn_secrets (NMSecretAgentSimpleRequest *request,
...
+
+out:
+	if (error) {
+		nm_secret_agent_simple_request_cancel (request, error);
+	} else {
+		g_signal_emit (request->self, signals[REQUEST_SECRETS], 0,
+		               request->request_id, title, message, secrets);
+	}
+
+	g_clear_pointer (&keyfile, g_key_file_unref);
```
thom311: gs_unref_keyfile

thom311:
I added a unit-test for it, at th/pr/173 branch. Could you incorporate this branch into your pull-request?

(that just means: if you agree with th/pr/173, take the branch as-is as new HEAD for the pull-request)

clients/common/nm-secret-agent-simple.c
lkundrak: "It should be used on all platforms" -- glib Manual

That it doesn't do anything is an implementation detail of the current version. A future version of GLib may choose to change it to do something.

clients/common/nm-secret-agent-simple.c
lkundrak: This is equivalent.

clients/common/nm-secret-agent-simple.c
lkundrak: This is a "cannot happen" sort of a warning. If it happens, it's a bug. We should be loud, and this is about the only chance.

clients/common/nm-secret-agent-simple.c
lkundrak: Doesn't seem that ugly to me.

lkundrak:
> I added a unit-test for it, at th/pr/173 branch. Could you incorporate this branch into your pull-request?

Done

fgiudici:
The branch LGTM but there is a failure in NetworkManagerCI on pptp authentication that needs investigation before merging

lkundrak:
Ok, the CI failure is this: https://gitlab.gnome.org/GNOME/NetworkManager-pptp/merge_requests/5

I guess we shall skip the auth-dialog path if DISPLAY and WAYLAND_DISPLAY are unset for the time being, because *all* the plugins seem to be broken this way. Or perhaps add a flag to the .name file to signal that the auth-dialog works without a display server.

PR #174, opened 2018-07-30
agent-manager: order newer agents befor the old one

lkundrak:
This is a mere debugging convenience thing: e.g. if you run, but want to
check whether nm-applet or nmcli agent works fine, it's convenient that
the agent you run later gets a chance to deal with the secrets requests
first.

Is seems to do the job and is simpler that adding some more complicated
policy (e.g. introducing priorities or something).


bengal:
LGTM

thom311:
> agent-manager: upgrade a logging level of new agent notice
    
is this really useful for <info>? Shouldn't it at least contain some more information, like the PID, or an identifier for the secret-agent?

Anyway, lgtm too.

PR #175, opened 2018-07-30
lr/compiler-options: sync with network-manager-applet

lkundrak:
Basically applies here too:

Before:

```
$ sh autogen.sh
...
checking whether "-Wformat-nonliteral" works as expected... not supported
checking whether "-Wformat-security" works as expected... not supported
checking whether "-Wno-format-y2k" works as expected... not supported
```

After:

```
$ sh autogen.sh
...
checking whether "-Wformat-nonliteral" works as expected... yes
checking whether "-Wformat-security" works as expected... yes
checking whether "-Wno-format-y2k" works as expected... yes
```

### DIFF
--- a/m4/compiler_options.m4
+++ b/m4/compiler_options.m4
@@ -92,7 +92,7 @@ if test "$GCC" = "yes" -a "$set_more_warnings" != "no"; then
 		      -Wno-unused-parameter \
 		      ; do
 		dnl GCC 4.4 does not warn when checking for -Wno-* flags (https://gcc.gnu.org/wiki/FAQ#wnowarning)
-		_NM_COMPILER_FLAG([$(printf '%s' "$option" | sed 's/^-Wno-/-W/')], [],
+		_NM_COMPILER_FLAG([-Wall $(printf '%s' "$option" | sed 's/^-Wno-/-W/')], [],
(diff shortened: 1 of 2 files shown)


thom311:
I don't understand why this patch is the right fix.

The macro is called like
```
NM_COMPILER_WARNINGS(CFLAGS, ${more_warnings_default})
```
hence, `[$1]` is `CFLAGS` (in this case). I think, if you are going to modify `CFLAGS`, you need to safe and restore it, like in the lines before with `CFLAGS_SAVED="$$1"`


If all you want to do, is ensure that `_NM_COMPILER_FLAG` is called with `-Wall`, why don't you:
```
»···»···_NM_COMPILER_FLAG([-Wall $(printf '%s' "$option" | sed 's/^-Wno-/-W/')], [],
»···»···                  [CFLAGS_MORE_WARNINGS="$CFLAGS_MORE_WARNINGS $option"], [])
```

Also, `CFLAGS_MORE_WARNINGS` is now used, but nowhere explicitly initialized.

lkundrak:
> Isn't unconditionally resetting CFLAGS also undesired w.r.t. 0999ebd ?

Yes. The commit originally came from nm-a where it wasn't the case.

> If all you want to do, is ensure that _NM_COMPILER_FLAG is called with -Wall, why don't you:

That looks good. Modified it that way.

thom311:
lgtm now

PR #176, opened 2018-07-31
th/connection-cardinality-rh1555012

thom311:
Early version of connection-cardinality feature ([rh#1555012](https://bugzilla.redhat.com/show_bug.cgi?id=1555012)), also known as wildcard profiles.

The entire feature consists of 2 parts:

1) allowing profiles to be active on multiple devices at the same time (connection-cardinality)
2) extended matching features. Currently, we match profiles on devices only based on `connection.type`, `connection.interface-name`, `ethernet.mac-address` and `wireless.mac-address`. More matching features should be added.

Please review this pull request. It only addresses 1). And it is not yet thoroughly tested, quite possibly things still needs fixing. Also, more testing will be required. Nonetheless, early review is desired to speed up the process, because this feature is requested to be finished very soon.

PR #177, opened 2018-08-01
th/connection-multi-connect-rh1555012

thom311:
(obsoletes #176)

Early version of connection-cardinality feature (rh#1555012), also known as wildcard profiles.

The entire feature consists of 2 parts:

   1) allowing profiles to be active on multiple devices at the same time (connection-cardinality)
   2) extended matching features. Currently, we match profiles on devices only based on connection.type, connection.interface-name, ethernet.mac-address and wireless.mac-address. More matching features should be added.

Please review this pull request. It only addresses 1). And it is not yet thoroughly tested, quite possibly things still needs fixing. Also, more testing will be required. Nonetheless, early review is desired to speed up the process, because this feature is requested to be finished very soon.


libnm-core/nm-setting-connection.c
```diff
@@ -1764,6 +1801,22 @@ nm_setting_connection_class_init (NMSettingConnectionClass *setting_class)
...
 	                       G_PARAM_STATIC_STRINGS));
 
+	/**
+	 * NMSettingConnection:multi-connect:
+	 *
+	 * Specifies whether the profile can be active multiple times at a particulare
+	 * moment. The values are of type #NMConnectionMultiConnect.
+	 *
+	 * Since: 1.14
+	 */
```
bengal: Please also document the ifcfg-rh variable.

NEWS
```diff
@@ -9,6 +9,8 @@ with the later release.
 USE AT YOUR OWN RISK.  NOT RECOMMENDED FOR PRODUCTION USE!
 
 * Added support for IEEE 802.15.4 and 6LowPAN devices.
+* Support activating profile multiple times via connection.connect-multiple
```
bengal: The property name is connection.multi-connect.

src/nm-manager.c
```diff
@@ -4273,7 +4275,12 @@ _internal_activate_device (NMManager *self, NMActiveConnection *active, GError *
 	/* Check slaves for master connection and possibly activate them */
 	autoconnect_slaves (self, connection, device, nm_active_connection_get_subject (active));
 
-	{
+	multi_connect = _nm_connection_get_multi_connect (NM_CONNECTION (connection));
+	if (   multi_connect == NM_CONNECTION_MULTI_CONNECT_MULTIPLE
+	    || (   multi_connect == NM_CONNECTION_MULTI_CONNECT_MANUAL_MULTIPLE
+	        && nm_active_connection_get_activation_reason (active) != NM_ACTIVATION_REASON_AUTOCONNECT)) {
```
bengal: Perhaps it would be better to explicitly check that the reason is USER_REQUEST here?

Should MANUAL_MULTIPLE allow to activate, for example, connections with reason EXTERNAL, ASSUME and AUTOCONNECT_SLAVES?

thom311:
addressed comments and repushed.

PR #178, opened 2018-08-01
bg/update-dev-state-file-rh1593282

bengal:
Update the device state file every time the device is connected,
disconnected, or becomes unmanaged.  In this way, NM becomes more
robust against crashes or forced terminations because it can resume
the previous device state seamlessly.

https://bugzilla.redhat.com/show_bug.cgi?id=1593282


thom311:
lgtm

bengal:
Merged to master.

PR #182, opened 2018-08-12
th/wifi-gbytes-ssid

thom311:
cleanup handling the SSID as a GBytes, instead of GByteArray.

For one, GBytes is immutable and ref-countable, which is nicer API.
Also, at various places, we treat the SSID as GBytes already. Differing types, requires lots of annoying conversions.

### DIFF
--- a/libnm-core/tests/test-general.c
+++ b/libnm-core/tests/test-general.c
@@ -6559,74 +6559,161 @@ test_nm_utils_enum (void)
 /*****************************************************************************/
 
 static void
-do_test_utils_str_utf8safe (const char *str, const char *expected, NMUtilsStrUtf8SafeFlags flags)
+_do_test_utils_str_utf8safe_unescape (const char *str, const char *expected, gsize expected_len)
 {
-	const char *str_safe, *s;
-	gs_free char *str2 = NULL;
-	gs_free char *str3 = NULL;
+	gsize l;
+	const char *s;
+	gs_free gpointer buf_free_1 = NULL;
+	gs_free char *str_free_1 = NULL;
 
-	str_safe = nm_utils_str_utf8safe_escape (str, flags, &str2);
+	s = nm_utils_buf_utf8safe_unescape (str, &l, &buf_free_1);
+	g_assert_cmpint (expected_len, ==, l);
+	g_assert_cmpstr (s, ==, expected);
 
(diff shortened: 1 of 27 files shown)


shared/nm-utils/nm-shared-utils.c
```diff
@@ -117,6 +117,37 @@ nm_utils_strbuf_append (char **buf, gsize *len, const char *format, ...)
...
+ * @bytes: (allow-none): a #GBytes array to compare. Note that
+ *   %NULL is treated like an #GBytes array of length zero.
+ * @arr: the data pointer with @arr_len bytes
+ * @arr_len: the length of the data pointer
+ *
+ * Returns: %TRUE if @bytes contains the same data as @arr. As a
+ *   special case, a %NULL @bytes is treated like an empty array.
+ */
+gboolean
+nm_utils_gbytes_equals (GBytes *bytes,
```
lkundrak: I can't imagine this being used in a way that isn't entirely confusing and a major obstacle to readability.
I strongly suggest you just drop it.

shared/nm-utils/nm-shared-utils.c
```diff
@@ -146,6 +146,21 @@ nm_utils_gbytes_equals (GBytes *bytes,
...
 }
 
+GVariant *
+nm_utils_gbytes_to_variant_ay (GBytes *bytes)
+{
+	const guint8 *p;
+	gsize l;
+
+	if (!bytes) {
+		/* for convenience, accept NULL to return an empty variant */
```
lkundrak: Indicate this quirk in the function name? (e.g. how g_strcmp0() does.)

libnm-core/tests/test-general.c
```diff
@@ -6559,74 +6559,161 @@ test_nm_utils_enum (void)
 /*****************************************************************************/
```
lkundrak: s/WI-Fi/Wi-Fi/ s/commoly/commonly/ in the commit message

shared/nm-utils/nm-shared-utils.c
bengal: Since the function is used only in one place, I tend to agree that it would be better to open-code it.

The rest LGTM.

thom311:
merged

PR #183, opened 2018-08-17
IWD KnownNetwork changes

balrog-kun:
These commits mostly update the use of the IWD DBus API with the changes seen in this commit: https://git.kernel.org/pub/scm/network/wireless/iwd.git/commit/doc/knownnetworks-api.txt?id=7d3656858ffd105ee5951f934be6ca123956d1b3

The last three commits are unrelated and likely need more discussion, they're a different approach to make EAP networks usable with the IWD backend without manually creating NM profiles for each network we want to connect to (which has to have been configured on the IWD side already), the first approach was described here: https://mail.gnome.org/archives/networkmanager-list/2018-June/msg00023.html

### DIFF
--- a/src/devices/wifi/nm-iwd-manager.c
+++ b/src/devices/wifi/nm-iwd-manager.c
@@ -34,8 +34,13 @@
 /*****************************************************************************/
 
 typedef struct {
-	char *name;
+	const char *name;
 	NMIwdNetworkSecurity security;
+	char buf[0];
+} KnownNetworkId;
+
+typedef struct {
+	GDBusProxy *known_network;
 } KnownNetworkData;
 
 typedef struct {
@@ -45,7 +50,7 @@ typedef struct {
 	GDBusObjectManager *object_manager;
 	guint agent_id;
 	char *agent_path;
-	GSList *known_networks;
+	GHashTable *known_networks;
 } NMIwdManagerPrivate;
 
 struct _NMIwdManager {
@@ -273,6 +278,30 @@ register_agent (NMIwdManager *self)
 
 /*****************************************************************************/
 
+static guint
+known_network_id_hash (KnownNetworkId *id)
+{
+	return g_str_hash (id->name) + id->security;
+}
+
+static gboolean
+known_network_id_equal (KnownNetworkId *a, KnownNetworkId *b)
+{
+	return g_str_equal (a->name, b->name) && a->security == b->security;
+}
+
+static void
+known_network_data_free (KnownNetworkData *network)
+{
+	if (!network)
+		return;
+
+	g_object_unref (network->known_network);
+	g_slice_free (KnownNetworkData, network);
+}
+
+/*****************************************************************************/
+
 static void
 set_device_dbus_object (NMIwdManager *self, GDBusInterface *interface,
                         GDBusObject *object)
@@ -367,13 +396,6 @@ object_added (NMIwdManager *self, GDBusObject *object)
 	}
 
 	g_list_free_full (interfaces, g_object_unref);
-}
-
-static void
-known_network_free (KnownNetworkData *network)
-{
-	g_free (network->name);
-	g_free (network);
 }
 
 static void prepare_object_manager (NMIwdManager *self);
@@ -515,16 +537,9 @@ nm_iwd_manager_is_known_network (NMIwdManager *self, const char *name,
                                  NMIwdNetworkSecurity security)
 {
 	NMIwdManagerPrivate *priv = NM_IWD_MANAGER_GET_PRIVATE (self);
-	const GSList *iter;
+	KnownNetworkId kn_id = { name, security };
 
-	for (iter = priv->known_networks; iter; iter = g_slist_next (iter)) {
-		const KnownNetworkData *network = iter->data;
-
-		if (!strcmp (network->name, name) && network->security == security)
-			return true;
-	}
-
-	return false;
+	return g_hash_table_contains (priv->known_networks, &kn_id);
 }
 
 /*****************************************************************************/
@@ -542,6 +557,10 @@ nm_iwd_manager_init (NMIwdManager *self)
 	                  G_CALLBACK (device_added), self);
 
 	priv->cancellable = g_cancellable_new ();
+
+	priv->known_networks = g_hash_table_new_full ((GHashFunc) known_network_id_hash,
+	                                              (GEqualFunc) known_network_id_equal, g_free,
+	                                              (GDestroyNotify) known_network_data_free);
 
 	prepare_object_manager (self);
 }
@@ -574,7 +593,8 @@ dispose (GObject *object)
 
 	nm_clear_g_cancellable (&priv->cancellable);
 
-	g_slist_free_full (priv->known_networks, (GDestroyNotify) known_network_free);
+	if (priv->known_networks)
+		g_hash_table_unref (priv->known_networks);
 	priv->known_networks = NULL;
 
 	if (priv->manager) {
--- a/src/devices/wifi/nm-iwd-manager.c
+++ b/src/devices/wifi/nm-iwd-manager.c
@@ -83,6 +83,32 @@ G_DEFINE_TYPE (NMIwdManager, nm_iwd_manager, G_TYPE_OBJECT)
 
 /*****************************************************************************/
 
+static const char *
+get_variant_string_or_null (GVariant *v)
+{
+	if (!v)
+		return NULL;
+
+	if (   !g_variant_is_of_type (v, G_VARIANT_TYPE_STRING)
+	    && !g_variant_is_of_type (v, G_VARIANT_TYPE_OBJECT_PATH))
+		return NULL;
+
+	return g_variant_get_string (v, NULL);
+}
+
+static const char *
+get_property_string_or_null (GDBusProxy *proxy, const char *property)
+{
+	gs_unref_variant GVariant *value = NULL;
+
+	if (!proxy || !property)
+		return NULL;
+
+	value = g_dbus_proxy_get_cached_property (proxy, property);
+
+	return get_variant_string_or_null (value);
+}
+
 static void
 agent_dbus_method_cb (GDBusConnection *connection,
                       const char *sender, const char *object_path,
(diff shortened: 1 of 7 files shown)


vhumpa:

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -307,24 +332,105 @@ interface_added (GDBusObjectManager *object_manager, GDBusObject *object,
...
+	if (!priv->running)
+		return;
 
-	set_device_dbus_object (self, interface, object);
+	g_return_if_fail (G_IS_DBUS_PROXY (interface));
+
+	proxy = G_DBUS_PROXY (interface);
+
+	if (!strcmp (g_dbus_proxy_get_interface_name (proxy),
+	             NM_IWD_DEVICE_INTERFACE)) {
```
thom311: Let's cache the interface-name in a variable, and use `nm_streq*()` instead of `strcmp()`?
```C
    interface = g_dbus_proxy_get_interface_name (proxy);
    if (nm_streq0 (interface, NM_IWD_DEVICE_INTERFACE)) {
        ...
```

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -307,24 +332,105 @@ interface_added (GDBusObjectManager *object_manager, GDBusObject *object,
...
+		else if (!strcmp (type_str, "psk"))
+			security = NM_IWD_NETWORK_SECURITY_PSK;
+		else if (!strcmp (type_str, "8021x"))
+			security = NM_IWD_NETWORK_SECURITY_8021X;
+		else
+			return;
+
+		g_variant_unref (value);
+		value = g_dbus_proxy_get_cached_property (proxy, "Name");
+		if (!value)
```
thom311: everywhere: need to check compatible `value` type (string in this case).

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -307,24 +332,105 @@ interface_added (GDBusObjectManager *object_manager, GDBusObject *object,
...
+			return;
+
+		id = known_network_id_new (g_variant_get_string (value, NULL), security);
```
thom311: why is the `security` part of the ID? Shouldn't the ID be the D-Bus path of the known-network, instead of the `(name, security)` pair?

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -307,24 +332,105 @@ interface_added (GDBusObjectManager *object_manager, GDBusObject *object,
...
+
+		data = g_slice_new (KnownNetworkData);
+		data->known_network = (GDBusProxy *) g_object_ref (proxy);
```
thom311: let's not do unnecessary casts.

src/devices/wifi/nm-wifi-utils.h
```diff
@@ -26,6 +26,7 @@
 #include "nm-setting-wireless.h"
 #include "nm-setting-wireless-security.h"
 #include "nm-setting-8021x.h"
+#include "nm-iwd-manager.h"
```
thom311: I think `nm-wifi-utils.h` must not include iwd related headers. If you need it for the enum, then move the definition of the enum as well.

src/devices/wifi/nm-wifi-utils.c
```diff
@@ -814,3 +814,26 @@ nm_wifi_utils_is_manf_default_ssid (const GByteArray *ssid)
...
+	key_mgmt = nm_setting_wireless_security_get_key_mgmt (s_wireless_sec);
+	nm_assert (key_mgmt);
+
+	if (!strcmp (key_mgmt, "none") || !strcmp (key_mgmt, "ieee8021x"))
+		return NM_IWD_NETWORK_SECURITY_WEP;
+
+	if (!strcmp (key_mgmt, "wpa-psk"))
+		return NM_IWD_NETWORK_SECURITY_PSK;
+
+	nm_assert (!strcmp (key_mgmt, "wpa-eap"));
```
thom311: this assert seems much too strict. Already today, `key_mgmt` can be `"wpa-none"`

While at it, `nm_streq()` and `NM_IN_STRSET()` would be better than `strcmp()`.

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -307,24 +332,105 @@ interface_added (GDBusObjectManager *object_manager, GDBusObject *object,
...
+		set_device_dbus_object (self, proxy, object);
+		return;
+	}
+
+	if (!strcmp (g_dbus_proxy_get_interface_name (proxy),
+	             NM_IWD_KNOWN_NETWORK_INTERFACE)) {
+		KnownNetworkId *id;
+		KnownNetworkData *data;
+		NMIwdNetworkSecurity security;
+		gs_unref_variant GVariant *value;
```
thom311: variables with `__attribute__((cleanup()))` must be initialized at definition. In this case ` = NULL`.

Otherwise, there is a compiler warning `-Werror=maybe-uninitialized`, in particular in combination with `-fexceptions`.

src/supplicant/nm-supplicant-config.c
```diff
@@ -1057,6 +1057,9 @@ nm_supplicant_config_add_setting_8021x (NMSupplicantConfig *self,
 			fast = TRUE;
 			priv->fast_required = TRUE;
 		}
+
+		if (method && num_eap == 1 && strcasecmp (method, "external") == 0)
```
thom311: using `strcasecmp()` here is most certainly wrong, because `NMSetting8021x::verify()` ensures that the method values matches literally (case-sensitive). Existing code should be fixed, but at least new code should not get it wrong too.

In fact, using `strcasecmp()` in NetworkManager daemon is always wrong, because the function is locale dependent. If case-insensitive comparison is needed, `g_ascii_strcasecmp()` would be better. But that is not the case here.

Teh code here rejects "external" only, if it is present alone in the list of EAP methods (good). But the lines above with `ADD_STRING_LIST_VAL()` should filter out "external". As we anyway iterate over `num_eap` here, it seems simpler to drop `ADD_STRING_LIST_VAL()` macro and just create the supplicant option here (thereby skipping over "external").

Also, you cannot `return FALSE`, without also setting `@error`.

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -114,27 +114,25 @@ agent_dbus_method_cb (GDBusConnection *connection,
 	                                               network_path,
 	                                               NM_IWD_NETWORK_INTERFACE);
 	value = g_dbus_proxy_get_cached_property (G_DBUS_PROXY (network), "Device");
-	device_path = g_variant_get_string (value, NULL);
-
-	if (!device_path) {
+	if (!value) {
```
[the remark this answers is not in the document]
balrog-kun: Yes, good point I guess.  I'm not sure how much of that is already checked by the dbus daemon, pretty sure things like PropertiesChanged events don't have their dict parameters checked against the interface introspection string.

src/devices/wifi/nm-iwd-manager.c
balrog-kun: Ok, makes sense.

src/devices/wifi/nm-iwd-manager.c
thom311: It depends on the compiler and compiler options. Only because your current compilation doesn't report the warning, doesn't mean it's fine. Also, only because you initialize the variable before returning (and you know that there are no C++ exceptions that cause to leave the scope prematurely), is not enough to make the compiler happy.

thom311: What I mean is: yes, there is no problem in practice, but the compiler warning is still emitted. You cannot argue with that, all compiler warnings must be avoided (worked-around) even if they are false-positives.

src/devices/wifi/nm-iwd-manager.c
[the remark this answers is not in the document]
thom311: AFAIS, nowhere in NetworkManager is it declared that "Device" property on "net.connman.iwd.Network" interface (on the network object) is of type "s". Hence, gdbus cannot pre-validate the type (*). When you fetch the GVariant as cached-proxy, the type depends entirely on what iwd currently exports on it's D-Bus interface. Sure, iwd's D-Bus API is defined and hopefully stable. But NetworkManager must not trust that, but check that the type is as expected. Or at least, behave gracefully. Calling `g_variant_get_string()` on a GVariant which is not a string type is a bug. That's not gracefully skipping the value.

(*) for example, in gdbus you can pass introspection information to `g_dbus_proxy_new()`. In that case, indeed the types will be checked by the library. But you didn't specify the iwd D-Bus API in NetworkManager (nor does it seem a good idea to do so).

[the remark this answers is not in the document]
balrog-kun: Yep, none of that is validated by NM or its libraries, what I meant is that the D-Bus daemon would be able to validate the types in most signals and properties.  But then again D-Bus is an external component.
Indeed IWD's API is not considered stable in the 0.x versions and even in 1.0 some interfaces will be marked experimental and will not maintain backwards compatibility.

src/devices/wifi/nm-iwd-manager.c
balrog-kun: Well, I simplified this whole variable away from the code now.  I didn't consider exceptions because this is compiled as C code but I guess a warning could be justified in that case.

src/devices/wifi/nm-iwd-manager.c
thom311: Interesting. So how does IWD cope with multiple SSIDs being in the vicinity? Will it just pick one randomly/deterministically, and hide all others?

balrog-kun: It will have to assume they're multiple BSSes in one ESS / WLAN and treat them as one network.  The spec says SSID == ESSID, we don't currently try to detect conflicts and I'm not sure how much could actually be done there.

Note IWD only exposes whole networks to NM over dbus so we'll only select the best BSS when we're connecting.

thom311:
Hi Andrew,

I merged the branch to master as 3ec6a4479cda8482e6aef00c4f9d85ded4d73f99 (after ACK from @bengal  and @fgiudici).

While doing so, I rebased the branch onto master. There were trivial conflicts that I resolved thereby. Apart from that, the commits that bear your name are basically unchanged.

On top of that, I added a few more patches. Please have a look, and speak up if you dislike them.


Note, that in a week or two, we will release NetworkManager 1.14.0. I didn't actually test the branch. If you find the time, it would be great if you could verify that NM+iwd works well for the release. Of course, we can always fix stuff later as well...

Thanks!!

balrog-kun:
Thanks for the work on bringing this up to nm standards.  I have missed some of the things you commented on here even though I agreed with them, sorry.

Also thanks for the heads up about 1.14.0, I'll still send one or two fixes to the iwd backend this week.

PR #184, opened 2018-08-23
th/settings-delegate-connection

thom311:
Refactor `NMSettingsConnection` to not implement `NMConnection` interface, but instead delegate to a `NM(Simple)Connection` member.

The point is, to better separate what `NMSettingsConnection` is supposed to do and not also let it implement the `NMConnection` interface.

### DIFF
--- a/src/NetworkManagerUtils.c
+++ b/src/NetworkManagerUtils.c
@@ -805,8 +805,10 @@ nm_utils_match_connection (NMConnection *const*connections,
 		return NULL;
 
 	for (; *connections; connections++) {
-		NMConnection *candidate = NM_CONNECTION (*connections);
+		NMConnection *candidate = *connections;
 		GHashTable *diffs = NULL;
+
+		nm_assert (NM_IS_CONNECTION (candidate));
 
 		if (match_filter_func) {
 			if (!match_filter_func (candidate, match_filter_data))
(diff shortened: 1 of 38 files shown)


thom311:
btw, the larger picture for this change:

I want to solve [bgo#772414](https://bugzilla.gnome.org/show_bug.cgi?id=772414), which is about /usr and /run directories for keyfile.

For that, I want to first solve [bgo#744711](https://bugzilla.gnome.org/show_bug.cgi?id=744711), which is about decoupling the storage of profiles, from NMSettingsConnection.

1) `NMSettingsConnection` shall no longer implement `NMConnection`.

2) `NMSettingsConnection` should not be subclassed to `NMSKeyfileConnection`, but delegate to some storage/backend instances how to persist a profile. This should make it possible (at least in principle) to migrate profiles from one setting plugin to another. Also, thereby pure in-memory profiles could be handled by `NMSettingsConnection` itself, which just needs to remember whether the storage/backend is up-to-date.

I'd like to do these two steps first, to simplify what `NMSettingsConnection` does. Before tackling keyfile directories.

bengal:
The branch looks good to me.

vbenes:
Thomas, we have a crash here:
https://faf.lab.eng.brq.redhat.com/faf//reports/8026/

thom311:
merged

PR #186, opened 2018-08-26
[th/systemd] reimport sources from systemd upstream

thom311:
Nothing noteworthy here. Just our regular import.


bengal:
LGTM

thom311:
merged to master

PR #188, opened 2018-08-27
th/gbytes-equal-mem

thom311:
Resurrect `nm_utils_gbytes_equal_mem()` from #182. I think the concerns raised there are addressed by this new version.

I repost this for inclusion, because I once again imported systemd code and marvelled at how beautiful their code is. And they really add *a lot* of helper function on top of what glibc provides. For example, trivialities like

```C
int getenv_bool(const char *p) {
        const char *e;

        e = getenv(p);
        if (!e) 
                return -ENXIO;

        return parse_boolean(e);
}
```
Whether this is good or bad style, is ultimately a matter of opinion. I think crafting and curating a neat set of utility functions is indeed good style, and we should get better at that. We should find common functionality, and move it to functions which get re-used.

C is first and foremost a procedural programming language, we shouldn't shy away from adding procedures.

Compare also [c-util](https://github.com/c-util) where Tom and David go a step further and build stand-alone utility libraries.

---

> The only serious thing here is the nm_utils_gbytes_equals() which I find to be bad style.
>
> Encountering a routine whose operation you can't guess from its name while reading the source essentially leaves the reader two options: either make assumptions about what it does and potentially overlook problems (what types it accepts? does it return bool or 1/0/-1? how does it treat either argument being NULL?), or interrupt their reading while descending to the utility code, wasting time. Both bad. Open-coding it would result in clearer code.

It's now named `nm_utils_gbytes_equal_mem()`, is that better? It our convention that functions named `*equal*()` or `*equals*()` return a boolean (not -1/0/1 which is for ordering and commonly named `*cmp*()`, `nm_setting_compare()` and `nm_setting_diff()` are notable exception to this).

For all functions the rule of thumb is: do not pass %NULL unless it's explicitly allowed. That is true for `strcmp()`, `g_object_unref()`, etc. A lot of functions accept and do something sensible in face of %NULL arguments. That is for pure convenience, for example `free(NULL)` does nothing. Interestingly, `memcpy (dst, NULL, 0)` is not safe by C standard. A reviewer who cares can trivially figure out the behaviour on %NULL, and to a reviewer who doesn't care: the usage of `nm_utils_gbytes_equal_mem()` is sensible and defensive in face of %NULL. If you don't care: it just does the right thing.

---

> Since the function is used only in one place, I tend to agree that it would be better to open-code it.

Now it's used at 2 places (+ several places in tests, do they count?). I think GBytes is one of the better data types in glib. We should leverage it more because it's ref-counted and immutable. We should have useful helper functions that suit our needs.

### DIFF
--- a/libnm-core/nm-core-internal.h
+++ b/libnm-core/nm-core-internal.h
@@ -238,12 +238,6 @@ _nm_auto_ip_route_unref (NMIPRoute **v)
 }
 #define nm_auto_ip_route_unref nm_auto (_nm_auto_ip_route_unref)
 
-GPtrArray *_nm_utils_copy_slist_to_array (const GSList *list,
-                                          NMUtilsCopyFunc copy_func,
-                                          GDestroyNotify unref_func);
-GSList    *_nm_utils_copy_array_to_slist (const GPtrArray *array,
-                                          NMUtilsCopyFunc copy_func);
-
 GPtrArray *_nm_utils_copy_array (const GPtrArray *array,
                                  NMUtilsCopyFunc copy_func,
                                  GDestroyNotify free_func);
--- a/libnm-core/nm-utils.c
+++ b/libnm-core/nm-utils.c
@@ -623,39 +623,6 @@ _nm_utils_copy_strdict (GHashTable *strdict)
 }
 
 GPtrArray *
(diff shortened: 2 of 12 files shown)


bengal:
Nice, LGTM.

thom311:
merged

PR #189, opened 2018-08-27
libnm-core: support private keys encrypted with AES-{192,256}-CBC

bengal:
If I try to use the `sample/sample-keys/client-pass.key` from OpenVPN source tree, NM complains that the key is invalid, because it is encrypted with AES-256-CBC. Add support for the missing encryption algorithms.


libnm-core/crypto.c
```diff
@@ -158,7 +158,13 @@ parse_old_openssl_key_file (const guint8 *data,
 				goto parse_error;
 			}
 		} else if (!strncmp (p, DEK_INFO_TAG, strlen (DEK_INFO_TAG))) {
+			const char *const known_ciphers[] = { CIPHER_DES_EDE3_CBC,
```
thom311: static?

libnm-core/crypto.c
```diff
@@ -187,13 +193,13 @@ parse_old_openssl_key_file (const guint8 *data,
...
 			/* Get the private key cipher */
-			if (!strcasecmp (p, "DES-EDE3-CBC")) {
-				cipher = g_strdup (p);
-			} else if (!strcasecmp (p, "DES-CBC")) {
-				cipher = g_strdup (p);
-			} else if (!strcasecmp (p, "AES-128-CBC")) {
-				cipher = g_strdup (p);
-			} else {
+			for (i = 0; i < G_N_ELEMENTS (known_ciphers); i++) {
+				if (!strcasecmp (p, known_ciphers[i])) {
```
thom311: `strcasecmp()` is locale dependent. I doubt, we should consider the locale when comparing strings. There is `g_ascii_strcasecmp()` instead.

libnm-core/crypto_nss.c
```diff
@@ -269,7 +271,10 @@ crypto_encrypt (const char *cipher,
 
 	if (!strcmp (cipher, CIPHER_DES_EDE3_CBC))
 		cipher_mech = CKM_DES3_CBC_PAD;
-	else if (!strcmp (cipher, CIPHER_AES_CBC))
+	else if (NM_IN_STRSET(cipher,
```
thom311: space before '('

PR #192, opened 2018-08-31
bg/llmnr - Allow controlling DNS LLMNR setting

bengal:
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/13


src/devices/nm-device.c
```diff
@@ -2181,6 +2181,34 @@ _get_mdns (NMDevice *self)
...
+	if (connection)
+		llmnr = nm_setting_connection_get_llmnr (nm_connection_get_setting_connection (connection));
+
+	if (llmnr == NM_SETTING_CONNECTION_LLMNR_DEFAULT) {
+		gs_free char *value = NULL;
+
+		value = nm_config_data_get_connection_default (NM_CONFIG_GET_DATA,
+		                                               "connection.llmnr",
+		                                               self);
+		llmnr = _nm_utils_ascii_str_to_int64 (value,
```
thom311: maybe it's time to add a helper `nm_config_data_get_connection_default_int64()`? We are doing this repeatedly.

libnm-core/nm-setting-connection.c
```diff
@@ -1098,12 +1117,23 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
...
 		             NM_CONNECTION_ERROR_INVALID_PROPERTY,
-		             _("mdns value %d is not valid"), priv->mdns);
+		             _("value %d is not valid"), priv->mdns);
 		g_prefix_error (error, "%s.%s: ", NM_SETTING_CONNECTION_SETTING_NAME,
 		                NM_SETTING_CONNECTION_MDNS);
 		return FALSE;
 	}
 
+	if (   priv->llmnr < NM_SETTING_CONNECTION_LLMNR_DEFAULT
+	    || priv->llmnr > NM_SETTING_CONNECTION_LLMNR_YES) {
```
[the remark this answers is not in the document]
bengal: Ok.

src/devices/nm-device.c
bengal: How about now?

src/nm-ip4-config.c
```diff
@@ -2933,13 +2933,21 @@ nm_ip4_config_hash (const NMIP4Config *self, GChecksum *sum, gboolean dns_only)
...
 
 	val = nm_ip4_config_llmnr_get (self);
-	g_checksum_update (sum, (const guint8 *) &val, sizeof (val));
-
-	val = nm_ip4_config_get_dns_priority (self);
-	g_checksum_update (sum, (const guint8 *) &val, sizeof (val));
+	if (val != NM_SETTING_CONNECTION_LLMNR_DEFAULT)
+		g_checksum_update (sum, (const guint8 *) &val, sizeof (val));
+
+	/* FIXME: the DNS priority should be considered as well,
```
thom311: Where is this assumption made? Can you at the same time also add a FIXME comment at the place where this assumption is made?

Recently, I started grouping FIXME comments like "FIXME(shutdown)" and "FIXME(copy-on-write-connection)". Can you group these two FIXME statements together?


Btw, the approach of computing the MD5 sum of NMIPXConfig instance for comparing them is significantly flawed. We should implement proper hash/equals/cmp functions, like we commonly do. That warrants a FIXME comment in extra big letters.


Btw, in case you didn't know, instead of "fixup! " in the commit message, you can use "squash!". Also, `git commit --squash=$COMMIT`.

bengal: Ok, I've added a FIXME and improved the existing one.

PR #194, opened 2018-09-07
iwd-fixes

balrog-kun:
Two fixes and an update for IWD DBus changes

### DIFF
--- a/src/devices/wifi/nm-device-iwd.c
+++ b/src/devices/wifi/nm-device-iwd.c
@@ -67,7 +67,8 @@ static guint signals[LAST_SIGNAL] = { 0 };
 
 typedef struct {
 	GDBusObject *   dbus_obj;
-	GDBusProxy *    dbus_proxy;
+	GDBusProxy *    dbus_device_proxy;
+	GDBusProxy *    dbus_station_proxy;
 	CList           aps_lst_head;
 	NMWifiAP *      current_ap;
 	GCancellable *  cancellable;
@@ -246,7 +247,7 @@ get_ordered_networks_cb (GObject *source, GAsyncResult *res, gpointer user_data)
 	                                      G_VARIANT_TYPE ("(a(osns))"),
 	                                      &error);
 	if (!variant) {
-		_LOGE (LOGD_WIFI, "Device.GetOrderedNetworks failed: %s",
+		_LOGE (LOGD_WIFI, "Station.GetOrderedNetworks failed: %s",
 		       error->message);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


vhumpa:

thom311:
as the iwd plugin switches to using newer iwd D-Bus API (loosing compatibility with older versions), we should somehow state which API is suitable. Quite possibly log a message in `nm_iwd_manager_init()` with
```
  _LOGI ("requiring iwd >= 0.8");
```
(or whatever appropriate)

thom311:
>    wifi/iwd: use the new 'Station' DBus interface
>
>    The following commit between IWD 0.7 and 0.8 splits the previous Device

It seems the referenced commit was between 0.6 and 0.7.

If this gets merged, is 0.6 API still working, what about 0.7?

Is the current assumption, that NetworkManager's master only works against current iwd master branch? It seems, when we do 1.14.0 release, then the targeted iwd API should correspond to at least some exact version (like 0.7), even if this version is still unstable. Otherwise, how would somebody know, which iwd version to use in combination with NM 1.14.0?

thom311:
In the linked wiki, I see that 921db9132ec64c2d is compatible with 0.5, 0.6, and 0.7.

I think such compatibility, is generally desirable. Yes, the iwd API is not stable, and support for such versions might be dropped later, but if NM is only working with a few particular iwd versions, it becomes complicated for the user to get the pieces together. Would it be too complicated to support both 0.7 and 0.8? From the patch it seems, such compatibility, could be easily doable.

---

btw, in the wiki:
> NM configuration syntax allows for other possibilties, see NM documentation in nm-settings(5). 

This should reverence `man 5 NetworkManager.conf`

balrog-kun:
> Would it be too complicated to support both 0.7 and 0.8?

Yep, it should actually be easy, let me try doing that.  (in fact 0.8 is temporarily backwards compatible but having few IWD clients so far we prefer them to use the latest API so it can serve as a code example)

> This should reverence `man 5 NetworkManager.conf`

Thank you, will fix.

thom311:
Last remaining patch got merged as well. Thanks Andrew!!

balrog-kun:
Hi Thomas, so one final API change was made yesterday to one of the methods on the new interface which breaks compatibilty.  It is supposed to be the last such API update for the next few versions.  This doesn't break 0.7 but means NM will not work with the upcoming IWD 0.8 so it would be great to still get this one change into a 1.14.0 release candidate.  I will create a pull request and let you decide when/whether to pull it.

(the docs commit I linked did actually reflect this coming change but I had forgotten about it, and couldn't have tested it until now anyway)

PR #195, opened 2018-09-07
bg/n-acd-update: update n-acd code from upstream

bengal:
Update the n-acd code to the upstream version, after the release of version 1 with stable API and (optional) support for eBPF filtering.


configure.ac
```diff
@@ -518,6 +518,15 @@ case $with_suspend_resume in
...
 esac
 
+# eBPF support
+AC_ARG_WITH(ebpf,
+            AS_HELP_STRING([--with-ebpf=yes|no], [Build with eBPF support (default: yes)]),
+            [], [with_ebpf=yes])
+if test "$with_ebpf" != "yes" -a "$with_ebpf" != "no"; then
+	AC_MSG_ERROR(--with-ebpf must be one of [yes, no])
+fi
+AM_CONDITIONAL(WITH_EBPF, test "${with_ebpf}" = "yes")
```
thom311: shouldn't we autodetect support, and only provide the --with-ebpf option if somebody needs to change the autodetection?

src/devices/nm-acd-manager.c
```diff
@@ -472,7 +496,15 @@ dispose (GObject *object)
 	NMAcdManager *self = NM_ACD_MANAGER (object);
 	NMAcdManagerPrivate *priv = NM_ACD_MANAGER_GET_PRIVATE (self);
 
+	g_clear_pointer (&priv->hwaddr, g_free);
```
thom311: nm_clear_g_free()?

src/devices/nm-acd-manager.c
```diff
@@ -472,7 +496,15 @@ dispose (GObject *object)
...
 	g_clear_pointer (&priv->addresses, g_hash_table_destroy);
+	g_clear_pointer (&priv->channel, g_io_channel_unref);
+	nm_clear_g_source (&priv->event_id);
+
+	if (priv->acd) {
+		n_acd_unref (priv->acd);
+		priv->acd = NULL;
+	}
```
thom311: nm_clear_pointer(&priv->acd, n_acd_unref)?

src/devices/nm-acd-manager.c
```diff
@@ -450,18 +472,20 @@ nm_acd_manager_init (NMAcdManager *self)
...
 
+	g_return_val_if_fail (ifindex > 0, NULL);
 	g_return_val_if_fail (hwaddr, NULL);
 	g_return_val_if_fail (hwaddr_len == ETH_ALEN, NULL);
 
 	self = g_object_new (NM_TYPE_ACD_MANAGER, NULL);
 	priv = NM_ACD_MANAGER_GET_PRIVATE (self);
 	priv->ifindex = ifindex;
-	memcpy (priv->hwaddr, hwaddr, ETH_ALEN);
+	priv->hwaddr = g_memdup (hwaddr, hwaddr_len);
```
thom311: nm_memdup(). There is little difference in practice, but I think we should restrain ourself from API that is error prone.

src/devices/nm-acd-manager.c
```diff
@@ -54,10 +51,14 @@ static guint signals[LAST_SIGNAL] = { 0 };
 
 typedef struct {
 	int            ifindex;
-	guint8         hwaddr[ETH_ALEN];
+	guint8         *hwaddr;
+	size_t         hwaddr_len;
```
thom311: guint? Because at other places, the length is limited to guint type. Or, adjust the other places.

src/devices/nm-acd-manager.c
```diff
@@ -450,18 +472,20 @@ nm_acd_manager_init (NMAcdManager *self)
...
 NMAcdManager *
-nm_acd_manager_new (int ifindex, const guint8 *hwaddr, size_t hwaddr_len)
+nm_acd_manager_new (int ifindex, const guint8 *hwaddr, guint hwaddr_len)
 {
 	NMAcdManager *self;
 	NMAcdManagerPrivate *priv;
 
+	g_return_val_if_fail (ifindex > 0, NULL);
 	g_return_val_if_fail (hwaddr, NULL);
 	g_return_val_if_fail (hwaddr_len == ETH_ALEN, NULL);
```
thom311: if you assert that the hwaddr_len is ETH_ALEN, no `priv->hwaddr_len` is needed.

src/devices/nm-acd-manager.c
```diff
@@ -174,115 +168,130 @@ nm_acd_manager_add_address (NMAcdManager *self, in_addr_t address)
...
 {
 	NMAcdManagerPrivate *priv = NM_ACD_MANAGER_GET_PRIVATE (self);
-	NAcdConfig *config;
-	int r, fd;
+	NAcdProbeConfig *probe_config;
+	int r;
 
-	r = n_acd_new (&info->acd);
+	r = n_acd_probe_config_new (&probe_config);
 	if (r) {
```
thom311: at various places, isn't the convention to check for error with `if (r < 0)`?

src/devices/nm-acd-manager.c
```diff
@@ -117,21 +118,15 @@ static const char *
 _acd_error_to_string (int error)
```
thom311: the name does not seem optimal. Positive numbers are not really errors, are they?

src/devices/nm-acd-manager.c
bengal: ACD functions return 0 on success. On error they return a positive (N_ACD_E_PREEMPTED, N_ACD_E_INVALID_ARGUMENT) or negative (-errno) error code, so the code looks correct.

src/devices/nm-acd-manager.c
bengal: Yes, N_ACD_E_PREEMPTED and N_ACD_E_INVALID_ARGUMENT are errors.

_N_ACD_E_SUCCESS of course is not, but I think _acd_error_to_string() is fine as a name for a function that translates error codes to string.

configure.ac
bengal: ebpf is supported since kernel 3.19, which is more than 3 years old.
I think defaulting to yes is fine, while still allowing to disable it explicitly.

src/devices/nm-acd-manager.c
bengal: Okay.

src/devices/nm-acd-manager.c
bengal: Sure.

bengal:
Fixed the issues above and repushed.

src/devices/nm-acd-manager.c
thom311: ok. That was not obvious to me. Maybe `if (r != N_ACD_E_SUCCESS)`?

thom311: Or, fine as is with `if (r)`. If you prefer that.

src/devices/nm-acd-manager.c
```diff
@@ -117,21 +117,15 @@ static const char *
...
-	case N_ACD_E_STOPPED:
-		return "no more events (engine stopped)";
 	case N_ACD_E_PREEMPTED:
 		return "preempted";
 	case N_ACD_E_INVALID_ARGUMENT:
 		return "invalid argument";
-	case N_ACD_E_BUSY:
-		return "busy";
 	}
 	return NULL;
```
thom311: maybe `g_return_val_if_reached (NULL)`. In this case, n-acd is a static library that is compiled in. This isn't (really) an external component, and an unknown error code *is* a bug that we should assert against. And then drop the `acd_error_to_string()` macro.

bengal: Okay.

PR #196, opened 2018-09-08
docs/test: add check that gtk-doc contains patch to generate proper documentation

thom311:
In libnm, we prefer opaque typedefs. gtk-doc needs to be patched to properly
support that. Add a check for that.

See-also: https://gitlab.gnome.org/GNOME/gtk-doc/merge_requests/2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1217,7 +1217,7 @@ else
 	if test "$enable_gtk_doc" = "yes"; then
 		# large parts of the documentation require introspection/pygobject to extract
 		# the documentation out of the source files. You cannot enable gtk-doc without alone.
-		AC_MSG_ERROR(["--with-gtk-doc requires --enable-introspection"])
+		AC_MSG_ERROR(["--enable-gtk-doc requires --enable-introspection"])
 	fi
 	have_introspection=no
 fi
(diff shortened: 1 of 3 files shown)


bengal:
> Distributors may use the tarball, and don't regenerate the documentation (that's why we pack it after all).
> Or, if they generate the documentation, they really should generate valid documentation, and go through the effort of patching their gtk-doc. The test points out something they _should_ fix.
> Or, if they don't want, the can easily disable the test by setting an environment variable
> Or, if above is all too complicated, they are welcome to ignore the test failure.

I agree with this, it seems useful to detect during the build that we are generating wrong documentation.

fgiudici:
> > Distributors may use the tarball, and don't regenerate the documentation (that's why we pack it after all).
> > Or, if they generate the documentation, they really should generate valid documentation, and go through the effort of patching their gtk-doc. The test points out something they _should_ fix.
> > Or, if they don't want, the can easily disable the test by setting an environment variable
> > Or, if above is all too complicated, they are welcome to ignore the test failure.
> 
> I agree with this, it seems useful to detect during the build that we are generating wrong documentation.
+1

thom311:
I changed my opinion again, and the test no longer fails (unless setting `NMTST_CHECK_GTK_DOC=1`). Following Lubomir's suggestion.

Reason:
 - a user who generates docs for his own needs, he likely wouldn't care (or bother enough for patching gtk-doc)
 - when distro maintainers build the NM package, they cannot freely ensure that gtk-doc package is suitable. E.g. building the package for F28 in koji, would just fail the test. That was circumvented by disabling the test in contrib/rpm's spec file. That I disabled the test in our own spec file, already shows that this shouldn't be enabled by default

The only time when this test really needs to fail, is when building the source tarball for distribution.

I extended now the `contrib/fedora/rpm/build_clean.sh` script, to do that. Now, the correct way to generate a source-tarball for distribution is calling `./contrib/fedora/rpm/build_clean.sh -r`. Please use that.

thom311:
merged

PR #199, opened 2018-09-10
th/dhcp-error-reason

thom311:
Got an assertion failure when NM tried to activate a Wi-Fi profile on a wireguard link (that is now already prevented on master).

The wireguard link is Layer3 only, so no MAC address. When starting internal DHCP client, it would lead the an assertion in dhcp_network_bind_raw_socket(). Fail DHCP early when we have no MAC address.


thom311:
merged

PR #204, opened 2018-09-15
bg/meson-fixes

bengal:
Some other meson-related fixes.


thom311:
Generally lgtm.

However, follow-up issues:

---

```
$ meson build -D introspection=false 

meson.build:717:4: ERROR:  Problem encountered: Vala bindings require GObject introspection
```
This should automatically disable vapi.

---
```
meson build -D introspection=false -D vapi=false -D docs=true
```
should not work, because documentation requires introspection.

bengal:
Fixed some issues and repushed.

tools/meson-post-install.sh
```diff
@@ -0,0 +1,40 @@
...
+localstatedir=$5
+
+[ -n "$DESTDIR" ] && DESTDIR=${DESTDIR%%/}/
+
+mv "${DESTDIR}${datadir}/bash-completion/completions/nmcli-completion" \
+   "${DESTDIR}${datadir}/bash-completion/completions/nmcli"
+
+for alias in nmtui-connect nmtui-edit nmtui-hostname; do
+    ln -sf nmtui "${DESTDIR}${bindir}/$alias"
+done
```
thom311: if tui isn't enabled, these symlinks shouldn't be installed.

thom311:
With
```
  $ meson build -D introspection=false -D vapi=true
```
we get
```
meson.build:717: WARNING: Vala bindings are disabled because they require GObject introspection
```
but it seems to me a hard failure would be in order.

tools/meson-post-install.sh
bengal: Fixed.

bengal:
> but it seems to me a hard failure would be in order.

Yeah, I changed the option to a auto|true|false choice.

thom311:
lgtm

PR #206, opened 2018-09-16
IWD fixes

balrog-kun:


vhumpa:

src/devices/wifi/nm-iwd-manager.c
```diff
@@ -786,6 +786,20 @@ nm_iwd_manager_is_known_network (NMIwdManager *self, const char *name,
...
 }
 
+GDBusProxy *
+nm_iwd_manager_get_dbus_interface (NMIwdManager *self, const char *path,
+                                   const char *name)
+{
+	NMIwdManagerPrivate *priv = NM_IWD_MANAGER_GET_PRIVATE (self);
+	GDBusInterface *interface;
+
+	g_return_val_if_fail (priv->object_manager, NULL);
```
thom311: asserting that the object_manager is set, seems cumbersome, if not a bug.
At least, in device's `act_stage2_config()`, it's not clear that the object manager is set. That is, because the object-manager can be unset any time during a name-owner-changed event.

balrog-kun: Right, fixing this, I actually meant to just return if `priv->object_manager` was unset. `act_stage2_config` probably wouldn't be called after the device state changes to unavailable though.

lkundrak:
_Ack_ from me

thom311:
merged to master.

thanks Andrew!

PR #208, opened 2018-09-19
lr/wpan-channel: add ability to set a channel

lkundrak:
Fixes https://www.phoronix.com/forums/forum/software/desktop-linux/1046289-networkmanager-1-14-rc1-shows-wireguard-interfaces-sr-iov-device-configuration-6lowpan?p=1046311#post1046311


thom311:
please rebase on master, so that CI may run. The current HEAD fails to build due to an issue not related to this branch

libnm-core/nm-setting-wpan.c
```diff
@@ -270,5 +357,33 @@ nm_setting_wpan_class_init (NMSettingWpanClass *klass)
...
+	 * IEEE 802.15.4 channel page.
+	 **/
+	g_object_class_install_property
+	        (object_class, PROP_PAGE,
+	         g_param_spec_int (NM_SETTING_WPAN_PAGE, "", "",
+	                            NM_SETTING_WPAN_PAGE_DEFAULT,
+		                    IEEE802154_MAX_PAGE,
+		                    NM_SETTING_WPAN_PAGE_DEFAULT,
+	                            G_PARAM_READWRITE |
+	                            G_PARAM_STATIC_STRINGS));
```
thom311: indentation

thom311: If you restrict the range here, you don't need to `verify()` it. But restricting the range in the GOjbect property is generally not a good idea, because it means the range cannot be extended. It may be better to assume a range of `G_MININT32 .. G_MAXINT32` and validate it in `verify()`.

Also, `verify()` validates that the value is not larger than MAX, but it does not validate that the value is smaller than `-1`. It should (provided, you verify the range there).

libnm-core/nm-setting-wpan.c
```diff
@@ -270,5 +357,33 @@ nm_setting_wpan_class_init (NMSettingWpanClass *klass)
 	                            G_PARAM_READWRITE |
 	                            G_PARAM_STATIC_STRINGS));
 
+	/**
+	 * NMSettingWpan:page:
+	 *
+	 * IEEE 802.15.4 channel page.
```
thom311: `Since:`

Also, can you describe a bit more, about the valid range, and the default value -1.

libnm-core/nm-setting-wpan.c
```diff
@@ -270,5 +357,33 @@ nm_setting_wpan_class_init (NMSettingWpanClass *klass)
...
+		                    IEEE802154_MAX_PAGE,
+		                    NM_SETTING_WPAN_PAGE_DEFAULT,
+	                            G_PARAM_READWRITE |
+	                            G_PARAM_STATIC_STRINGS));
+
+	/**
+	 * NMSettingWpan:channel:
+	 *
+	 * IEEE 802.15.4 channel.
+	 **/
```
thom311: `Since:`

and, describe a bit more what this is, that it needs to be configured together with the page, that the range is so and so, and that -1 is the default/unset.

libnm-core/nm-setting-wpan.c
```diff
@@ -26,22 +26,34 @@
...
+	PROP_PAGE,
+	PROP_CHANNEL,
 };
 
 typedef struct {
 	char *mac_address;
 	guint16 pan_id;
 	guint16 short_address;
+	gint16 page;
+	gint16 channel;
```
thom311: `gint16` is larger that the range of the property below (which is restricted to `-1` ... `IEEE802154_MAX_*`.

Below, I suggested, that the range of the GObject property should not be restricted.

either way, choose a int type that fits the size of the range from the GObject property. As it is now, gint8 would suffice. If the range is G_MININT32/G_MAXINT32, it would need gint32.

libnm-core/nm-setting-wpan.c
```diff
@@ -117,6 +129,38 @@ nm_setting_wpan_get_short_address (NMSettingWpan *setting)
...
 
+/**
+ * nm_setting_wpan_get_page:
+ * @setting: the #NMSettingWpan
+ *
+ * Returns: the #NMSettingWpan:page property of the setting
+ *
+ * Since: 1.16
+ **/
+gint16
```
thom311: The type should match the range of the Object property.

src/devices/nm-device-wpan.c
```diff
@@ -182,6 +183,15 @@ act_stage1_prepare (NMDevice *device, NMDeviceStateReason *out_failure_reason)
...
+		page = nm_setting_wpan_get_page (s_wpan);
+		if (!nm_platform_wpan_set_channel (platform, ifindex, page, channel)) {
+			_LOGW (LOGD_DEVICE, "unable to set the channel");
+			goto out;
+		}
+	}
+
 	ret = NM_ACT_STAGE_RETURN_SUCCESS;
 out:
         nm_device_bring_up (device, TRUE, NULL);
```
thom311: can you drive-by fix indentation?

libnm-core/nm-setting-wpan.c
lkundrak: Fixed; along with the checkpatch.pl changes to detect the whitespace sins.

src/devices/nm-device-wpan.c
lkundrak: okay...

PR #210, opened 2018-09-25
bg/preserve-routes-down-rh1626004

bengal:
Preserve the device route when the interface is brought down and up again.


thom311:
lgtm

PR #212, opened 2018-09-26
Add meson support to RPM spec file

bengal:
Allowing building the RPM using meson.


introspection/meson.build
```diff
@@ -53,23 +53,31 @@ ifaces = [
...
     docbook: 'dbus'
   )
 
-  # FIXME: gnome.gdbus_codegen should return the docbook as a target result
+  sources += res
+
   content_files += join_paths(meson.current_build_dir(), 'dbus-' + iface_xml)
+
+  # FIXME: res is an array only since 0.46. Documentation won't work
+  # with older versions
```
thom311: `FIXME` makes it sound like there would be something to fix. But there isn't. The issue is handled as best as it could. Either `s/FIXME:/Note:/` or `s/FIXME: //`. Alternatively, the fixme comment should vaguely describe how the fix looks like.

PR #213, opened 2018-09-26
acd: make NMAcdManager no GObject

thom311:
NMAcdManager is a rather simple instance.

It does not need (thread-safe) ref-counting, in fact, having
it ref-counted makes it slighly ugly that we connect a signal,
but never bother to disconnect it (while the ref-counted instance
could outlife the signal subscriber).

We also don't need GObject signals. They have more overhead
and are less type-safe than a regular function pointers. Signals
would make sense, if there could be multiple independent listeners,
but that just doesn't make sense.

Implementing it as a plain struct is less lines of code, and less
runtime over head.

Also drop the possiblitiy to reset the NMAcdManager instance.
It wasn't needed and I think it was buggy because it wouldn't
reset the n-acd instance.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1668,12 +1668,12 @@ src_libNetworkManager_la_SOURCES = \
 	src/nm-checkpoint-manager.c \
 	src/nm-checkpoint-manager.h \
 	\
-	src/devices/nm-device.c \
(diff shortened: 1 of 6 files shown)


thom311:
merged

PR #214, opened 2018-09-27
Avoid warnings when configuring IPv6 tokens

bengal:
Kernel complains with:

  platform-linux: link: change 10: token: set IPv6 address generation token to ::bbbb
  platform-linux: do-change-link[10]: failure changing link: failure 22 (Invalid argument)

if we try to set an IPv6 token in ip_config_merge_and_apply() and we
haven't set accept_ra=1 yet. Since the flag is set when starting
router discovery, ensure that we don't try to set a token before that.

https://bugzilla.redhat.com/show_bug.cgi?id=1560652


thom311:
lgtm

PR #216, opened 2018-10-01
device: always cancel WPS when secret-request fails

thom311:
See the logfile at [1], for how activating is attempted first and waits
30 seconds for WPS to fail. I think that is undesired, because a secret
agent registered in the meantime.

A possible fix would be, that when
  - WPS is pending
  - the secret request already failed
  - another secret-agent registers
then the activation (and WPS) is aborted and autoconnect may be tried
again, possibly with secrets provided by the new secret-agent.

However, this patch goes a step further: it always cancels activation
when the secret request fails. That means, WPS only works while the
user is also prompted for a secret. That makes sense to me, because
an action from the user is required. However, without secret prompt,
the user wouldn't be aware of that and is unlikely to press the WPS
push botton.

[1] https://bugzilla.opensuse.org/show_bug.cgi?id=1079672#c33

### DIFF
--- a/src/devices/wifi/nm-device-wifi.c
+++ b/src/devices/wifi/nm-device-wifi.c
@@ -1685,20 +1685,37 @@ wifi_secrets_cb (NMActRequest *req,
 	g_return_if_fail (nm_act_request_get_settings_connection (req) == connection);
 
 	if (error) {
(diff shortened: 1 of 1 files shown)


thom311:
merged to master.

PR #217, opened 2018-10-01
bg/dhcp-failed-rh1625901 (wip)

bengal:
Don't start DHCP grace period when the client is not running.

https://bugzilla.redhat.com/show_bug.cgi?id=1625901


thom311:
lgtm

thom311:
v2 lgtm

PR #220, opened 2018-10-04
bg/covscan: fix some coverity warnings

bengal:


libnm-core/nm-setting-team.c
```diff
@@ -125,9 +125,9 @@ nm_team_link_watcher_new_ethtool (int delay_up,
 	NMTeamLinkWatcher *watcher;
 	const char *val_fail = NULL;
 
-	if (delay_up < 0 || delay_up > G_MAXINT32)
+	if (delay_up < 0)
```
thom311: I don't know. This is not so great. The checks against G_MAXINT32 do make sense (presuming, that gint could be larger than 32 bits -- which they are not in practice). Is there any glib architecture, where int != gint32?

bengal: I think on Linux an int is 32 bit on all architectures. But if we want to keep the check just to be sure, it's fine with me.

bengal:
How about now?

libnm-core/nm-setting-team.c
thom311: Why not https://gist.github.com/thom311/362c5f905470e3580b98299198692a5b ?
it is similar in nature to existing `_NM_INT_NOT_NEGATIVE()`.

This gives the concept a name, and by looking at the code, the reviewer can obviously see what the purpose was. That is different from the implicit knowledge that int==gint32 (and that there is a static-assert for that somewhere).

bengal: Looks good, I've updated the commit.

bengal:
Merged.

PR #221, opened 2018-10-05
IWD AP and Ad-hoc modes

balrog-kun:


vhumpa:

src/devices/wifi/nm-device-iwd.c
```diff
@@ -559,23 +596,46 @@ check_connection_compatible (NMDevice *device, NMConnection *connection, GError
...
+			if (!is_connection_known_network (connection)) {
+				nm_utils_error_set_literal (error, NM_UTILS_ERROR_CONNECTION_AVAILABLE_TEMPORARY,
+				                            "802.1x connections must have IWD provisioning files");
+				return FALSE;
+			}
+		} else if (!NM_IN_SET (security, NM_IWD_NETWORK_SECURITY_NONE, NM_IWD_NETWORK_SECURITY_PSK)) {
 			nm_utils_error_set_literal (error, NM_UTILS_ERROR_CONNECTION_AVAILABLE_TEMPORARY,
-			                            "802.1x profile is not a known network");
+			                            "IWD backend only supports Open, PSK and 802.1x network "
+						    "authentication in Infrastructure mode");
```
thom311: indentation

src/devices/wifi/nm-device-iwd.c
```diff
@@ -440,37 +442,58 @@ cleanup_association_attempt (NMDeviceIwd *self, gboolean disconnect)
...
+	GError *error = NULL;
 
-	ctx->callback (G_OBJECT (ctx->self), res, ctx->user_data);
+	variant = g_dbus_proxy_call_finish (G_DBUS_PROXY (source), res, &error);
+	if (variant)
+		g_task_return_boolean (task, TRUE);
+	else if (!nm_utils_error_is_cancelled (error, FALSE))
+		g_task_return_error (task, error);
+	else
+		return;
```
thom311: doesn't this leak `task`? Why not?

src/devices/wifi/nm-device-iwd.c
```diff
@@ -481,15 +504,24 @@ deactivate_async (NMDevice *device,
 {
 	NMDeviceIwd *self = NM_DEVICE_IWD (device);
 	NMDeviceIwdPrivate *priv = NM_DEVICE_IWD_GET_PRIVATE (self);
-	DeactivateContext *ctx;
+	GTask *task;
```
thom311: until now, we didn't use GTask (at all). Obviously, iwd doesn't use glib at all, but compared to Connman, which also doesn't use GTask either (and prides itself of not using gobject). Is using this API a good idea? How would it look like without this API?

To me, GTask seems to provide features we don't need. @bengal, WDYT?

balrog-kun: I can easily change this to GSimpleAsyncResult. I'd never used either before but chose GTask because GSimpleAsyncResult is documented as deprecated in favour of GTask.

thom311: OK. I think we should get rid of all GAsyncResult/GTask API as much as we can. Since NMDeviceClass.deactivate_async() follows this pattern, it bubbles all the way down. This should be refactored (will do separately). The patch is good for the moment.

balrog-kun: Ok! The only instances are in device/wwan and device/bluetooth so I assume this won't conflict with this patch and will be done afterwards.

balrog-kun:
Oh BTW one problem with this code is that the psk is not available in the wireless-security setting in the applied connection when an AP connection is activated, I undestand there's some reasoning behind clearing the secrets on the applied connection.  Instead I read the PSK from the settings connection but that still sometimes fails after the connection is used a second time, and I have no idea why.  I guess I should be using nm_act_request_get_secrets (was going to do this in a separate patch) but I couldn't yet understand from the code exactly why.

bengal:
the branch looks fine to me

PR #223, opened 2018-10-07
all: remove \n endings from log calls

lkundrak:
The extra newlines look bad when logging to the console.

### DIFF
--- a/src/dns/nm-dns-systemd-resolved.c
+++ b/src/dns/nm-dns-systemd-resolved.c
@@ -129,7 +129,7 @@ call_done (GObject *source, GAsyncResult *r, gpointer user_data)
 	if (!v) {
 		if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
 			return;
-		_LOGW ("Failed: %s\n", error->message);
+		_LOGW ("Failed: %s", error->message);
 		g_error_free (error);
 	}
 }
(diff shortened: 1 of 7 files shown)


src/dns/nm-dns-systemd-resolved.c
```diff
@@ -129,7 +129,7 @@ call_done (GObject *source, GAsyncResult *r, gpointer user_data)
 	if (!v) {
 		if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
 			return;
-		_LOGW ("Failed: %s\n", error->message);
+		_LOGW ("Failed: %s", error->message);
```
thom311: while at it, can you please improve the logging messges?

`"Failed: %s"` is not good. `"%s"` below even less good.

Optimally, every logging line is unique enough, that (given the sources) one can unambiguously find the line which logged the message. That means, `error->message` should be part of a longer sentence that is unique among all logging messages. Arguably, `"Failed: %s"` is currently rather unique. But it also is not very expressive: *what* failed?

Yes, ` src/dns/nm-dns-systemd-resolved.c` was already pre-existing (not your doing), sorry for pointing this out here. I don't "blame" you personally. Would be great if you already look at these logging messages, that you take the burden of improving them. Thanks.

lkundrak: Is "no" a good answer here? Does it block a review?

It's somewhat irrelevant to otherwise (intentionally) mechanical change. Also, it seems to me that a much better approach to adjust the logs to your liking (that I definitely don't object) would be to just do the change -- it could easily be less effort than asking me to do it and I'd be happy to review (esp. these days when I'm generally busier with changing diapers than with changing code).

thom311: Sure.

PR #225, opened 2018-10-09
th/cli-drop-some-global

thom311:
I think, nmcli should use less global variables, and prefer to pass around the state that it requires. Likewise, it should aim to pass around less state, or immutable state.

A small branch, with some cleanup in that direction.


thom311:
thanks. merged.

PR #226, opened 2018-10-10
cli: check we have an active connection before showing hint

bengal:
If the activation fails even before the active connection instance is
created, we get the following:

 $ nmcli connection up vpn1
 libnm-CRITICAL **: nm_active_connection_get_connection: assertion 'NM_IS_ACTIVE_CONNECTION (connection)' failed
 nmcli-CRITICAL **: active_connection_hint: assertion 'connection' failed
 Error: Connection activation failed: Not authorized to control networking.

Check that we have an active connection before showing the hint.

Fixes: bc6c042d5461028a2fac53ffcdb36837e596a1a5

### DIFF
--- a/clients/cli/connections.c
+++ b/clients/cli/connections.c
@@ -2352,6 +2352,9 @@ active_connection_hint (GString *return_text,
 	const GPtrArray *devices;
 	guint i;
 
+	if (!active)
+		return;
+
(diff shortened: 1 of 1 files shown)


thom311:
lgtm, but I think it needs slight adjustments after rebase

PR #227, opened 2018-10-11
th/no-gasyncresult

thom311:
I think the `GAsyncResult` pattern is less than optimal.

Maybe it serves a purpose in public glib API, where it better works for language bindings/introspection. For NetworkManager core, it's just a nuisance.

It results in
 - more lines of code (look at the patch)
 - more indirection of calling layers of code
 - less control of what actually happens because of the additional layers of code
 - less type-safety (instead of directly having a callback with suitable arguments, there is a `GAsyncReadyCallback`. Also, `g_simple_async_result_set_op_res_gpointer()` etc. looses type information).
 - more resource overhead (`GAsyncResult` is itself a `GObject`, ref-counted and unnecessary "features". In most cases, we need a simple heap allocation to package the user-data and relevant state. `nm_utils_user_data_pack()` is enough).

### DIFF
--- a/libnm-core/nm-core-internal.h
+++ b/libnm-core/nm-core-internal.h
@@ -305,6 +305,10 @@ void _nm_dbus_errors_init (void);
 
 extern gboolean _nm_utils_is_manager_process;
 
+gboolean _nm_dbus_typecheck_response (GVariant *response,
+                                      const GVariantType *reply_type,
+                                      GError **error);
+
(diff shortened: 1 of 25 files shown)


thom311:
there are quite some FIXME comments. I may address them still in this branch...

bengal:
LGTM

thom311:
merged

PR #228, opened 2018-10-11
th/ndisc-addr-lifetime

thom311:
ndisc: fix updating address lifetime on Router Announcement according to RFC4862
    
This is a denial-of-service protection, where a malicious router advertisement can expire the addresses.
    
See-also: https://github.com/systemd/systemd/commit/6554550f35a7976f9110aff94743d3576d5f02dd
See-also: https://tools.ietf.org/search/rfc4862#section-5.5.3
    
https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/1796622

### DIFF
--- a/src/ndisc/nm-lndp-ndisc.c
+++ b/src/ndisc/nm-lndp-ndisc.c
@@ -221,10 +221,10 @@ receive_ra (struct ndp *ndp, struct ndp_msg *msg, gpointer user_data)
 				.preferred = ndp_msg_opt_prefix_preferred_time (msg, offset),
 			};
 
-			if (address.preferred > address.lifetime)
-				address.preferred = address.lifetime;
-			if (nm_ndisc_complete_and_add_address (ndisc, &address))
-				changed |= NM_NDISC_CONFIG_ADDRESSES;
+			if (address.preferred <= address.lifetime) {
+				if (nm_ndisc_complete_and_add_address (ndisc, &address))
+					changed |= NM_NDISC_CONFIG_ADDRESSES;
+			}
 		}
 	}
 	ndp_msg_opt_for_each_offset(offset, msg, NDP_MSG_OPT_ROUTE) {
--- a/src/ndisc/nm-lndp-ndisc.c
+++ b/src/ndisc/nm-lndp-ndisc.c
@@ -222,7 +222,7 @@ receive_ra (struct ndp *ndp, struct ndp_msg *msg, gpointer user_data)
 			};
 
 			if (address.preferred <= address.lifetime) {
-				if (nm_ndisc_complete_and_add_address (ndisc, &address))
+				if (nm_ndisc_complete_and_add_address (ndisc, &address, now))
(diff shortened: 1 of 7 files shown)


thom311:
Also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/57

src/ndisc/nm-lndp-ndisc.c
```diff
@@ -221,10 +221,10 @@ receive_ra (struct ndp *ndp, struct ndp_msg *msg, gpointer user_data)
...
 			};
 
-			if (address.preferred > address.lifetime)
-				address.preferred = address.lifetime;
-			if (nm_ndisc_complete_and_add_address (ndisc, &address))
-				changed |= NM_NDISC_CONFIG_ADDRESSES;
+			if (address.preferred <= address.lifetime) {
+				if (nm_ndisc_complete_and_add_address (ndisc, &address))
+					changed |= NM_NDISC_CONFIG_ADDRESSES;
+			}
```
bengal: Should we also check 5.5.3.b:

>     b)  If the prefix is the link-local prefix, silently ignore the
>       Prefix Information option.

? Especially because later we assert:

```
	nm_assert (!IN6_IS_ADDR_UNSPECIFIED (&new->address));
	nm_assert (!IN6_IS_ADDR_LINKLOCAL (&new->address));
```

src/ndisc/nm-ndisc.c
```diff
@@ -421,20 +422,33 @@ nm_ndisc_add_address (NMNDisc *ndisc, const NMNDiscAddress *new)
...
-			*item = *new;
+		if (new->lifetime == 0) {
+			g_array_remove_index (rdata->addresses, i);
 			return TRUE;
 		}
+
+		if (   get_expiry (item) == get_expiry (new)
+		    && get_expiry_preferred (item) == get_expiry_preferred (new)
+		    && (   from_ra
+		        || item->dad_counter != new->dad_counter))
```
bengal: Shouldn't this be `item->dad_counter == new->dad_counter` ?

src/ndisc/nm-lndp-ndisc.c
thom311: I think that is done already:
- for RAs: https://github.com/NetworkManager/NetworkManager/blob/cec7ade86c2ac2a20a5b3cdabcf375ece347dd76/src/ndisc/nm-lndp-ndisc.c#L198
- for shared-mode: https://github.com/NetworkManager/NetworkManager/blob/cec7ade86c2ac2a20a5b3cdabcf375ece347dd76/src/devices/nm-device.c#L3548

Indeed, the latter doesn't cover `IN6_IS_ADDR_UNSPECIFIED()`. Adding a patch for that...

fgiudici:
LGTM

src/ndisc/nm-ndisc.c
thom311: true

thom311:
merged to master. Backports to stable-branches still TODO.

PR #230, opened 2018-10-12
systemd: don't make NetworkManager D-Bus activatable

lkundrak:
If the NetworkManager daemon has been stopped manually we don't want it
to be autostarted by a client request.

[lkundrak@v3.sk: The auto-activation is probably more surprising than useful.
Services that need NetworkManager API should depend on NetworkManager service
directly.

I have no idea what purpose does the D-Bus service file serve nowadays,
but it looks rather hacky (really, activating /bin/false) and the comment
in it suggests that the autoactivating behavior was not intended anyway.
Debian has been shipping this for quite some time and no complains have been
heard.]


lkundrak:
> But don't you want to delete the file data/org.freedesktop.NetworkManager.service.in as well?

Oh, good catch, I indeed intended to.

I think I did a `git rm data/org.freedesktop.NetworkManager.service` without realizing the file is templated. Why was it even templated, there's nothing to expand there...

PR #231, opened 2018-10-12
th/move-to-shared

thom311:
The aim is that our code under `src/` is less entangled.

A minor reason, that it can be used independently. A bigger reason, to improve our code structure.

`shared/nm-utils` is a collection of bits that are independent of libnm or NetworkManager daemon. Move `nm_utils_get_monotonic_timestamp*()` there, and add a test, to see whether the shared component can be build independently.

also, make "nm-logging.c" only depend on "shared/nm-utils". With that, I think platform itself could become an independent library. Again, the purpose may not be to use platform alone, but to add clearer sepration. Note however, that `src/initrd/nmi-ibft-reader.c` does use platform albeit not being part of the daemon process. So, it is already multi purpose.


src/nm-core-utils.c
```diff
@@ -1931,17 +1931,7 @@ monotonic_timestamp_get (struct timespec *tp)
...
-		struct tm tm;
-		char s[255];
-
-		strftime (s, sizeof (s), "%Y-%m-%d %H:%M:%S", localtime_r (&now, &tm));
-		nm_log_dbg (LOGD_CORE, "monotonic timestamp started counting 1.%09ld seconds ago with "
-		                       "an offset of %lld.0 seconds to %s (local time is %s)",
-		                       tp->tv_nsec, (long long) -monotonic_timestamp_offset_sec,
-		                       clock_mode == 1 ? "CLOCK_BOOTTIME" : "CLOCK_MONOTONIC", s);
-	}
+	_nm_utils_monotonic_timestamp_initalized (tp, monotonic_timestamp_offset_sec, clock_mode == 1);
```
bengal: initalized -> initialized

Makefile.am
```diff
@@ -204,6 +204,34 @@ DISTCLEANFILES += $(polkit_policy_DATA)
...
+	shared/nm-utils/tests/test-shared-general.c \
+	$(NULL)
+
+shared_nm_utils_tests_test_shared_general_LDADD = \
+	$(GLIB_LIBS) \
+	shared/libcsiphash.la \
+	$(NULL)
+
+###############################################################################
+
```
bengal: Can you also add this test to meson?

src/nm-core-utils.c
thom311: fixed

shared/nm-utils/tests/test-shared-general.c
```diff
@@ -0,0 +1,77 @@
+/* -*- Mode: C; tab-width: 4; indent-tabs-mode: t; c-basic-offset: 4 -*- */
```
lkundrak: A nit: could we perhaps stop adding these?
A source file is *not* the right place to configure anyone's editor.

shared/nm-utils/nm-test-utils.h
```diff
@@ -1089,6 +1091,8 @@ __define_nmtst_static(02, 1024)
 __define_nmtst_static(03, 1024)
 #undef __define_nmtst_static
 
+#if defined (__NM_UTILS_H__) || defined (NM_UTILS_H)
```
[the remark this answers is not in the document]
thom311: `shared/nm-utils/nm-test-utils.h` does this at several places already:

```

#ifdef NM_SETTING_IP_CONFIG_H
static inline void
...
#endif /* NM_SETTING_IP_CONFIG_H */
```
and
```
#if defined(__NM_SIMPLE_CONNECTION_H__)
»···»···type_g = nm_setting_lookup_type (type);
#else
»···»···type_g = nm_connection_lookup_setting_type (type);
#endif
```

The reason is, that `nm-test-utils.h` is a header-only "library" of testing utilities. It is used by all our tests. The right way to use it is:
  - including "nm-test-utils.h" as last file. 
  - before, including other headers that turn on functionality in "nm-test-utils.h".

The alternative, would either require multiple header files, and you explicitly include them. Or even not using a header-only approach. IMO both makes it only more cumbersome to write tests. As it is, it's convenient to use. Also, almost all our tests do either include "libnm", "libnm-utils", or "libnm-core" (thus, usually, the code is available). This test is an exception, but other parts of `nm-test-utils.h` are still interesting to have here.

It's not documented, but it's actually impossible to get wrong, because you either lack functionality (and don't care), or your code won't compile.

This also means our include guards are meaningful and should have well-defined names. `#pragma once` would not work here. But I don't mind.

shared/nm-utils/tests/test-shared-general.c
thom311: Done. Plus added check to `checkpatch.pl`. ACK that too?

Makefile.am
thom311: Done. How about now?

Yes, we currently build the same source files multiple times. That is wrong, we should build intermedate static libraries. However, the reason is, that currently we pass different CFLAGS depending on where the shared files are used. So, cleaning this up, is more effort (and beyond the point of this branch). TODO.

bengal: LGTM

PR #232, opened 2018-10-15
[th/select-device-for-activation] core: prefer managed device over unmanaged ones when activating profile

thom311:
With

  $ nmcli connection up "$PROFILE" ifname "$DEVICE"

it's clear that the user means the particular device. That also
is taken as a command to make $DEVICE as managed, in case it was
unmanaged before.

If the user just issues

  $ nmcli connection up "$PROFILE"

without a particular device, then we should prefer devices which
are already marked as managed. In fact, when we have multiple
candidates, select the best candidate based on the device's state.

Also, if the profile is not multi-connect "single", no longer prefer
the same device on which the profile is already active.

https://bugzilla.redhat.com/show_bug.cgi?id=1639254

### DIFF
--- a/src/nm-manager.c
+++ b/src/nm-manager.c
@@ -3328,6 +3328,49 @@ nm_manager_get_devices (NMManager *manager)
 	return &NM_MANAGER_GET_PRIVATE (manager)->devices_lst_head;
 }
 
+typedef enum {
+	DEVICE_ACTIVATION_PRIO_NONE,
+	DEVICE_ACTIVATION_PRIO_UNMANAGED,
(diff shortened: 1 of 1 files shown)


thom311:
this also fixes a problem, that if you have a profile which applies to multiple device, and there are two suitable candidates. If one candidate is already active with another profile, then `nmcli connection up `$PROFILE"` will choose the disconnected device. That simply makes more sense I think.

Of course, the robust solution is always, to either explicitly specify the device during `nmcli connection up`, or restrict the profile to only one device.

thom311:
merged

thom311:
Follow up fixes merged as 9cf64e2ee23fc1bebe13196460a78a4ee8651ade

PR #234, opened 2018-10-17
[RFC] bg/vpn-ask-username-rh1535517

bengal:
Add a new vpn property indicating that the username should be interactively asked to agents every time the connection gets activated.

The branch introduces new mechanism that uses the `x-vpn-interactive-` prefix for hints in the secret request to tell the agent that those are not connection properties but another type of secret that must be asked every time and not stored to disk. This prefix can be used for example also to ask the answer to a dynamic challenge with OpenVPN.

Perhaps this solution is a bit hacky and a better approach would be to change the format of hints in the API from `(as)` to something more structured like `(a(uus))` where each hint contains a type, flags and a value string. For example, the type could be one of:

 - message (replaces the current `x-vpn-message` hint)
 - secret-property (replace the current standard hints)
 - secret-interactive (e.g. username, dynamic-challenge)
 - info (arbitrary informative field, for example to show the username) 

and flags:

 - echo_value: whether the value entered by user should be echoed or not. For example, for the username it would be TRUE, while for a password FALSE.
 - ignore_secrets_from_setting: this tells whether the agent should also ask to user secrets from the setting in addition to hints. For example, if the hint only contains the username the agent should also ask secrets from the setting; but when hints contain a password it means that the password was wrong and it should only ask a password.

This would make the code more generic but requires extending the current D-Bus API:

 * SecretAgent.GetSecrets()
 * VPNPlugin.SecretsRequired()
 * VPNPlugin.NewSecrets()

and also the internal one, adding conversion functions between one format and the other. This is requires non trivial effort. Any opinion?


clients/common/nm-secret-agent-simple.c
```diff
@@ -392,7 +392,18 @@ add_vpn_secret_helper (GPtrArray *secrets, NMSettingVpn *s_vpn, const char *name
 		                                            NM_SETTING (s_vpn),
 		                                            name,
 		                                            nm_setting_vpn_get_service_type (s_vpn));
+	} else if (g_str_has_prefix (name, NM_VPN_INTERACTIVE_SECRET_PREFIX)) {
```
thom311: this is indeed quite ugly. Above you pass `name` to `get_vpn_secret_flags()`, like it was a regular secret which would be in the `s_vpn` dictionary. Do you expect to find name there? Otherwise, you should first check for that. Actually, the entire order is ugly.

Below, `add_vpn_secrets()` calls `add_vpn_secret_helper()` with `p->name`. Good. In that case, `p->name` is clearly a secret in `s_vpn`. But the other caller of `add_vpn_secret_helper()` assumes that the hints name can be found in `s_vpn`? It seems so ugly.

thom311:
It doesn't seem a problem in principle to extend the D-Bus API with a new request, is it?
Especially, because the secret agents register at the server, and provide "capabilities". So, the secret agent can indicate whether it supports the new API.

thom311:
`clients/common` also asks for non-secrets, like username (see `NM_SECRET_AGENT_SECRET_TYPE_PROPERTY`). That of course does not work, because a secret-request (currently) cannot inject non-secret properties. 

This code seems to come from https://gitlab.gnome.org/GNOME/gnome-shell/commit/2ebdc81c8f68b0e32e71cab5f522c6bd5a83997d, but I don't think that ever worked...

vbenes:
still see some crash:
http://faf.lab.eng.brq.redhat.com/faf/reports/11154/

glats:
Hi! i'm wondering if this still going.
Thans!

PR #237, opened 2018-10-18
th/nm-1-14-keyfile-changes

thom311:
We are about to release 1.14.2

nm-1-14 branch recently saw backports of lr/initrd branch. This introduced a new keyfile directory `/var/run/NetworkManager/system-connections`.

I think it's important to iron details out before release, because this affects how keyfiles are persisted to disk. It will be hard to do changes later which prevent to read configurations written by an older version of NetworkManager.

These are all clean backports from master branch, that bring two bits:

- files in /run must have a file extension `.nmconnection`.
- keyfiles allow that `connection.uuid` is missing. In that case it is computed by MD5 hashing the full path. The problem is now, if we have two files `/etc/NetworkManager/system-connection/lala.nmconnection` and `/run/NetworkManager/system-connection/lala.nmconnection`, then they both should generate the same UUID. This is also important, if you copy a keyfile from one directory to another. Then the UUID shouldn't change. The solution is here, to always hash the etc-dirname + basename, regardless of the actual path.

Open pull request, mainly for CI.

### DIFF
--- a/libnm-core/nm-keyfile-internal.h
+++ b/libnm-core/nm-keyfile-internal.h
@@ -95,7 +95,6 @@ typedef struct {
 } NMKeyfileReadTypeDataWarn;
 
 NMConnection *nm_keyfile_read (GKeyFile *keyfile,
-                               const char *keyfile_name,
                                const char *base_dir,
                                NMKeyfileReadHandler handler,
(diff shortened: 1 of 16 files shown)


bengal:
LGTM

PR #238, opened 2018-10-19
th/file-is-in-path

thom311:
add a helper function for checking whether a file is inside a patch. With this,
`nmcli connection load //etc/NetworkManager/system-connections/profile` no longer fails

Also, there is something new here: I expose and use API from internal `src/systemd`. Up to now, we avoided that, and only used a minimal of systemd code that we required. I think there are a loot of goodies there, and we should carefully pick a few that we like and use them. Of course, we need to add unit-tests for them (ensuring that a future systemd import won't break them).

Also, don't directly include systemd code, because then it easily gets out of hand. Instead, have one particular header `src/systemd/nm-sd-utils.h` which exposes this functionality in a clean way.

### DIFF
--- a/src/NetworkManagerUtils.c
+++ b/src/NetworkManagerUtils.c
@@ -34,6 +34,7 @@
 
 #include "platform/nm-platform.h"
 #include "nm-auth-utils.h"
+#include "systemd/nm-sd-utils.h"
 
 /*****************************************************************************/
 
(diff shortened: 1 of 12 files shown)


thom311:
ahh, I forgot.

> core: add "nm-sd-utils.h" to access system internal helper

exposes 2 more functions (`nm_sd_utils_path_equal()` and `nm_sd_utils_path_simplify()`). I used them on another WIP branch, but unexpectedly didn't use it here. So, they are still here. I would prefer to merge them, to keep the other WIP branch smaller. They will be used later.

src/systemd/nm-sd-utils.c
```diff
@@ -0,0 +1,45 @@
+/* This library is free software; you can redistribute it and/or
```
bengal: I like the idea of reusing code, but I'm worried that this will make imports of systemd code more complex when the internal systemd API changes.

 As long as we reuse only few functions, I think this is useful.

src/systemd/src/basic/path-util.c
```diff
@@ -372,9 +375,14 @@ char *path_simplify(char *path, bool kill_dots) {
...
-        if (absolute && t == path)
-                *(t++) = '/';
+        /* Special rule, if we stripped everything, we either need a "/" (for the root directory)
+         * or "." for the current directory */
+        if (t == path) {
+                if (absolute)
+                        *(t++) = '/';
+                else
+                        *(t++) = '.';
+        }
```
bengal: Won't this create a conflict when the code changes in systemd and we synchronize it? We should probably have the fix accepted in systemd before.

src/NetworkManagerUtils.c
```diff
@@ -34,6 +34,7 @@
 
 #include "platform/nm-platform.h"
 #include "nm-auth-utils.h"
+#include "systemd/nm-sd-utils.h"
```
bengal: In commit message: patch -> path

src/systemd/nm-sd-utils.c
thom311: Agree. I mentioned, that this must be used carefully. Also, all the functions that I currently exposed are arguably trivial (why would systemd change them?) and they have unit-tests in NM. The rule should be, that such code must have unit-tests.

src/systemd/src/basic/path-util.c
thom311: it's already fixed in systemd upstream. I added a URL to the systemd commit to the commit message.

src/NetworkManagerUtils.c
thom311: fixed

thom311:
merged

PR #241, opened 2018-10-22
cli: initialize readline before installing the redisplay handler

lkundrak:
Otherwise readline decides to initialize terminal handling at the first
readline call, and if that happens at the point it sees our
non-echoing rl_redisplay.

At that point, unless already intialized, readline wrongly convinces itself we
do our own handling of terminal peculiarities (such as cursor movement, or
erases).  We do not -- we merely wrap the stock rl_redisplay(), temporarily
hiding the actual characters.

The rl_initialize() in nmc_readline_echo()s fixes broken line editing in
password prompts that weren't preceded a previous non-password prompt.
The other one is there for consistency only. (I guess we should be
initializing readline before use anyway; although it seems to initialize
itself anyway if we fail to do so...)

### DIFF
--- a/clients/cli/common.c
+++ b/clients/cli/common.c
@@ -941,6 +941,8 @@ nmc_readline (const NmcConfig *nmc_config,
 	va_list args;
 	gs_free char *prompt = NULL;
 
+	rl_initialize ();
+
 	va_start (args, prompt_fmt);
 	prompt = g_strdup_vprintf (prompt_fmt, args);
 	va_end (args);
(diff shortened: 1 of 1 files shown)


thom311:
lgtm, but from the commit message is not clear what is the actual issue (and how to reproduce it). Does this fix a bugzilla?

lkundrak:
> lgtm,

Thanks

> but from the commit message is not clear what is the actual issue (and how to reproduce it).

"The rl_initialize() in nmc_readline_echo()s fixes broken line editing in
password prompts that weren't preceded a previous non-password prompt."

This, in other words, would be: whenever it asks for a password and noting else (e.g. no username), the line editing doesn't work. At all.

> Does this fix a bugzilla?

No.

PR #242, opened 2018-10-22
shared/vpn-plugin-utils: load the editor from the same place as plugin

lkundrak:
If passed a relative path, load the editor .so from the same directory
as the plugin .so. This is useful for development, as it allows running
the editor plugin from the build tree conveniently.


thom311:
lgtm

PR #243, opened 2018-11-02
[th/dhcp-client-id-mac-rh1640494]

thom311:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1640494
Also a continuation of:
  - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/32
  - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/38
  - https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/40


vhumpa:
rebuild

vhumpa:
rebuild

vhumpa:
The beaker systems we use to test veth are currently down I understand. So I've changed those systems to different VMs temporalily to have this working until at least the gitlab ci is done (which being 'veth' can run on openstack for this code repo).

src/nm-core-utils.c
```diff
@@ -2308,60 +2308,142 @@ nm_utils_is_specific_hostname (const char *name)
...
-char *
-nm_utils_machine_id_read (void)
+/*****************************************************************************/
+
+static const UuidData *
+_machine_id_get (void)
 {
-	gs_free char *contents = NULL;
-	int i;
+	static const UuidData *volatile p_uuid_data;
```
bengal: why volatile, is it to make the function thread-safe? Since NM doesn't use threads I think we can ignore this problem

src/devices/nm-device.c
```diff
@@ -8271,56 +8266,61 @@ generate_duid_ll (const guint8 *hwaddr /* ETH_ALEN bytes */)
...
-	g_checksum_free (sum);
+		/* Hash the machine ID so it's not leaked to the network */
+		sum = g_checksum_new (G_CHECKSUM_SHA256);
+		g_checksum_update (sum, (const guchar *) machine_id, sizeof (*machine_id));
+		nm_utils_checksum_get_digest (sum, digest.sha256);
+
+		G_STATIC_ASSERT_EXPR (sizeof (digest.sha256) > sizeof (digest.uuid));
+		p = generate_duid_uuid (&digest.uuid);
+
+		if (!g_atomic_pointer_compare_and_exchange (&global_duid, NULL, p)) {
```
bengal: Likewise, can we just avoiding the extra code for thread-safety? Also because, like you said, it is not entirely thread-safe.

src/tests/test-general.c
```diff
@@ -1912,6 +1914,99 @@ test_machine_id_read (void)
...
+test_nm_sd_utils_generate_default_dhcp_client_id (gconstpointer test_data)
+{
+	const guint8 HASH_KEY[16] = { 0x80, 0x11, 0x8c, 0xc2, 0xfe, 0x4a, 0x03, 0xee, 0x3e, 0xd6, 0x0c, 0x6f, 0x36, 0x39, 0x14, 0x09 };
+	/* We run the test twice with two ifindexes.
+	 *
+	 * One is "1", which we expect to exist and having a name "lo".
+	 * The other is a random number, which we expect not to exist.
+	 *
+	 * Regardless of whether the ifindex actually exists, the tests are
+	 * supposed to pass. However, we our expectations are not met, we
```
bengal: s/we/when/

shared/nm-utils/nm-shared-utils.h
```diff
@@ -380,6 +380,46 @@ char **_nm_utils_strv_cleanup (char **strv,
...
+		\
+		nm_assert (_sum); \
+		\
+		_len = G_N_ELEMENTS (arr); \
+		\
+		g_checksum_get_digest (_sum, (arr), &_len); \
+		nm_assert (_len == G_N_ELEMENTS (arr)); \
+	} G_STMT_END
+
+#define nm_utils_checksum_get_digest_len(sum, buf, len) \
```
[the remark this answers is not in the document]
thom311: I chose the "len" suffix because it reminded me of `g_string_append_len()`. Sure, the "get-digest-len" name can be confusing. I didn't like "get-digest" for the name, but I chose it because the macro is a straight wrapper around `g_checksum_get_digest()`.

When criticizing naming, please suggest an alternative that is preferable :)

src/nm-core-utils.c
thom311: The code is not only used by NetworkManager binary, but also tests, nm-initrd-generator, nm-iface-helper, etc. These are all single-threaded (are they?). Note that glib does start multiple threads and GDBus is multithreadded API, however we are not supposed to use this function outside of the main thread. True.

These utility functions are at the bottom of the call-stack. I don't mind that NMManager is not thread-safe, because clearly it would be complicated, inefficient and unnecessary even to do so. But having lower layer function not thread-safe, seems very undesirable to me (out of principle). Code in `shared/nm-utils` and `src/nm-core-utils.c` and `libnm-core/nm-utils.c` should just be thread-safe.

Most of such singleton-patterns only require that we initialize *one* instance (and in case of races to to throw away the looser of the race). I wish, for such a trivial case we would find a preferred wat to implement it, which is
 1) thread-safe
 2) with little performance impact
 3) with little readability impact (simplicity)


About 1), I am indeed not entirely sure whether this is thread safe on all architectures that we support. I *think* it is, is it not? Same for `generate_duid_from_machine_id()`. I think that is actualyl thread-safe, isn't it?

src/devices/nm-device.c
thom311: maybe. See above.

src/tests/test-general.c
thom311: fixed

src/nm-core-utils.c
bengal: > The code is not only used by NetworkManager binary, but also tests, nm-initrd-generator, nm-iface-helper, etc. These are all single-threaded (are they?). Note that glib does start multiple threads and GDBus is multithreadded API, however we are not supposed to use this function outside of the main thread. True.

I agree.

> But having lower layer function not thread-safe, seems very undesirable to me (out of principle). Code in `shared/nm-utils` and `src/nm-core-utils.c` and `libnm-core/nm-utils.c` should just be thread-safe.

In my opinion, since those functions are part of the NM library of utilities and none of the NM components is multithreaded, making those functions thread-safe just adds a bit of unneeded complexity. But anyway, I don't mind too much.

Regarding the current implementation, I think the 'volatile' keyword is not needed since we are always using dedicated functions (with internal memory barriers) to access the variable.

thom311: > Regarding the current implementation, I think the 'volatile' keyword is not needed since we are 
> always using dedicated functions (with internal memory barriers) to access the variable.

You are right. To my understanding, volatile only matters when you read/write the variable. But these variables are never directly accessed. All reads/writes happens like `g_atomic_pointer_get(&variable)`, which already casts the variable's address as volatile. I think declaring the variable as volatile has not effect in practice, but it seems it bit more correct since we indeed always access the variable by taking a (volatile) pointer to it. That's why I would keep it.

shared/nm-utils/nm-shared-utils.h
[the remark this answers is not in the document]
bengal: ok, I understand now. I thought something about get_digest_check_len(), but doesn't sound so good in the end. Feel free to keep it as is.

thom311:
There are two CI failures which are unrelated to this branch. Merging to master

PR #246, opened 2018-11-06
lr/modemu-ppc64: contrib/modemu.pl fixes for IBM POWER

lkundrak:
In light of recent events, we're deprecating x86.


bengal:
LGTM

PR #247, opened 2018-11-06
rpm: add a bundled(systemd) provide

lkundrak:
Required by Fedora Packaging Standards [1] and also really useful.

[1] https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries#Requirement_if_you_bundle

### DIFF
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -216,6 +216,18 @@ BuildRequires: libubsan
 %endif
 %endif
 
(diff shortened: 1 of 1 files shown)


contrib/fedora/rpm/NetworkManager.spec
```diff
@@ -216,6 +216,17 @@ BuildRequires: libubsan
 %endif
 %endif
 
+# NetworkManager uses parts of systemd-networkd in its internal DHCP client.
```
bengal: Can you say that we use it also for IPv4LL and LLDP?
The rest LGTM.

PR #248, opened 2018-11-07
team: make teamd log to syslog

bengal:
When teamd is started in foreground it logs to stderr and its output
appears in the journal as coming from NM, which is quite confusing.

Use the new TEAM_LOG_OUTPUT environment variable [1] to tell teamd to
log through syslog, unless NetworkManager is in debug mode.

[1] https://github.com/jpirko/libteam/commit/e47d5db53873

https://bugzilla.redhat.com/show_bug.cgi?id=1596917

### DIFF
--- a/src/devices/team/nm-device-team.c
+++ b/src/devices/team/nm-device-team.c
@@ -34,6 +34,7 @@
 #include "NetworkManagerUtils.h"
 #include "devices/nm-device-private.h"
 #include "platform/nm-platform.h"
+#include "nm-config.h"
 #include "nm-core-internal.h"
 #include "nm-ip4-config.h"
 #include "nm-dbus-compat.h"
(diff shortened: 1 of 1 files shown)


src/devices/team/nm-device-team.c
```diff
@@ -479,11 +480,25 @@ teamd_child_setup (gpointer user_data)
...
 
+static const char **
+teamd_env (void)
+{
+	const char **env = g_new0 (const char *, 2);
+
+	if (nm_config_get_is_debug (nm_config_get ()))
+		env[0] = "TEAM_LOG_OUTPUT=stderr";
+	else
+		env[0] = "TEAM_LOG_OUTPUT=syslog";
```
thom311: does teamd need no environment variables at all? I wonder if there are some important once that we need to pass on as well?

bengal: It doesn't need any environment variable; indeed at the moment when NetworkManager is spawned by systemd (and teamd inherits the environment) the only variables set are PATH, LANG and the various LC_*, which are not used by teamd.

thom311: ACK

bengal:
Merged as 94ea3a5e1b14878807c6e664efe47f802707595d.

thom311:
See also: https://bugzilla.redhat.com/show_bug.cgi?id=1647414

PR #249, opened 2018-11-09
lr/ac-sort: improve sorting of active connections

lkundrak:
This aims to improve two things:

1.) Sort Hotspots higher (in "nmcli c", "nmcli d" and "nmcli")
2.) Get rid of duplicates in sorting policy


thom311:
Typo: "to want to see topost."

The order of the output seems unstable, which is the cause for the CI tests (I presume).

Could you first fix the tests?

thom311:
lgtm

PR #250, opened 2018-11-15
libnm/remote-connection: fix "flags" property getter

lkundrak:


thom311:
Could you add a "Fixes" comment? Aside that, +1.

PR #252, opened 2018-11-21
Some random OLPC/Wi-Fi fixes

lkundrak:


src/devices/wifi/nm-device-olpc-mesh.c
```diff
@@ -82,7 +82,19 @@ G_DEFINE_TYPE (NMDeviceOlpcMesh, nm_device_olpc_mesh, NM_TYPE_DEVICE)
...
+	NMDeviceOlpcMeshPrivate *priv = NM_DEVICE_OLPC_MESH_GET_PRIVATE (self);
+
+	/* We shall always have a companion if we're >= DISCONENCTED, and this
+	 * ought not be called until then. */
+	g_return_val_if_fail (priv->companion, FALSE);
+
+	/* We must not attempt to autoconnect when the companion is connected or
+	 * connecting, * because we'd tear down its connection. */
+	if (nm_device_get_state (priv->companion) > NM_DEVICE_STATE_DISCONNECTED)
+		return FALSE;
```
thom311: `NMPolicy` is the one that calls `nm_device_autoconnect_allowed()`. From the patch it's not clear, that we issue an autoconnect check at the right time (for example, when the companion disconnects). In essence, whenever the return value of `get_autoconnect_allowed()` might change, we must also trigger a new check for autoconnect.

It's also not clear, that when the system starts and the OLPC device and the companion are autoconnect devices, who wins to activate first.

Shouldn't the Wi-Fi device in turn also block autoconnect when the OLCP is active?

clients/common/nm-meta-setting-desc.c
```diff
@@ -6661,7 +6661,6 @@ static const NMMetaPropertyInfo *const property_infos_OLPC_MESH[] = {
 	PROPERTY_INFO_WITH_DESC (NM_SETTING_OLPC_MESH_SSID,
 		.is_cli_option =                TRUE,
 		.property_alias =               "ssid",
-		.inf_flags =                    NM_META_PROPERTY_INF_FLAG_REQD,
```
thom311: > For what seem to be historical reasons there's a default SSID.

where do you see a default? `verify()` in `libnm-core/nm-settings-olpc-mesh.c` seems to require a SSID.

clients/cli/devices.c
```diff
@@ -841,8 +841,8 @@ usage_device_wifi (void)
 	              "ARGUMENTS := [list [ifname <ifname>] [bssid <BSSID>] [--rescan yes|no|auto]]\n"
 	              "\n"
 	              "List available Wi-Fi access points. The 'ifname' and 'bssid' options can be\n"
-	              "used to list APs for a particular interface, or with a specific BSSID. The"
-	              "--rescan flags tells whether a new wifi scan should be triggered.\n"
+	              "used to list APs for a particular interface, or with a specific BSSID. The\n"
+	              "--rescan flag tells whether a new wifi scan should be triggered.\n"
```
thom311: > cli: fix a couple of typoes 

is the spelling "typoes" here intentional (a.k.a. a joke?).

The spelling fixes is in user-visible + translatable strings. The commit message should point that out, because it makes a difference that it is not a trivial spelling fix in a code comment (which would be far less visible to the user).

src/devices/wifi/nm-device-olpc-mesh.c
lkundrak: > `NMPolicy` is the one that calls `nm_device_autoconnect_allowed()`. From the patch it's not clear, that we issue an autoconnect check at the right time (for example, when the companion disconnects). In essence, whenever the return value of `get_autoconnect_allowed()` might change, we must also trigger a new check for autoconnect.

Makes sense. I'll look into that.

> It's also not clear, that when the system starts and the OLPC device and the companion are autoconnect devices, who wins to activate first.

That's undefined, and should be. It's about the same as if you have a hotspot and a client Wi-Fi connection at the same time. One of them wins, not clear which one.

In practice, the client connection always loses because it takes time for the scan to finish and the connection to become available.

There's one difference: the priority is not considered because it can't work across devices. Hence my remark in the commit message about the OLPC interface not being a separate device.

This is probably not a problem in practice, since when one has an autoconnect mesh connection there's use sense in also having other Wi-Fi connections autoconnectable on the same radio.

> Shouldn't the Wi-Fi device in turn also block autoconnect when the OLCP is active?

Depends. I think it sort of makes sense for a regular Wi-Fi interface to be able to preempt the mesh one (so that it acts as a fallback of sorts).

clients/common/nm-meta-setting-desc.c
lkundrak: Hmmm. `nm-device-olpc-mesh.c`:

```
        if (!nm_setting_olpc_mesh_get_ssid (s_mesh)) {
                gs_unref_bytes GBytes *ssid = NULL;

                ssid = g_bytes_new_static (DEFAULT_SSID, NM_STRLEN (DEFAULT_SSID));
                g_object_set (G_OBJECT (s_mesh), NM_SETTING_OLPC_MESH_SSID, ssid, NULL);
        }
```

Perhaps it's stupid that there's a default SSID, but I guess it used to be the case in past and perhaps there still could be connections without a SSID?

No idea. What do you think? I'm fine also with dropping these lines.

clients/cli/devices.c
lkundrak: > is the spelling "typoes" here intentional (a.k.a. a joke?).

No, I honestly thought it's spelled this way.

Will fix that and also enhance the commit message as you suggest.

clients/common/nm-meta-setting-desc.c
thom311: this code is in `complete_connection()`, which is only called during `AddAndActivate`, which in turn is called during `nmcli device connect ...`. That is fine but has no relevance for what constitutes a valid profile, because `complete_connection()` gets partial, invalid connections as input.

On the other hand, in `libnm-core/nm-setting-olpc-mesh.c` `verify()`, you see that a valid profile (currently) requires the SSID set, so nmcli also must require it. I think the patch should be dropped.

lkundrak: Yes. Dropping it.

src/devices/wifi/nm-device-olpc-mesh.c
lkundrak: > > Shouldn't the Wi-Fi device in turn also block autoconnect when the OLCP is active?
> 
> Depends. I think it sort of makes sense for a regular Wi-Fi interface to be able to preempt the mesh one (so that it acts as a fallback of sorts).

I'm actually wrong here. It does block the autoconnect by the means of a NM_DEVICE_AUTOCONNECT_ALLOWED signal handler, companion_autoconnect_allowed_cb(), that returns a value. (For that matter, I find signal handlers returning a value somehow backwards, but hey...)

lkundrak:
Branch updated to address the above.

clients/cli/devices.c
thom311: commit message seems not enhanced (yet?).

Should we also refrain from "wifi" and write "Wi-Fi" instead? Above: "... tells whether a new wifi scan ..."

src/devices/wifi/nm-device-olpc-mesh.c
```diff
@@ -262,6 +274,11 @@ companion_state_changed_cb (NMDeviceWifi *companion,
 	NMDeviceOlpcMesh *self = NM_DEVICE_OLPC_MESH (user_data);
 	NMDeviceState self_state = nm_device_get_state (NM_DEVICE (self));
 
+	if (   old_state > NM_DEVICE_STATE_DISCONNECTED
+	    && state <= NM_DEVICE_STATE_DISCONNECTED) {
+		nm_device_emit_recheck_auto_activate (NM_DEVICE (self));
+	}
```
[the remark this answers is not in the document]
lkundrak: I'm doing this when both the conditional and the body fits in one line. Otherwise it looks confusing to me.

clients/cli/devices.c
lkundrak: > commit message seems not enhanced (yet?).

It is. It says in details what the patch does. It does make no claim that it's trivial, and a negative claim "it's not trivial" doesn't seem appropriate. Most commits are not -- we mark those which positively are.

> Should we also refrain from "wifi" and write "Wi-Fi" instead? Above: "... tells whether a new wifi scan ..."

Yes.

PR #253, opened 2018-11-21
lr/dispatcher-split: split the dispatcher into an optional package

lkundrak:
This aims to improve things for packages that install dispatcher scripts; adds support for scripts in /usr and makes the whole dispatcher optional.

Packages that actually use the dispatcher (I'm in the process of compiling the list for Fedora) shall depend on NetworkManager-dispatcher or /usr/libexec/nm-dispatcher (if they want to support older NM as well).

### DIFF
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -242,6 +242,8 @@ BuildRequires: libubsan
 %endif
 %endif
 
+Provides: %{name}-dispatcher%{?_isa} = %{epoch}:%{version}-%{release}
+
 # NetworkManager uses various parts of systemd-networkd internally, including
 # DHCP client, IPv4 Link-Local address negotiation or LLDP support.
 # This provide is essentially here so that NetworkManager shows on Security
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -824,13 +824,15 @@ fi
 %{_sbindir}/%{name}
 %{_bindir}/nmcli
 %{_datadir}/bash-completion/completions/nmcli
-%dir %{_sysconfdir}/%{name}/
+%dir %{_sysconfdir}/%{name}
+%dir %{_sysconfdir}/%{name}/conf.d
 %dir %{_sysconfdir}/%{name}/dispatcher.d
 %dir %{_sysconfdir}/%{name}/dispatcher.d/pre-down.d
 %dir %{_sysconfdir}/%{name}/dispatcher.d/pre-up.d
 %dir %{_sysconfdir}/%{name}/dispatcher.d/no-wait.d
 %dir %{_sysconfdir}/%{name}/dnsmasq.d
 %dir %{_sysconfdir}/%{name}/dnsmasq-shared.d
+%dir %{_sysconfdir}/%{name}/system-connections
 %config(noreplace) %{_sysconfdir}/%{name}/NetworkManager.conf
 %{_bindir}/nm-online
 %{_libexecdir}/nm-ifup
@@ -847,8 +849,6 @@ fi
 %if %{with nmtui}
 %exclude %{_mandir}/man1/nmtui*
 %endif
-%dir %{_sysconfdir}/%{name}
-%dir %{_sysconfdir}/%{name}/conf.d
 %dir %{nmlibdir}
 %dir %{nmlibdir}/conf.d
 %dir %{nmlibdir}/VPN
@@ -857,7 +857,6 @@ fi
 %{_mandir}/man7/nmcli-examples.7*
 %{_mandir}/man8/*
 %dir %{_localstatedir}/lib/NetworkManager
-%dir %{_sysconfdir}/NetworkManager/system-connections
 %dir %{_sysconfdir}/sysconfig/network-scripts
 %{_datadir}/dbus-1/system-services/org.freedesktop.nm_dispatcher.service
 %{_datadir}/polkit-1/actions/*.policy
--- a/dispatcher/nm-dispatcher.c
+++ b/dispatcher/nm-dispatcher.c
@@ -466,12 +466,6 @@ check_permissions (struct stat *s, const char **out_error_msg)
 	g_return_val_if_fail (out_error_msg != NULL, FALSE);
 	g_return_val_if_fail (*out_error_msg == NULL, FALSE);
 
-	/* Only accept regular files */
-	if (!S_ISREG (s->st_mode)) {
-		*out_error_msg = "not a regular file.";
-		return FALSE;
-	}
-
 	/* Only accept files owned by root */
 	if (s->st_uid != 0) {
 		*out_error_msg = "not owned by root.";
@@ -573,59 +567,104 @@ dispatch_one_script (Request *request)
 	return FALSE;
 }
 
-static GSList *
-find_scripts (const char *str_action)
+static int
+_compare_basenames (gconstpointer a, gconstpointer b)
 {
-	GDir *dir;
-	const char *filename;
-	GSList *sorted = NULL;
-	GError *error = NULL;
-	const char *dirname;
+	const char *basename_a = strrchr (a, '/');
+	const char *basename_b = strrchr (b, '/');
+	int ret;
 
-	if (   strcmp (str_action, NMD_ACTION_PRE_UP) == 0
-	    || strcmp (str_action, NMD_ACTION_VPN_PRE_UP) == 0)
-		dirname = NMD_SCRIPT_DIR_PRE_UP;
-	else if (   strcmp (str_action, NMD_ACTION_PRE_DOWN) == 0
-	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
-		dirname = NMD_SCRIPT_DIR_PRE_DOWN;
-	else
-		dirname = NMD_SCRIPT_DIR_DEFAULT;
+	nm_assert (basename_a);
+	nm_assert (basename_b);
+
+	ret = strcmp (++basename_a, ++basename_b);
+	if (ret)
+		return ret;
+
+	nm_assert_not_reached ();
+	return 0;
+}
+
+static void
+_find_scripts (GHashTable *scripts, const char *base, const char *subdir)
+{
+	const char *filename;
+	gs_free char *dirname = NULL;
+	GError *error = NULL;
+	GDir *dir;
+
+	dirname = g_build_filename (base, "dispatcher.d", subdir, NULL);
 
 	if (!(dir = g_dir_open (dirname, 0, &error))) {
 		g_message ("find-scripts: Failed to open dispatcher directory '%s': %s",
 		           dirname, error->message);
 		g_error_free (error);
-		return NULL;
+		return;
 	}
 
 	while ((filename = g_dir_read_name (dir))) {
-		char *path;
-		struct stat st;
-		int err;
-		const char *err_msg = NULL;
-
 		if (!check_filename (filename))
 			continue;
 
-		path = g_build_filename (dirname, filename, NULL);
+		g_hash_table_insert (scripts,
+		                     g_strdup (filename),
+		                     g_build_filename (dirname, filename, NULL));
+	}
+
+	g_dir_close (dir);
+}
+
+static GSList *
+find_scripts (const char *str_action)
+{
+	gs_unref_hashtable GHashTable *scripts = NULL;
+	GSList *script_list = NULL;
+	GHashTableIter iter;
+	const char *subdir = NULL;
+	char *path;
+	char *filename;
+
+	if (   strcmp (str_action, NMD_ACTION_PRE_UP) == 0
+	    || strcmp (str_action, NMD_ACTION_VPN_PRE_UP) == 0)
+		subdir = "pre-up.d";
+	else if (   strcmp (str_action, NMD_ACTION_PRE_DOWN) == 0
+	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
+		subdir = "pre-down.d";
+
+	scripts = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+	_find_scripts (scripts, NMLIBDIR, subdir);
+	_find_scripts (scripts, NMCONFDIR, subdir);
+
+	g_hash_table_iter_init (&iter, scripts);
+	while (g_hash_table_iter_next (&iter, (gpointer *) &filename, (gpointer *) &path)) {
+		struct stat st;
+		char *link_target;
+		int err;
+		const char *err_msg = NULL;
+
+		link_target = g_file_read_link (path, NULL);
+		if (g_strcmp0 (link_target, "/dev/null") == 0) {
+			g_free (link_target);
+			continue;
+		}
+		g_free (link_target);
 
 		err = stat (path, &st);
 		if (err)
 			g_warning ("find-scripts: Failed to stat '%s': %d", path, err);
-		else if (S_ISDIR (st.st_mode))
+		else if (!S_ISREG (st.st_mode))
 			; /* silently skip. */
 		else if (!check_permissions (&st, &err_msg))
 			g_warning ("find-scripts: Cannot execute '%s': %s", path, err_msg);
 		else {
 			/* success */
-			sorted = g_slist_insert_sorted (sorted, path, (GCompareFunc) g_strcmp0);
-			path = NULL;
+			script_list = g_slist_prepend (script_list, g_strdup (path));
+			continue;
 		}
-		g_free (path);
 	}
-	g_dir_close (dir);
 
-	return sorted;
+	return g_slist_sort (script_list, _compare_basenames);
 }
 
 static gboolean
@@ -635,6 +674,7 @@ script_must_wait (const char *path)
 	gs_free char *dir = NULL;
 	gs_free char *real = NULL;
 	char *tmp;
+
 
 	link = g_file_read_link (path, NULL);
 	if (link) {
@@ -648,8 +688,7 @@ script_must_wait (const char *path)
 
 		dir = g_path_get_dirname (link);
 		real = realpath (dir, NULL);
-
-		if (real && !strcmp (real, NMD_SCRIPT_DIR_NO_WAIT))
+		if (real && !g_str_has_suffix (real, "/no-wait.d"))
 			return FALSE;
 	}
 
--- a/src/nm-dispatcher.c
+++ b/src/nm-dispatcher.c
@@ -15,7 +15,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  *
- * Copyright (C) 2004 - 2017 Red Hat, Inc.
+ * Copyright (C) 2004 - 2018 Red Hat, Inc.
  * Copyright (C) 2005 - 2008 Novell, Inc.
  */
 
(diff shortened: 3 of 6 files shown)


lkundrak:
Just for the record; these are the packages I need to fix up when we add support for scripts in /usr:

```
$ curl http://download.eng.brq.redhat.com/pub/fedora/linux//development/rawhide/Everything/x86_64/os/repodata/b75083ece5738cb920dbada5395e0030f89bc390145760fc9a5a9c0159bca9fe-filelists.xml.gz |zcat >filelist.xml
$ tidy -w 65535 -mi -xml -raw filelist.xml
$ awk -F\" '/name=/ {n=$4} /NetworkManager\/dispatcher\.d/ {print n"\t"$0}' filelist.xml
NetworkManager      <file type="dir">/etc/NetworkManager/dispatcher.d</file>
NetworkManager      <file type="dir">/etc/NetworkManager/dispatcher.d/no-wait.d</file>
NetworkManager      <file type="dir">/etc/NetworkManager/dispatcher.d/pre-down.d</file>
NetworkManager      <file type="dir">/etc/NetworkManager/dispatcher.d/pre-up.d</file>
NetworkManager-dispatcher-routing-rules     <file>/etc/NetworkManager/dispatcher.d/10-ifcfg-rh-routes.sh</file>
NetworkManager-dispatcher-routing-rules     <file>/etc/NetworkManager/dispatcher.d/no-wait.d/10-ifcfg-rh-routes.sh</file>
NetworkManager-dispatcher-routing-rules     <file>/etc/NetworkManager/dispatcher.d/pre-up.d/10-ifcfg-rh-routes.sh</file>
chrony      <file>/etc/NetworkManager/dispatcher.d/20-chrony</file>
clatd       <file type="dir">/etc/NetworkManager/dispatcher.d</file>
clatd       <file>/etc/NetworkManager/dispatcher.d/50-clatd</file>
cloud-init          <file>/etc/NetworkManager/dispatcher.d/hook-network-manager</file>
ddclient            <file>/etc/NetworkManager/dispatcher.d/50-ddclient</file>
dhcp-client         <file type="dir">/etc/NetworkManager/dispatcher.d</file>
dhcp-client         <file>/etc/NetworkManager/dispatcher.d/11-dhclient</file>
dhcp-server         <file type="dir">/etc/NetworkManager/dispatcher.d</file>
dhcp-server         <file>/etc/NetworkManager/dispatcher.d/12-dhcpd</file>
dnssec-trigger      <file>/etc/NetworkManager/dispatcher.d/01-dnssec-trigger</file>
elementary-capnet-assist            <file>/etc/NetworkManager/dispatcher.d/90captive_portal_test</file>
inadyn-mt           <file>/etc/NetworkManager/dispatcher.d/30-inadyn</file>
iscsi-initiator-utils       <file>/etc/NetworkManager/dispatcher.d/04-iscsi</file>
iscsi-initiator-utils       <file>/etc/NetworkManager/dispatcher.d/04-iscsi</file>
modem-manager-gui-cm-NetworkManager         <file>/etc/NetworkManager/dispatcher.d/95-mmgui-timestamp-notifier</file>
olpc-update         <file>/etc/NetworkManager/dispatcher.d/olpc-update-ifup</file>
puppet      <file type="dir">/etc/NetworkManager/dispatcher.d</file>
puppet      <file>/etc/NetworkManager/dispatcher.d/98-puppet</file>
samba-winbind       <file>/etc/NetworkManager/dispatcher.d/30-winbind</file>
sendmail            <file>/etc/NetworkManager/dispatcher.d/10-sendmail</file>
shorewall-init      <file>/etc/NetworkManager/dispatcher.d/01-shorewall</file>
squid       <file>/etc/NetworkManager/dispatcher.d/20-squid</file>
tlp-rdw     <file>/etc/NetworkManager/dispatcher.d/99tlp-rdw-nm</file>
yum-NetworkManager-dispatcher       <file>/etc/NetworkManager/dispatcher.d/yum-NetworkManager-dispatcher</file>
```

tools/meson-post-install.sh
```diff
@@ -27,6 +27,9 @@ for dir in "${pkgconfdir}/conf.d" \
            "${pkgconfdir}/dnsmasq.d" \
            "${pkgconfdir}/dnsmasq-shared.d" \
            "${pkglibdir}/conf.d" \
+           "${pkglibdir}/dispatcher.d/no-wait.d" \
+           "${pkglibdir}/dispatcher.d/pre-down.d" \
+           "${pkglibdir}/dispatcher.d/pre-up.d" \
```
thom311: could you reorder this commit to be last, so that the intermediate commits of the branch still make sense?

src/nm-dispatcher.c
```diff
@@ -15,7 +15,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  *
- * Copyright (C) 2004 - 2017 Red Hat, Inc.
+ * Copyright (C) 2004 - 2018 Red Hat, Inc.
```
thom311: Commit-message: "Follow up patches make dispatcher optional."

the dispatcher was always optional, you were always welcome to mask the service. The commit message makes it sound that wouldn't be the case.

thom311: The claim that the fast-path is not hit in usual scenarios does not seem correct. The monitors are also for `pre-up.d` and `pre-down.d` directories, and these are quite often empty. Especially `pre-down.d` is always empty, and AFAIK, only `NetworkManager-dispatcher-routing-rules` package installs a script in `pre-up.d` (and this is an optional package, only on RHEL/Fedora).

The patch saves 150 LOC. Sure, and extra D-Bus call just to figure out that there is nothing to do is relatively fast. I say, "relatively", because NM is competing with networkd, which can configure an interface in a matter of milliseconds. How long does it really take? You need to context-switch to dbus-daemon, NM-dispatcher, dbus-daemon and back.

dispatcher/nm-dispatcher.c
```diff
@@ -627,6 +619,26 @@ find_scripts (const char *str_action)
...
+
+	if (   strcmp (str_action, NMD_ACTION_PRE_UP) == 0
+	    || strcmp (str_action, NMD_ACTION_VPN_PRE_UP) == 0)
+		subdir = "pre-up.d";
+	else if (   strcmp (str_action, NMD_ACTION_PRE_DOWN) == 0
+	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
+		subdir = "pre-down.d";
+
+	sorted = _find_scripts (sorted, NMCONFDIR, subdir);
+	sorted = _find_scripts (sorted, NMLIBDIR, subdir);
```
thom311: while at it, why not also check `/run/NetworkManager/dispatcher.d`? It seems to add no extra effort.

dispatcher/nm-dispatcher.c
```diff
@@ -627,6 +619,26 @@ find_scripts (const char *str_action)
 	}
 	g_dir_close (dir);
 
+out:
+	return sorted;
```
thom311: above `g_slist_insert_sorted` is wrong. Now that you the files are in different paths, you need to sort based on the base-name.

A simple fix would be to implement a compare function that looks into the full path and compares the base-names.

Instead of incrementally constructing the list, it seems better to just construct it via `g_slist_prepend()` and do one final g_slist_sort().

dispatcher/nm-dispatcher.c
```diff
@@ -650,8 +664,9 @@ script_must_wait (const char *path)
 
 		dir = g_path_get_dirname (link);
 		real = realpath (dir, NULL);
-
-		if (real && !strcmp (real, NMD_SCRIPT_DIR_NO_WAIT))
+		if (real)
+			basename = g_path_get_basename (real);
+		if (basename && !strcmp (basename, "no-wait.d"))
```
thom311: I think this should be just `if (real && g_str_has_suffix (real, "/no-wait.d"))`

contrib/fedora/rpm/NetworkManager.spec
```diff
@@ -237,6 +237,16 @@ Ethernet, Bridge, Bond, VLAN, Team, InfiniBand, Wi-Fi, mobile broadband
...
 
 
+%package dispatcher
+Summary: NetworkManager dispatcher
+Group: System Environment/Base
+Requires: %{name}%{?_isa} = %{epoch}:%{version}-%{release}
+
+%description dispatcher
+This package contains NetworkManager dispatcher. It provides support for
+running custom scripts upon various events signalled by NetworkManager.
```
thom311: I am not against this patch. But it's not sufficiently clear to me whether this package split is a good idea or necessary.

Could you describe the advantages in the commit message more in depth? And why you think this is a good idea.

Note, that the dispatcher (systemd) service already can be masked. I think, `systemctl disable NetworkManager-dispatcher.service` should also suffice, as that would remove the D-Bus socket unit (didn't test that though).

thom311: This is essentially a package-split. I think you need to make proper use of `Obsoletes` and `Provides`.

tools/meson-post-install.sh
lkundrak: I think would break the RPM build which needs the directories.

dispatcher/nm-dispatcher.c
lkundrak: I'm struggling to figure out an use for it that wouldn't be sort of stupid. I'd prefer if someone actually needed it first.

thom311: OK, makes sense.

src/nm-dispatcher.c
lkundrak: > Commit-message: "Follow up patches make dispatcher optional."
> 
> the dispatcher was always optional, you were always welcome to mask the service. The commit message makes it sound that wouldn't be the case.

Okay. Dropped that part.

> The claim that the fast-path is not hit in usual scenarios does not seem correct. The monitors are also for `pre-up.d` and `pre-down.d` directories, and these are quite often empty. Especially `pre-down.d` is always empty, and AFAIK, only `NetworkManager-dispatcher-routing-rules` package installs a script in `pre-up.d` (and this is an optional package, only on RHEL/Fedora).
> 
> The patch saves 150 LOC. Sure, and extra D-Bus call just to figure out that there is nothing to do is relatively fast. I say, "relatively", because NM is competing with networkd, which can configure an interface in a matter of milliseconds. How long does it really take? You need to context-switch to dbus-daemon, NM-dispatcher, dbus-daemon and back.

Ah, I didn't really think of that. You're right.

A quick look at the Bustle graph shows 12ms first (activation i suppose) and then under a milisecond for subsequent calls. I guess that's okay?
lala

dispatcher/nm-dispatcher.c
lkundrak: Okay. Fixing...

contrib/fedora/rpm/NetworkManager.spec
lkundrak: I've now replaced the patch with one that merely adds a Provide that packages that install dispatcher scripts can depend on.

I intend to submit it some later point, taking your comments into account.

dispatcher/nm-dispatcher.c
```diff
@@ -575,29 +575,38 @@ dispatch_one_script (Request *request)
...
+	int ret;
+
+	g_return_val_if_fail (basename_a, 0);
+	g_return_val_if_fail (basename_b, 0);
+
+	ret = strcmp (++basename_a, ++basename_b);
+	if (ret)
+		return ret;
+
+	return strcmp (a, b);
```
thom311: ok, `basename_a` and `basename_b` are the same.

That means, `/etc/NetworkManager/.../$basename_a` shadows `/usr/lib/NetworkManager/.../$basename_a`. Just comparing them with `strcmp()` does not correctly prefer `/etc` or `/usr`. Ok, the final final `strcmp()` doesn't really matter here, because there are larger issues. You not only need to sort such two entries, you need to skip the shadowed one from running! But preferably, not without a debug message in the log.


Also, `_find_scripts()` only return scripts that are executable. Wouldn't it be nice, if the user could prevent `/usr/lib/NetworkManager/dispatcher.d/$SCRIPT` from running, by putting a non-executable file in `/etc` (or a symlink to `/dev/null`)? That means, `_find_scripts()` must return all paths in the first moment, and collect the information about whether it's executable or not, and use that to figure out which scripts are shadowed.

thom311: maybe this is the point, where we need unit-tests for this code...

thom311: I think you not only track the filenames, but also (at least) `struct stat` for the permission check. If you happen to do that, I think we should not use `GSList`. That is just the worst data structure.

src/nm-dispatcher.c
bengal: > I don't think we need to optimize this further for simplicity, because IMO it's self-contained and simple already.

I agree, the complexity of monitoring dispatcher files doesn't seem too much to me, and offers some (small) performance improvements.

lkundrak: This is an important part though:

> Let's remove them now, making it easier to modify the dispatcher to look for scripts in other places.

We'd need more watches and we'd also need the daemon to know about them, which adds complexity to the daemon I'd like to avoid.

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
 		if (!check_filename (filename))
 			continue;
 
-		path = g_build_filename (dirname, filename, NULL);
+		link_target = g_file_read_link (path, NULL);
+		if (g_strcmp0 (link_target, "/dev/null") == 0) {
+			g_free (link_target);
+			continue;
+		}
+		g_free (link_target);
```
thom311: `gs_free char *link_target = NULL;`

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
+	    || strcmp (str_action, NMD_ACTION_VPN_PRE_UP) == 0)
+		subdir = "pre-up.d";
+	else if (   strcmp (str_action, NMD_ACTION_PRE_DOWN) == 0
+	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
+		subdir = "pre-down.d";
+
+	scripts = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+	_find_scripts (scripts, NMLIBDIR, subdir);
+	_find_scripts (scripts, NMCONFDIR, subdir);
```
thom311: ok, this puts all files into the scripts hash, shadowing files according to the same filename. But only below, it checks for the file permissions. I think it's correct that a non-executable file (owned by root) shadows an executable script. But if the shadowing file is a directory or owned by the wrong user or writable to non-root, I think it should not shadow a script. Or at least, it's a question that should be considered before agreeing that shadowing only happens based on the filename alone, without regard of owner/permissions/type.

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
+
+	g_hash_table_iter_init (&iter, scripts);
+	while (g_hash_table_iter_next (&iter, (gpointer *)&filename, (gpointer *)&path)) {
 		struct stat st;
+		char *link_target;
 		int err;
 		const char *err_msg = NULL;
 
 		if (!check_filename (filename))
```
thom311: `check_filename()` should be done by `_find_scripts()`. If the filename isn't valid, it can be rejected right away. Then you also don't need the "filename" variable here.

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
+	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
+		subdir = "pre-down.d";
+
+	scripts = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+	_find_scripts (scripts, NMLIBDIR, subdir);
+	_find_scripts (scripts, NMCONFDIR, subdir);
+
+	g_hash_table_iter_init (&iter, scripts);
+	while (g_hash_table_iter_next (&iter, (gpointer *)&filename, (gpointer *)&path)) {
```
thom311: space between cast and variable `(gpointer *) &path`.

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
-find_scripts (const char *str_action)
+static int
+_compare_basenames (gconstpointer a, gconstpointer b)
+{
+	const char *basename_a = strrchr (a, '/');
+	const char *basename_b = strrchr (b, '/');
+	int ret;
+
+	g_return_val_if_fail (basename_a, 0);
+	g_return_val_if_fail (basename_b, 0);
```
thom311: it's easy to see that this won't fail. Please use `nm_assert()`.

Also, the first `strcmp()` is not supposed to return equality. Add another `nm_assert_not_reached (); return 0;` instead of `strcmp (a, b);`.

dispatcher/nm-dispatcher.c
```diff
@@ -620,14 +666,13 @@ find_scripts (const char *str_action)
 			g_warning ("find-scripts: Cannot execute '%s': %s", path, err_msg);
```
thom311: I think we should also skip files with size of 0. We allow such empty files to shadow another file, but it shouldn't actually be run.

lkundrak: Systemd shadows things with /dev/null symlinks. I think that we should do as well, but without a warning.

dispatcher/nm-dispatcher.c
lkundrak: Ok.

dispatcher/nm-dispatcher.c
```diff
@@ -575,41 +575,87 @@ dispatch_one_script (Request *request)
...
+	char *filename;
+
+	if (   strcmp (str_action, NMD_ACTION_PRE_UP) == 0
+	    || strcmp (str_action, NMD_ACTION_VPN_PRE_UP) == 0)
+		subdir = "pre-up.d";
+	else if (   strcmp (str_action, NMD_ACTION_PRE_DOWN) == 0
+	         || strcmp (str_action, NMD_ACTION_VPN_PRE_DOWN) == 0)
+		subdir = "pre-down.d";
+
+	scripts = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
```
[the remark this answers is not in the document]
lkundrak: Ok

dispatcher/nm-dispatcher.c
lkundrak: I can't see why anything but file name would decide whether something shadows something else. It just takes extra effort for no benefit.

Also the permission check is just an extra safety net. Nobody but root can actually install the scripts anyway.

dispatcher/nm-dispatcher.c
lkundrak: Ok.

dispatcher/nm-dispatcher.c
lkundrak: Good point.

dispatcher/nm-dispatcher.c
lkundrak: Okay.

lkundrak:
Updated the branch.

vbenes:
https://faf.lab.eng.brq.redhat.com/faf//reports/10981/ crash

vbenes:
https://faf.lab.eng.brq.redhat.com/faf//reports/10982/
https://faf.lab.eng.brq.redhat.com/faf//reports/10983/
https://faf.lab.eng.brq.redhat.com/faf//reports/10984/

not sure how these differ

dispatcher/nm-dispatcher.c
```diff
@@ -573,59 +567,105 @@ dispatch_one_script (Request *request)
...
-
-		path = g_build_filename (dirname, filename, NULL);
+		}
+		g_free (link_target);
 
 		err = stat (path, &st);
 		if (err)
 			g_warning ("find-scripts: Failed to stat '%s': %d", path, err);
-		else if (S_ISDIR (st.st_mode))
+		else if (S_ISREG (st.st_mode))
```
thom311: this looks wrong

lkundrak:
> could you comment whether the CI crashes are related or what's the plan about them?

fixed

lkundrak:
- [ ] https://src.fedoraproject.org/rpms/chrony/pull-request/2
- [ ] https://src.fedoraproject.org/rpms/clatd/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/cloud-init/pull-request/4
- [x] https://src.fedoraproject.org/rpms/ddclient/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/dhcp/pull-request/4
- [ ] https://src.fedoraproject.org/rpms/dnssec-trigger/pull-request/3
- [ ] https://src.fedoraproject.org/rpms/elementary-capnet-assist/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/inadyn-mt/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/iscsi-initiator-utils/pull-request/6
- [ ] https://src.fedoraproject.org/rpms/modem-manager-gui/pull-request/1
- [x] https://src.fedoraproject.org/rpms/olpc-update/pull-request/2
- [ ] https://src.fedoraproject.org/rpms/puppet/pull-request/7
- [ ] https://src.fedoraproject.org/rpms/samba/pull-request/10
- [ ] https://src.fedoraproject.org/rpms/sendmail/pull-request/2
- [ ] https://src.fedoraproject.org/rpms/shorewall/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/squid/pull-request/1
- [ ] https://src.fedoraproject.org/rpms/tlp/pull-request/1

PR #254, opened 2018-11-23
cli: avoid crash on device disconnect

bengal:
https://bugzilla.redhat.com/show_bug.cgi?id=1546061

### DIFF
--- a/clients/cli/devices.c
+++ b/clients/cli/devices.c
@@ -2064,6 +2064,7 @@ typedef struct {
 	GSList *queue;
 	guint timeout_id;
 	gboolean cmd_disconnect;
+	GCancellable *cancellable;
 } DeviceCbInfo;
 
 static void device_cb_info_finish (DeviceCbInfo *info, NMDevice *device);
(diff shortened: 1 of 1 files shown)


thom311:
>   exits the main loop and then the callback for disconnect_async() is
>   executed anyway because it was already dispatched, leading to an
>   invalid memory access.

I don't think that's accurate. In this case, the callbacks `disconnect_state_cb()` register to "notify::state" of the device. Such property changed signal handlers are invoked synchronously, right NMDevice changes the property.
There is never a race, where we would register a handler, then the handler gets queued (in the background), we unregister the signal handler, and still get called afterwards.

I think a better fix would be to disconnect the signal handler when we no longer want to be called back.

thom311:
I see. I think the better solution is a GCancellable and checking `nm_utils_error_is_cancelled()` / `g_error_matches()` in `disconnect_device_cb()`.

clients/cli/devices.c
```diff
@@ -2366,9 +2347,6 @@ do_devices_disconnect (NmCli *nmc, int argc, char **argv)
 	if (nmc->timeout > 0)
 		info->timeout_id = g_timeout_add_seconds (nmc->timeout, device_op_timeout_cb, info);
 
-	g_signal_connect (nmc->client, NM_CLIENT_DEVICE_REMOVED,
-	                  G_CALLBACK (device_removed_cb), info);
```
lkundrak: I think this was added precisely because the device could disappear without switching a state to DISCONNECTED. Are you sure this cannot happen?

PR #257, opened 2018-12-05
settings: apply agent-owned secrets from the new connection on update

bengal:
When updating a connection passing agent-owned secret, they are lost
from @reread_connection after the settings-plugin persists the
connection. Therefore we need to cache and reapply them separately to
the connection so that they can be saved to secret agents later.

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/82


bengal:
Applied: a3383726dce09255b5413a6230be580856ab212c

PR #258, opened 2018-12-06
[bg/sriov-fixes] Some fixes related to SR-IOV

bengal:


libnm-core/nm-setting-sriov.c
```diff
@@ -33,6 +33,7 @@
  */
 struct _NMSettingSriov {
 	NMSetting parent;
+	gboolean enable;
```
thom311: could we reorder the fields so that they fill the gaps of the alignment? Basically, fields of the same type should follow each other. Also, inside a struct, it seems better to have `bool :1` bitfields. Not that it makes a difference in this case (in terms of size of the struct), but as a rule.

libnm-core/nm-connection.c
```diff
@@ -1253,15 +1253,37 @@ _normalize_ip_tunnel_wired_setting (NMConnection *self, GHashTable *parameters)
...
+			              NM_SETTING_SRIOV_ENABLE, TRUE,
+			              NULL);
+			ret = TRUE;
+		}
+	}
+
 
-	return _nm_setting_sriov_sort_vfs (s_sriov);
+	/* 2. remove the setting if everything is disabled */
+	if (!nm_setting_sriov_get_enable (s_sriov)) {
```
thom311: setting's `verify()`  must mirror the normalization steps, which is not the case for this normalization.

Also, it seems questionable, to have a property whose only value is always enforced via normalization. If you don't want to keep a sriov setting that is disabled, then there should not be a new sriov.enable property but the user should remove the SRIOV section entirely. Maybe that's anyway a better solution.

src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-writer.c
```diff
@@ -2220,18 +2220,22 @@ write_sriov_setting (NMConnection *connection, shvarFile *ifcfg)
...
+	gboolean enable = FALSE;
 
 	svUnsetAll (ifcfg, SV_KEY_TYPE_SRIOV_VF);
 
 	s_sriov = NM_SETTING_SRIOV (nm_connection_get_setting (connection, NM_TYPE_SETTING_SRIOV));
 	if (s_sriov)
-		num = nm_setting_sriov_get_total_vfs (s_sriov);
-	if (num == 0) {
+		enable = nm_setting_sriov_get_enable (s_sriov);
+	if (!enable) {
```
thom311: the reader/writer should be able to persist a setting without loss (full round-trip). Ok, SRIOV's `verify()` enforces that if the setting is not enabled, that the setting isn't present (somewhat). So, currently it doesn't make a difference. But in general, the ifcfg code should not have additional conditions on top of what NMSettingSriov already has. In the best case, it checks impossible conditions, in the worst case, the writer does not persist every possible setup. Less logic here, just plain setting the fields.

src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-reader.c
```diff
@@ -2275,12 +2274,15 @@ make_sriov_setting (shvarFile *ifcfg)
 		}
 	}
 
-	s_sriov = (NMSettingSriov *) nm_setting_sriov_new ();
-	g_object_set (s_sriov,
-	              NM_SETTING_SRIOV_TOTAL_VFS, total_vfs,
-	              NM_SETTING_SRIOV_VFS, vfs,
-	              NM_SETTING_SRIOV_AUTOPROBE_DRIVERS, (int) autoprobe_drivers,
-	              NULL);
+	if (enable || total_vfs || vfs) {
```
thom311: same comment as for the writer below.

libnm-core/nm-setting-sriov.c
bengal: ok

libnm-core/nm-connection.c
bengal: The problem with removing the sriov setting entirely is that it is not possible to do it from nmcli (is it?), where users can only control single properties.

src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-writer.c
bengal: This should be fixed now.

src/settings/plugins/ifcfg-rh/nms-ifcfg-rh-reader.c
bengal: For the reader I think we need a bit of logic to decide whether to instantiate a sriov setting or not.

libnm-core/nm-connection.c
thom311: there is with `nmcli connection edit`. If it isn't with command line, then that is a missing feature (of not very high priority).

src/devices/nm-device.c
```diff
@@ -4227,7 +4227,7 @@ device_init_static_sriov_num_vfs (NMDevice *self)
 		num_vfs = _nm_utils_ascii_str_to_int64 (value, 10, 0, G_MAXINT32, -1);
 		if (num_vfs >= 0) {
 			nm_platform_link_set_sriov_params (nm_device_get_platform (self),
-			                                   priv->ifindex, num_vfs, -1);
+			                                   priv->ifindex, num_vfs, NM_TERNARY_DEFAULT);
```
thom311: in the commit message, there is a "Fixes" comment, but it's not clear what the problem with the previous implementation is? Is there an actual bug? Then the commit message should highlight it (I didn't see it).

If it's only a code improvement (of which I am in favour), then the Fixes comment seems a bit too strong. Generally, I really like Fixes if they actually fix a problem. But here?

src/platform/nm-linux-platform.c
```diff
@@ -5782,7 +5782,7 @@ link_set_sriov_params (NMPlatform *platform,
 	                                                                                  "device/sriov_drivers_autoprobe"),
 	                                                        10, 0, G_MAXUINT, 0);
 	if (   current_num == num_vfs
-	    && (autoprobe == -1 || current_autoprobe == autoprobe))
+	    && (autoprobe == NM_TERNARY_DEFAULT || current_autoprobe == autoprobe))
```
thom311: I think, `nm_platform_sysctl_get_int_checked` above should only return 0, 1, and -1 (on error). Especially return `0` on error (above), is wrong.

bengal: Fixed, thanks.

bengal:
Merged to master as b6259f70564cb99e4ee6bf4b49a427898caf7328.

PR #259, opened 2018-12-11
wwan: rework when settings/device are blocked for autoconnection

aleksander0m:
The reasons to block autoconnection at settings level are not the same
as the ones to block autoconnection at device level.

E.g. if the SIM-PIN is wrong, you may want to block autoconnection
both at settings level (as the PIN configured in settings is wrong)
and at device level (so that no other setting is tried automatically).

For some other reasons, you may want to block autoconnection only at
setting level (e.g. wrong APN).

And for some other reasons you may want to block autoconnection at
device level only (e.g. SIM missing), so that the autoconnection
blocking is removed when the device goes away. This is especially
important with SIM hotplug events processed by ModemManager, as a
device without SIM will be removed from MM when a new SIM is
inserted, so that a completely new object is exposed in MM with the
newly detected SIM.


vhumpa:

thom311:
ok to test

fgiudici:
lgtm

thom311:
merged to master and nm-1-14.

Thanks!!

PR #260, opened 2018-12-13
devices,bluetooth: fix default CDMA number setting

aleksander0m:


vhumpa:

thom311:
merged.

Thank you!!


For the record, the bug has no actual consequences, because the defines are both `"number"`.

aleksander0m:
The defines are both "number", and if unset, MM will anyway default to "#777" so totally no consequences :) it just felt odd.

PR #261, opened 2018-12-13
settings,gsm: deprecate and stop using 'number' property

aleksander0m:
The 'number' property in GSM settings is a legacy thing that comes
from when ModemManager used user-provided numbers, if any, to connect
3GPP modems.

Since ModemManager 1.0, this property is completely unused for 3GPP
modems, and so it doesn't make sense to use it in the NetworkManager
settings. Ofono does not use it either.

For AT+PPP-based 3GPP modems, the 'number' to call to establish the
data connection is decided by ModemManager itself, e.g. for standard
GSM/UMTS/LTE modems it will connect a given predefined PDP context,
and for other modems like Iridium it will have the number to call
hardcoded in the plugin itself.

### DIFF
--- a/libnm-core/nm-setting-gsm.c
+++ b/libnm-core/nm-setting-gsm.c
@@ -100,6 +100,8 @@ nm_setting_gsm_new (void)
  * @setting: the #NMSettingGsm
  *
  * Returns: the #NMSettingGsm:number property of the setting
+ *
+ * Deprecated: 1.16:  user-provided values for this setting are no longer used.
  **/
 const char *
 nm_setting_gsm_get_number (NMSettingGsm *setting)
@@ -621,10 +623,10 @@ nm_setting_gsm_class_init (NMSettingGsmClass *klass)
 	/**
 	 * NMSettingGsm:number:
 	 *
-	 * Number to dial when establishing a PPP data session with the GSM-based
-	 * mobile broadband network.  Many modems do not require PPP for connections
-	 * to the mobile network and thus this property should be left blank, which
-	 * allows NetworkManager to select the appropriate settings automatically.
+	 * Legacy setting that used to help establishing PPP data sessions for
+	 * GSM-based modems.
+	 *
+	 * Deprecated: 1.16: user-provided values for this setting are no longer used.
 	 **/
 	g_object_class_install_property
 		(object_class, PROP_NUMBER,
--- a/src/devices/wwan/nm-modem-broadband.c
+++ b/src/devices/wwan/nm-modem-broadband.c
@@ -279,11 +279,6 @@ create_gsm_connect_properties (NMConnection *connection)
 	setting = nm_connection_get_setting_gsm (connection);
 	properties = mm_simple_connect_properties_new ();
 
-	/* TODO: not needed */
-	str = nm_setting_gsm_get_number (setting);
-	if (str)
-		mm_simple_connect_properties_set_number (properties, str);
-
 	/* Blank APN ("") means the default subscription APN */
 	str = nm_setting_gsm_get_apn (setting);
 	mm_simple_connect_properties_set_apn (properties, str ?: "");
@@ -692,10 +687,6 @@ complete_connection (NMModem *_self,
 			g_prefix_error (error, "%s: ", NM_SETTING_GSM_SETTING_NAME);
 			return FALSE;
 		}
-
-		/* TODO: This is not needed */
-		if (!nm_setting_gsm_get_number (s_gsm))
-			g_object_set (G_OBJECT (s_gsm), NM_SETTING_GSM_NUMBER, "*99#", NULL);
 
 		nm_utils_complete_generic (NM_PLATFORM_GET,
 		                           connection,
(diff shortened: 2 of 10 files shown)


vhumpa:

libnm-core/nm-setting-gsm.c
```diff
@@ -621,10 +614,10 @@ nm_setting_gsm_class_init (NMSettingGsmClass *klass)
...
 	 * NMSettingGsm:number:
 	 *
-	 * Number to dial when establishing a PPP data session with the GSM-based
-	 * mobile broadband network.  Many modems do not require PPP for connections
-	 * to the mobile network and thus this property should be left blank, which
-	 * allows NetworkManager to select the appropriate settings automatically.
+	 * Legacy setting that used to help establishing PPP data sessions for
+	 * GSM-based modems.
+	 *
+	 * Deprecated: 1.16: user-provided values for this setting are no longer used..
```
thom311: Double ".." at the end of sentence.

---

If you modify any of the gtk-doc settings, you need to regenerate `clients/common/settings-docs.h.in`. You'll also see that `make check` fails (if you build with gtk-doc). To fix it, `./configure` with `--enable-gtk-doc` and then run

```
   NM_TEST_REGENERATE=1 make -j 10 check
```

---

For other reasons, you also need to regenerate `clients/tests/test-client.check-on-disk/test_003.expected`. Doing that is a bit more tricky. See howto at [test-client.py](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/blob/90e9695af541d0e13c5327466f1414b35a04c73d/clients/tests/test-client.py#L14). Optimally, the howto there is sufficiently clear. If you struggle understanding what you need to so, please ask and we see how the howto can be improved.

src/devices/wwan/nm-modem-broadband.c
```diff
@@ -693,10 +693,6 @@ complete_connection (NMModem *_self,
 			return FALSE;
 		}
 
-		/* TODO: This is not needed */
```
thom311: this TODO was added more than 6 years ago :)

libnm-core/nm-setting-gsm.c
thom311: For some reasons, this "Deprecated" line confuses `tools/check-settings-docs.sh`. I wonder whether the comment should be adjusted, or the script improved...

aleksander0m: I tried to follow gtk-doc style to write the deprecation comment, not sure if that's how you usually do it.

aleksander0m: Does this ring any bell? This is when I run `NM_TEST_REGENERATE=1 make check-local-clients-tests-test-client`

NM_TEST_CLIENT_NMCLI_PATH=clients/cli/nmcli \
"/usr/bin/python" \
./clients/tests/test-client.py -v &> "./clients/tests/test-client.log" && r=ok; \
cat "./clients/tests/test-client.log"; \
test "$r" == ok
test_001 (__main__.TestNmcli) ... ERROR
test_002 (__main__.TestNmcli) ... ERROR
test_003 (__main__.TestNmcli) ... ERROR
test_004 (__main__.TestNmcli) ... ERROR

======================================================================
ERROR: test_001 (__main__.TestNmcli)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./clients/tests/test-client.py", line 784, in f
    self._nm_test_post()
  File "./clients/tests/test-client.py", line 766, in _nm_test_post
    content_new = ''.join([r['content'] for r in results])
TypeError: sequence item 0: expected str instance, bytes found

======================================================================
ERROR: test_002 (__main__.TestNmcli)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./clients/tests/test-client.py", line 784, in f
    self._nm_test_post()
  File "./clients/tests/test-client.py", line 766, in _nm_test_post
    content_new = ''.join([r['content'] for r in results])
TypeError: sequence item 0: expected str instance, bytes found

======================================================================
ERROR: test_003 (__main__.TestNmcli)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./clients/tests/test-client.py", line 784, in f
    self._nm_test_post()
  File "./clients/tests/test-client.py", line 766, in _nm_test_post
    content_new = ''.join([r['content'] for r in results])
TypeError: sequence item 0: expected str instance, bytes found

======================================================================
ERROR: test_004 (__main__.TestNmcli)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./clients/tests/test-client.py", line 784, in f
    self._nm_test_post()
  File "./clients/tests/test-client.py", line 766, in _nm_test_post
    content_new = ''.join([r['content'] for r in results])
TypeError: sequence item 0: expected str instance, bytes found

----------------------------------------------------------------------
Ran 4 tests in 7.119s

aleksander0m: Wait, I know what that is. I'm using Arch Linux, where /usr/bin/python is really python3. After reconfiguring with PYTHON=/usr/bin/python2 explicitly, it worked.

aleksander0m: I believe I did the changes now

thom311: that's a bit odd, because the test is supposed to work equally well with python2 and pyhon3. I am pretty sure we also test both versions (successfully).

thom311:
Merged to master. Thanks!!

PR #263, opened 2018-12-19
[th/dhcp-client-id-infiniband-rh1658057]

thom311:
See https://bugzilla.redhat.com/show_bug.cgi?id=1658057#c26 and following.


bengal:
LGTM

PR #264, opened 2018-12-21
device: ensure IP configuration is restored when link goes up

bengal:
When the link is up and goes down link_changed_cb() schedules
device_link_changed() to be run later. If the function is dispatched
when the link is already up again, it does not detect that the link
was down.

Fix this by storing in the device state that we saw the link down so
that device_link_changed() can properly restore the IP configuration.

https://bugzilla.redhat.com/show_bug.cgi?id=1636715


thom311:
lgtm

bengal:
Merged: 7bd193ef30fb2a1d4c328bb61412e35a5bccfced

PR #266, opened 2018-12-21
fg/assume bridge rh1593939

fgiudici:
When generating an in-memory connection from a bridge device configured outside NetworkManager we could fail due to STP related parameters having different values. This is true also when STP is disabled, so we may end up not taking over a device that would perfectly match a connection profile.
Note also that when STP is disabled, we reset STP related settings to their default.
Do the same when generating an in-memory connection from a configured device.

A special mention is required for the _cost_ value of a bridge port: we use '100' as default, which is different from the kernel one (which is dynamic). Skip the _path-cost_ property when checking if a connection can be assumed.

### DIFF
--- a/src/devices/nm-device-bridge.c
+++ b/src/devices/nm-device-bridge.c
@@ -173,33 +173,34 @@ typedef struct {
 	uint nm_default;
 	bool default_if_zero;
 	bool user_hz_compensate;
+	bool only_with_stp;
 } Option;
 
 static const Option master_options[] = {
-	{ NM_SETTING_BRIDGE_STP,                "stp_state",
+	{ NM_SETTING_BRIDGE_STP,                "stp_state", /* this must stay as the first item */
 	                                        0, 1, 1,
-	                                        FALSE, FALSE },
+	                                        FALSE, FALSE, FALSE },
 	{ NM_SETTING_BRIDGE_PRIORITY,           "priority",
 	                                        0, G_MAXUINT16, 0x8000,
-	                                        TRUE, FALSE },
+	                                        TRUE, FALSE, TRUE },
 	{ NM_SETTING_BRIDGE_FORWARD_DELAY,      "forward_delay",
 	                                        0, NM_BR_MAX_FORWARD_DELAY, 15,
-	                                        TRUE, TRUE },
+	                                        TRUE, TRUE, TRUE},
 	{ NM_SETTING_BRIDGE_HELLO_TIME,         "hello_time",
 	                                        0, NM_BR_MAX_HELLO_TIME, 2,
-	                                        TRUE, TRUE },
+	                                        TRUE, TRUE, TRUE },
 	{ NM_SETTING_BRIDGE_MAX_AGE,            "max_age",
 	                                        0, NM_BR_MAX_MAX_AGE, 20,
-	                                        TRUE, TRUE },
+	                                        TRUE, TRUE, TRUE },
 	{ NM_SETTING_BRIDGE_AGEING_TIME,        "ageing_time",
 	                                        NM_BR_MIN_AGEING_TIME, NM_BR_MAX_AGEING_TIME, 300,
-	                                        TRUE, TRUE },
+	                                        TRUE, TRUE, FALSE },
 	{ NM_SETTING_BRIDGE_GROUP_FORWARD_MASK, "group_fwd_mask",
 	                                        0, 0xFFFF, 0,
-	                                        TRUE, FALSE },
+	                                        TRUE, FALSE, FALSE },
 	{ NM_SETTING_BRIDGE_MULTICAST_SNOOPING, "multicast_snooping",
 	                                        0, 1, 1,
-	                                        FALSE, FALSE },
+	                                        FALSE, FALSE, FALSE },
 	{ NULL, NULL }
 };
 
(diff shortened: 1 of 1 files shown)


src/devices/nm-device-bridge.c
```diff
@@ -275,16 +276,32 @@ update_connection (NMDevice *device, NMConnection *connection)
...
 		s_bridge = (NMSettingBridge *) nm_setting_bridge_new ();
 		nm_connection_add_setting (connection, (NMSetting *) s_bridge);
 	}
 
-	for (option = master_options; option->name; option++) {
+	option = master_options;
+	nm_assert (nm_streq (option->name, "stp_state"));
+
+	stp = nm_platform_sysctl_master_get_option (nm_device_get_platform (device), ifindex, option->sysname);
+	nm_assert (stp);
```
thom311: you must not assert against results of IO. IO can fail for a multitude of reasons any time.

src/devices/nm-device-bridge.c
```diff
@@ -275,16 +276,32 @@ update_connection (NMDevice *device, NMConnection *connection)
...
+
+	stp_value = strtol (stp, NULL, 10);
```
thom311: don't use `strtol()`. It is locale dependent. Also, you lack error checking here. Use `_nm_utils_ascii_str_to_int64()`.

Yes, that is also below. Please add error checking there too.

Also, fix calling `g_object_set()` without checking that the integer value is in a suitable range. First of all, various of these properties are `g_param_spec_uint()` so passing `int value` is at least ugly (if not wrong). If you fix the uint vs. int issue, you could use `nm_g_object_set()`, which returns an error instead of asserting when the value is out of range.

src/devices/nm-device-bridge.c
```diff
@@ -170,17 +170,18 @@ typedef struct {
 	const char *sysname;
 	gboolean default_if_zero;
 	gboolean user_hz_compensate;
+	gboolean only_with_stp;
```
thom311: Please make these types `bool`. Otherwise it wastes 9*8 bytes (on common 64 bit archs), for no benefit.

src/devices/nm-device-bridge.c
fgiudici: agreed

src/devices/nm-device-bridge.c
fgiudici: We can use _nm_utils_ascii_str_to_int64() to do also the check on the nm allowed range: if out of that, the default nm value will be assumed.
Ok for value as a uint.

src/devices/nm-device-bridge.c
fgiudici: ok

libnm-core/nm-core-internal.h
```diff
@@ -101,6 +101,11 @@
 #define NM_BR_MIN_AGEING_TIME   0
 #define NM_BR_MAX_AGEING_TIME   1000000
 
+#define NM_BR_PORT_MAX_PRIORITY 63
```
thom311: typo in commit message (a spurious space at `move bridge port min/max/default values  to core-internal`).

src/devices/nm-device-bridge.c
```diff
@@ -168,26 +168,29 @@ complete_connection (NMDevice *device,
...
-	{ NM_SETTING_BRIDGE_GROUP_FORWARD_MASK, "group_fwd_mask", TRUE, FALSE },
-	{ NM_SETTING_BRIDGE_MULTICAST_SNOOPING, "multicast_snooping", FALSE, FALSE },
+	{ NM_SETTING_BRIDGE_STP, "stp_state", 0, 1, 1, FALSE, FALSE },
+	{ NM_SETTING_BRIDGE_PRIORITY, "priority", 0, G_MAXUINT16, 0x8000, TRUE, FALSE },
+	{ NM_SETTING_BRIDGE_FORWARD_DELAY, "forward_delay", 0, NM_BR_MAX_FORWARD_DELAY, 15, TRUE, TRUE },
+	{ NM_SETTING_BRIDGE_HELLO_TIME, "hello_time", 0, NM_BR_MAX_HELLO_TIME, 2, TRUE, TRUE },
+	{ NM_SETTING_BRIDGE_MAX_AGE, "max_age", 0, NM_BR_MAX_MAX_AGE, 20, TRUE, TRUE },
+	{ NM_SETTING_BRIDGE_AGEING_TIME, "ageing_time", NM_BR_MIN_AGEING_TIME, NM_BR_MAX_AGEING_TIME, 300, TRUE, TRUE },
+	{ NM_SETTING_BRIDGE_GROUP_FORWARD_MASK, "group_fwd_mask", 0, 0xFFFF, 0, TRUE, FALSE },
+	{ NM_SETTING_BRIDGE_MULTICAST_SNOOPING, "multicast_snooping", 0, 1, 1, FALSE, FALSE },
```
thom311: can you vertically align the properties, with spaces? This is hard hard to read.

Also, it's not clear to me what `in-memory mapped` here means.

src/devices/nm-device-bridge.c
```diff
@@ -303,6 +317,7 @@ update_connection (NMDevice *device, NMConnection *connection)
 				                                      option->nm_default);
 			}
 			g_object_set (s_bridge, option->name, value, NULL);
+
```
bengal: Extra empty line. The rest LGTM.

libnm-core/nm-core-internal.h
fgiudici: Fixed, thanks

src/devices/nm-device-bridge.c
fgiudici: Ok, bridge options are now aligned.
Removed the "in-memory mapped" term.

src/devices/nm-device-bridge.c
fgiudici: ops, extra line removed, thanks

libnm-core/nm-core-internal.h
fgiudici: fixed, thanks

thom311:
lgtm

PR #267, opened 2018-12-22
[th/nm-error]

thom311:
We have various forms to report a failure form functions:

1) return a boolean value indicating success/failure (or `NULL` for pointer values)
2) like 1), but also set a `GError`
3) systemd uses negative integer values which contains errors from `<errno.h>`.
4) `src/platform/nm-netlink.c` is like 3), except it extends the range of error numbers with own `NLE_*` values.
5) platform returns `NMPlatformError`, which
    - for negative results, it signals a failure with the error value from <errno.h>
    - zero is `NM_PLATFORM_ERROR_SUCCESS`.
    - positive values are `NM_PLATFORM_ERROR_*` failure code.

Unify 4) and 5) in a new `nm-error.h` error domain. This is like 3), except, that we extend the regular `<errno.h>` codes with values from `NME_*`.


thom311:
merged to master.

PR #268, opened 2018-12-22
core: set number of SR-IOV VFs asynchronously

bengal:
When changing the number of VFs the kernel can block for very long
time in the write() to sysfs, especially if autoprobe-drivers is
enabled. Do it asynchronously to avoid blocking the entire NM
process. Use g_output_stream_write_async(), that performs the write in
a different thread when writing to regular files.

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -128,6 +128,15 @@ typedef struct {
 	int ifindex;
 } DeleteOnDeactivateData;
 
+typedef struct {
+	NMDevice *device;
+	GCancellable *cancellable;
+	NMPlatformAsyncCallback callback;
+	gpointer callback_data;
+	guint num_vfs;
+	NMTernary autoprobe;
+} SriovOp;
+
 typedef void (*AcdCallback) (NMDevice *, NMIP4Config **, gboolean);
 
 typedef struct {
@@ -572,6 +581,12 @@ typedef struct _NMDevicePrivate {
 	} concheck_x[2];
 
 	guint check_delete_unrealized_id;
+	guint deactivate_pending_count;
+
+	struct {
+		SriovOp *pending;    /* SR-IOV operation currently running */
+		SriovOp *next;       /* next SR-IOV operation scheduled */
+	} sriov;
 
 	struct {
 		guint timeout_id;
@@ -579,7 +594,6 @@ typedef struct _NMDevicePrivate {
 		guint64 tx_bytes;
 		guint64 rx_bytes;
 	} stats;
-
 } NMDevicePrivate;
 
 G_DEFINE_ABSTRACT_TYPE (NMDevice, nm_device, NM_TYPE_DBUS_OBJECT)
@@ -4190,6 +4204,85 @@ nm_device_update_from_platform_link (NMDevice *self, const NMPlatformLink *plink
 	}
 }
 
+static void sriov_op_cb (GError *error, gpointer user_data);
+
+static void
+sriov_op_start (NMDevice *self, SriovOp *op)
+{
+	NMDevicePrivate *priv = NM_DEVICE_GET_PRIVATE (self);
+
+	nm_assert (!priv->sriov.pending);
+
+	op->cancellable = g_cancellable_new ();
+	op->device = g_object_ref (self);
+	priv->sriov.pending = op;
+
+	nm_platform_link_set_sriov_params_async (nm_device_get_platform (self),
+	                                         priv->ifindex,
+	                                         op->num_vfs,
+	                                         op->autoprobe,
+	                                         sriov_op_cb,
+	                                         op,
+	                                         op->cancellable);
+}
+
+static void
+sriov_op_cb (GError *error, gpointer user_data)
+{
+	SriovOp *op = user_data;
+	gs_unref_object NMDevice *self = op->device;
+	NMDevicePrivate *priv = NM_DEVICE_GET_PRIVATE (self);
+
+	nm_assert (op == priv->sriov.pending);
+
+	if (op->callback)
+		op->callback (error, op->callback_data);
+
+	nm_clear_g_cancellable (&op->cancellable);
+	g_slice_free (SriovOp, op);
+	priv->sriov.pending = NULL;
+
+	if (priv->sriov.next) {
+		sriov_op_start (self, priv->sriov.next);
+		priv->sriov.next = NULL;
+	}
+}
+
+static void
+sriov_op_queue (NMDevice *self,
+                guint num_vfs,
+                NMTernary autoprobe,
+                NMPlatformAsyncCallback callback,
+                gpointer callback_data)
+{
+	NMDevicePrivate *priv = NM_DEVICE_GET_PRIVATE (self);
+	GError *error = NULL;
+	SriovOp *op;
+
+	op = g_slice_new0 (SriovOp);
+	op->num_vfs = num_vfs;
+	op->autoprobe = autoprobe;
+	op->callback = callback;
+	op->callback_data = callback_data;
+
+	if (priv->sriov.next) {
+		/* Cancel the next operation immediately */
+		if (priv->sriov.next->callback) {
+			nm_utils_error_set_cancelled (&error, FALSE, NULL);
+			priv->sriov.next->callback (error, priv->sriov.next->callback_data);
+			g_clear_error (&error);
+		}
+		g_slice_free (SriovOp, priv->sriov.next);
+		priv->sriov.next = NULL;
+	}
+
+	if (priv->sriov.pending) {
+		priv->sriov.next = op;
+		g_cancellable_cancel (priv->sriov.pending->cancellable);
+	} else
+		sriov_op_start (self, op);
+}
+
 static void
 device_init_static_sriov_num_vfs (NMDevice *self)
 {
@@ -4204,10 +4297,8 @@ device_init_static_sriov_num_vfs (NMDevice *self)
 		                                          self,
 		                                          NULL);
 		num_vfs = _nm_utils_ascii_str_to_int64 (value, 10, 0, G_MAXINT32, -1);
-		if (num_vfs >= 0) {
-			nm_platform_link_set_sriov_params (nm_device_get_platform (self),
-			                                   priv->ifindex, num_vfs, NM_TERNARY_DEFAULT);
-		}
+		if (num_vfs >= 0)
+			sriov_op_queue (self, num_vfs, NM_TERNARY_DEFAULT, NULL, NULL);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


thom311:
if two or more writes are started (for the same device and sysctl), before the first is completed, how will this react? Shouldn't there be a per-sysctl-file queue, meaning, to keep track of the currently pending `GOutputStream`, and (if one exists), append to that (or skip writes that are still queued but are obsolte)?

Also, you add here API to write SR-IOV paramters asynchronously. Why is this not general API for writing a sysctl asynchronously?

thom311:
> if two or more writes are started (for the same device and sysctl), before the first is completed, how will this react? Shouldn't there be a per-sysctl-file queue, meaning, to keep track of the currently pending GOutputStream, and (if one exists), append to that (or skip writes that are still queued but are obsolte)?

`info->values[info->num_values++]` is infact a queue... but since NMPlatform does not rememer the pending writes, it will always create a new queue, instead of appending/replacing an existing one. It should be simple to extend, to keep the list of pending queues and see if one exists (and have a strategy for append/obsolete/replace/reorder).

bengal:
> if two or more writes are started (for the same device and sysctl), before the first is completed, how will this react? Shouldn't there be a per-sysctl-file queue, meaning, to keep track of the currently pending `GOutputStream`, and (if one exists), append to that (or skip writes that are still queued but are obsolte)?

Callers can either (a) wait for the previous write to finish before writing a new value or (b) cancel pending writes through the GCancellable. NMDevice does (b) because when we want to write a new value the old one becomes irrelevant.

bengal:
> Also, you add here API to write SR-IOV paramters asynchronously. Why is this not general API for writing a sysctl asynchronously?

Ok, added the sysctl_set_async API, currently used only for SR-IOV.

> `info->values[info->num_values++]` is infact a queue... but since NMPlatform does not rememer the pending writes, it will always create a new queue, instead of appending/replacing an existing one. It should be simple to extend, to keep the list of pending queues and see if one exists (and have a strategy for append/obsolete/replace/reorder).

I don't think this is useful, usually there is only one place where a given sysctl value is set and the caller already has to remember the current state (because of the callback and the gcancellable), so when a new value must be set, the caller can cancel the previous operation and start a new one.

src/platform/nm-linux-platform.c
```diff
@@ -5935,16 +5950,25 @@ link_set_sriov_params (NMPlatform *platform,
...
 
-	return TRUE;
+	i = 0;
+	if (current_num != 0)
+		values[i++] = "0\n";
+	if (num_vfs != 0)
+		values[i++] = nm_sprintf_bufa (128, "%u\n", num_vfs);
+	values[i++] = NULL;
+
+	nm_sprintf_buf (buf, "/sys/class/net/%s/device/sriov_numvfs", ifname);
```
thom311: this open a renaming race, that `NMP_SYSCTL_PATHID_NETDIR()` avoided.

src/platform/nm-linux-platform.c
```diff
@@ -5935,16 +5950,25 @@ link_set_sriov_params (NMPlatform *platform,
...
-	                             nm_sprintf_buf (buf, "%u", num_vfs))) {
-		_LOGW ("link: couldn't set SR-IOV num_vfs to %d: %s", num_vfs, strerror (errno));
-		return FALSE;
+	if (current_num == 0 && num_vfs == 0) {
+		if (callback) {
+			packed = nm_utils_user_data_pack (g_object_ref (platform),
+			                                  callback,
+			                                  data);
+			nm_utils_invoke_on_idle (sriov_idle_cb, packed, cancellable);
+		}
```
thom311: `nm_platform_sysctl_set_async()` returns a boolean to indicate whether the action was successfully started. On the one hand, I dislike this inconsistency (an async operation should never fail to start). On the other hand, it's actually simpler to implement and avoids an unnecessary idle action, so OK.

But here, you have  a case where you do invoke a trivial response in an idle handler. Either always schedule async actions (for consistency), or don't schedule such idle actions (for efficiency).

valgrind.suppressions
```diff
@@ -499,3 +499,18 @@
    fun:btrfs_reflink_with_progress
    ...
 }
+{
+   _g_daemon_vfs_init
```
thom311: do we really need this VFS overhead? Can you not create a GIOStream based on the filename alone without GFile?

src/platform/nm-linux-platform.c
bengal: Yes, but from what I see GLib only allows creating a GFileIOStream
from a GFile, and GFiles can only be constructed by filename, not by file
descriptor. I researched alternative ways to perform asynchronous I/O
on regular files and so far the GLib is the easiest to integrate with.

Alternatives are:

 - use the glibc POSIX asynchronous I/O API (man 7 aio). It internally
   uses a thread pool to perform the operation and runs a callback in
   one of such threads;

 - implements our own solution using threads, which seems not trivial;

 - or stick with the GLib API, and accept the chance of a race during
   interface rename.

What do you think?

thom311: other alternative: use glib's threadpool. I suppose that exists, because that is what GFileIOStream uses (I presume).

src/platform/nm-linux-platform.c
bengal: Fixed, the functions now always return asynchronously.

valgrind.suppressions
bengal: I've switched to g_task_run_in_thread() now.

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+                        GCancellable *cancellable)
+{
+	nm_auto_pop_netns NMPNetns *netns = NULL;
+	SysctlAsyncInfo *info = task_data;
+	GError *error = NULL;
+	char **value;
+
+	if (   info->dirfd < 0
+	    && !nm_platform_netns_push (info->platform, &netns)) {
+		g_set_error_literal (&error, 1, 0, "sysctl: failed changing namespace");
```
thom311: We should never pass "1, 0" as error domain/code. If you want less typing, use `nm_utils_error_set*()`.

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+	}
+
+	for (value = info->values; *value; value++) {
+		if (!sysctl_set_internal (info->platform,
+		                          info->pathid,
+		                          info->dirfd,
+		                          info->path,
+		                          *value)) {
+			g_set_error (&error,
+			             1, 0,
```
thom311: here too.

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+	ASSERT_SYSCTL_ARGS (pathid, dirfd, path);
+
+	if (dirfd >= 0) {
+		dirfd_dup = fcntl (dirfd, F_DUPFD_CLOEXEC, 0);
+		if (dirfd_dup < 0) {
+			if (!callback)
+				return;
+			errsv = errno;
+			g_set_error (&error,
+			             1, 0,
```
thom311: here too.  Also note `nm_utils_error_set_errno()`.

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+                              GCancellable *cancellable)
+{
+	gs_unref_object NMPlatform *platform = NULL;
+	gs_free_error GError *cancelled_error = NULL;
+	gs_free_error GError *error = NULL;
+	NMPlatformAsyncCallback callback;
+	gpointer callback_data;
+
+	g_cancellable_set_error_if_cancelled (cancellable, &cancelled_error);
+	nm_utils_user_data_unpack (user_data, &platform, &callback, &callback_data, &error);
```
thom311: could we always unpack the user-data as very first step in a callback?

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+			             1, 0,
+			             "sysctl: failure duplicating directory fd: %s",
+			             nm_strerror_native (errsv));
+			packed = nm_utils_user_data_pack (g_object_ref (platform),
+			                                  callback,
+			                                  data,
+			                                  error);
+			nm_utils_invoke_on_idle (sysctl_set_async_return_idle,
+			                         packed,
+			                         g_object_ref (cancellable));
```
thom311: you don't need to reference the cancellable. `nm_utils_invoke_on_idle()` must do that already. Then, in the callback you also don't need to unref it.

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+
+	if (!g_task_propagate_boolean (task, &error)) {
+		if (info->callback)
+			info->callback (error, info->callback_data);
+	} else {
+		values_str = g_strjoinv (", ", info->values);
+		platform = info->platform;
+		_LOGD ("sysctl: successfully set asynchronously '%s' to values '%s'",
+		       info->pathid ?: info->path, values_str);
+		if (info->callback)
```
thom311: let's move the invocation of the callback out of the `if` block (and merge the if-else blocks)

src/platform/nm-linux-platform.c
```diff
@@ -4187,6 +4188,201 @@ sysctl_set (NMPlatform *platform, const char *pathid, int dirfd, const char *pat
...
+
+	if (   info->dirfd < 0
+	    && !nm_platform_netns_push (info->platform, &netns)) {
+		g_set_error_literal (&error, 1, 0, "sysctl: failed changing namespace");
+		g_task_return_error (task, error);
+		return;
+	}
+
+	for (value = info->values; *value; value++) {
+		if (!sysctl_set_internal (info->platform,
```
thom311: before writing, you need to check whether the operation was cancelled in the meantime and abort.

src/devices/nm-device.c
```diff
@@ -4194,8 +4203,15 @@ device_init_static_sriov_num_vfs (NMDevice *self)
 		                                          NULL);
 		num_vfs = _nm_utils_ascii_str_to_int64 (value, 10, 0, G_MAXINT32, -1);
 		if (num_vfs >= 0) {
-			nm_platform_link_set_sriov_params (nm_device_get_platform (self),
-			                                   priv->ifindex, num_vfs, NM_TERNARY_DEFAULT);
+			nm_clear_g_cancellable (&priv->sriov_cancellable);
+			priv->sriov_cancellable = g_cancellable_new ();
+			nm_platform_link_set_sriov_params_async (nm_device_get_platform (self),
+			                                         priv->ifindex,
```
thom311: you cannot just "cancel && create-cancelleable && set-async"

If an async operation is going on, you need to cancel it, wait for the callback to return, and then invoke the async operation. Otherwise, there is a race.

bengal:
Repushed, now NMDevice waits for the cancellation of a pending SR-IOV operation before proceeding.

thom311:
Pushed two minor fixups. The branch looks very good do me

bengal:
Thanks, squashed the fixups and merged as 3a4a09b669b6d3498017fe5773af3e9aef2ae8a5.

PR #270, opened 2018-12-28
[th/shared-systemd]

thom311:
Move the `systemd/src/basic` part of our systemd clone to `shared/` and compile a (intermediate) static library of it.

This way, we will be able to use utilities from there in libnm-core (and in libnm). In particular, I need `unbase64mem()` for validating base64 encoding.


bengal:
It's too bad that glib doesn't provide functions for encoding/decoding
base64 with input validation, perhaps we should implement and submit
them upstream.

I'm a bit worried that depending on systemd internals will cause more
conflicts when we need to resynch the code. Anyway, we can merge this
and see if it causes too much work in the future (probably not, as
long as we reuse only few functions).

thom311:
> perhaps we should implement and submit them upstream.

Maybe. You are very welcome to do so. However, doing that would not help NetworkManager for years, because we couldn't just bump the glib requirement to the next glib release. Also, it seems hard to come up with a generally suitable API for glib, which makes everybody happy.

> I'm a bit worried that depending on systemd internals will cause more
conflicts when we need to resynch the code. 

I don't think that danger is too great, because:

- we don't use all internal API. Instead, only the one exposed via the the `nm-sd-utils-*.h` header.

- functions that are exposed, must all have tests (I'll add a test for `nm_sd_utils_unbase64char()` which is still missing).

- these functions aren't really complex. Even if upstream systemd is going to change them, it's not that it will cause us much effort to adjust our callers (in particular, if we have unit tests). If that's not the case, then we must be very careful about using such systemd API. Notably, API like `sd_dhcp_client` *is* relatively complex and we still use it. The decision to use complex API is a hard one, but using trivial API isn't really dangerous.

The alternative is copy&paste or implement our own variant. That's likely more effort. This way at least leverage the work of systemd (which also has tests and does fuzzing).

None of these points are new or changed by the fact that `libnm-systemd-core.la` is now independent and can be used by `libnm-core`. We still should be mindful about *what* we use.

thom311:
merged to master

PR #271, opened 2018-12-29
[th/wireguard-pt2]

thom311:
WIP for adding WireGuard support.

It not yet complete, in particular:

- it does not yet support peers (nor peers' allowed-ips), which makes it rather lacking.
- currently, ipv4.method still defaults to "auto", which isn't working. I wonder, whether "auto" should be forbidden entirely, or whether auto should get a new meaning (see commit adbb9eb2466913bc518ad1904b1727c49f08a2fc).
- not yet well tested.

The branch is base on https://github.com/NetworkManager/NetworkManager/pull/270


thom311:
Closing pull-request for the moment, as this branch is pending on two other branches which are still on review. Once that requisite work is done, it will be reopened.

thom311:
The work that was on this branch was split into different branches, and separate pull-requests opened for them.

there is nothing interesting left here...

PR #272, opened 2019-01-02
[th/various-fixes]

thom311:
Various bugfixes (and cleanups), that were done while working on the wireguard branch...

### DIFF
--- a/libnm-core/nm-keyfile.c
+++ b/libnm-core/nm-keyfile.c
@@ -106,17 +106,18 @@ static void
 setting_alias_parser (KeyfileReaderInfo *info, NMSetting *setting, const char *key)
 {
 	const char *setting_name = nm_setting_get_name (setting);
-	char *s;
 	const char *key_setting_name;
+	gs_free char *s = NULL;
 
 	s = nm_keyfile_plugin_kf_get_string (info->keyfile, setting_name, key, NULL);
-	if (s) {
-		key_setting_name = nm_keyfile_plugin_get_setting_name_for_alias (s);
-		g_object_set (G_OBJECT (setting),
-		              key, key_setting_name ?: s,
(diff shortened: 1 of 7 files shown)


libnm-core/nm-keyfile.c
```diff
@@ -2896,6 +2898,9 @@ nm_keyfile_read (GKeyFile *keyfile,
 	info.user_data = user_data;
 
 	groups = g_key_file_get_groups (keyfile, &length);
+	if (!groups)
+		length = 0;
+
```
bengal: g_key_file_get_*_list() always set the length, while g_key_file_get_keys() does not; this is a bit inconsistent in glib. Anyway I think it's correct not to assume that the length is set because the function can return an error.

OTOH, g_key_file_get_groups () always sets _length_ and can't fail so I think it's safe to rely on _length_ value.

thom311: hm, that's true. In several cases, it's not a severe bug.
- not severe for `g_key_file_get_groups()` as it never fails
- not severe for `g_key_file_get_*_list()` as they actually set the length (undocumented)
- not severe for our call of `g_key_file_get_keys()`, because in the circumstances we call the function it's not expected to fail.

Anyway, I'd keep all the changes there, simply to be explicit about it. It doesn't hurt (and it was only partly clear to me that it is in fact not needed). I'd update the commit message though, stating that it the issue is not severe.

bengal: > Anyway, I'd keep all the changes there, simply to be explicit about it. It doesn't hurt (and it was only partly clear to me that it is in fact not needed). I'd update the commit message though, stating that it the issue is not severe.

Okay.

thom311:
merged to master.

PR #273, opened 2019-01-03
[th/keyfile-fixes]

thom311:
A set of patches with keyfile and libnm fixes/rework/cleanup.

This is another early result from the wireguard branch.

### DIFF
--- a/libnm-core/nm-connection.c
+++ b/libnm-core/nm-connection.c
@@ -2034,6 +2034,73 @@ nm_connection_for_each_setting_value (NMConnection *connection,
 }
 
 /**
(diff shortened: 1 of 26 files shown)


thom311:
hm, with some quick profiling, the unit tests are notably faster with this change.
However, the binary size also increases drastically. That needs investigation...

bengal:
LGTM

thom311:
Thanks Beiamino.

Merged to master.


(last week CI was failing, but I am pretty confident that I fixed that. I cannot verify, because at the moment, CI breaks for other reasons.)

PR #274, opened 2019-01-06
[th/wireguard-pt2]

thom311:
Another early branch for WireGuard support.

This adds some platform related bits and is ready to merge.


bengal:
LGTM

PR #276, opened 2019-01-09
supplicant: set optional PMF using global supplicant property

bengal:
wpa_supplicant is going to change the global default for PMF from 0
(disabled) to 1 (optional) [1], so NM code needs to be adjusted to
work with all wpa_supplicant versions. Furthermore, it is better to
set optional PMF using the 'Pmf' property instead of the 'ieee80211w'
configuration option because the former better handles missing support
in driver [2].

Note that each interface in wpa_supplicant has its own copy of global
configuration and so 'global' options must still be set on each
interface. So, let's set Pmf=1 when each interface gets created and
override it with ieee80211w={0,2} if needed during association.

[1] http://lists.infradead.org/pipermail/hostap/2018-November/039009.html
[2] http://lists.infradead.org/pipermail/hostap/2019-January/039215.html

### DIFF
--- a/src/supplicant/nm-supplicant-config.c
+++ b/src/supplicant/nm-supplicant-config.c
@@ -869,11 +869,11 @@ nm_supplicant_config_add_setting_wireless_security (NMSupplicantConfig *self,
 
 		if (   !nm_streq (key_mgmt, "wpa-none")
 		    && NM_IN_SET (pmf,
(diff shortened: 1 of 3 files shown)


thom311:
lgtm

bengal:
Merged as a9ab50efb10dfb50cfe897c58afa300f8b07f6ba.

PR #277, opened 2019-01-09
[th/libnm-setting-compare]

thom311:
Rework NMSetting's `compare_property()` implementation.

Most of NMSetting is still entirely based on GObject properties. That will not work well with WireGuard peers, so NMSetting should allow to hook all kind of operations, even if the setting implementation does not use GObject properties as backend.

Rework `compare_property()` to move the pre-checks of whether to check the property based on the compare-flags (`should_compare_prop()`) into the virtual function, so that it can be implemented accordingly.


thom311:
> two crashes:

Should be fixed. Again...

bengal:
LGTM

thom311:
merged to master

PR #279, opened 2019-01-11
[th/libnm-settings-properties]

thom311:
Some refactoring to let also `libnm-core/` follow the patterns from `src/`, with respect to their GObject implementations.

The main reason I did this, is because whenever we create a new setting class, we start by copying an existing one. So, the existing ones should be good examples for how to do it.


bengal:
LGTM

PR #280, opened 2019-01-12
clients: fix appending to dns-property variable

bengal:
https://bugzilla.redhat.com/show_bug.cgi?id=1665649

### DIFF
--- a/clients/common/nm-meta-setting-desc.c
+++ b/clients/common/nm-meta-setting-desc.c
@@ -3350,7 +3350,7 @@ DEFINE_REMOVER_INDEX_OR_VALUE (_remove_fcn_ipv4_config_dns,
                                _validate_and_remove_ipv4_dns)
 
 static gboolean
-_set_fcn_ip4_config_dns_search (ARGS_SET_FCN)
+_set_fcn_ip_config_dns_search (ARGS_SET_FCN)
 {
 	gs_free const char **strv = NULL;
 	gsize i;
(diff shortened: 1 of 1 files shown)


clients/common/nm-meta-setting-desc.c
```diff
@@ -3396,7 +3396,6 @@ _set_fcn_ip_config_dns_options (ARGS_SET_FCN)
 
 	nm_assert (!error || !*error);
 
-	nm_setting_ip_config_clear_dns_options (NM_SETTING_IP_CONFIG (setting), TRUE);
```
thom311: it's correct, but it's rather ugly how this works.

The caller does
```
»···if (modifier != '-') {
»···»···/* Set/add value */
»···»···if (modifier != '+') {
»···»···»···/* We allow the existing property value to be passed as parameter,
»···»···»··· * so make a copy if we are going to free it.
»···»···»··· */
»···»···»···value = value_free = g_strdup (value);
»···»···»···nmc_setting_reset_property (setting, property_name, NULL);
»···»···}
»···»···if (!nmc_setting_set_property (client, setting, property_name, value, &local)) {
```

instead, there should be only one "set_property" function, that gets the modifier. And based on the modifier, the implementation decides whether to append/remove/reset. IOW, set/reset/remove functions should be merged, and act according to an modifier argument.

TODO.

bengal:
Merged as 9b0c1ef5d001ad0b2b9173e104871c91f4c95426.

PR #281, opened 2019-01-13
[th/wireguard-pt3]

thom311:
A major part for WireGuard support.

Still missing:

- nmcli support (but there is an example script that can be used instead)
- support for preshared-key (and requesting secrets)
- more testing
- see `TODO.txt` file

But I think it's at a point where it could be merged. We can make it perfect afterwards.

PR #282, opened 2019-01-13
[th/various for wireguard-1]

thom311:
Different smaller, individual patches that go before https://github.com/NetworkManager/NetworkManager/pull/281

### DIFF
--- a/libnm-core/nm-connection.c
+++ b/libnm-core/nm-connection.c
@@ -105,7 +105,7 @@ setting_changed_cb (NMSetting *setting,
 static gboolean
 _setting_release (gpointer key, gpointer value, gpointer user_data)
 {
-	g_signal_handlers_disconnect_by_func (user_data, setting_changed_cb, value);
+	g_signal_handlers_disconnect_by_func (value, setting_changed_cb, user_data);
 	return TRUE;
 }
 
--- a/libnm-core/nm-connection.c
+++ b/libnm-core/nm-connection.c
@@ -1832,26 +1832,14 @@ nm_connection_need_secrets (NMConnection *connection,
 void
 nm_connection_clear_secrets (NMConnection *connection)
 {
-	GHashTableIter iter;
-	NMSetting *setting;
-
-	g_return_if_fail (NM_IS_CONNECTION (connection));
-
-	g_hash_table_iter_init (&iter, NM_CONNECTION_GET_PRIVATE (connection)->settings);
-	while (g_hash_table_iter_next (&iter, NULL, (gpointer) &setting)) {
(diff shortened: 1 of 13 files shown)


shared/nm-utils/nm-shared-utils.h
```diff
@@ -102,6 +102,7 @@ typedef struct {
 	union {
 		guint8 addr_ptr[1];
 		in_addr_t addr4;
+		struct in_addr addr4_struct;
```
lkundrak: Why?

Also, if you actually need "struct in_addr", then you perhaps could drop the "in_addr_t" field, because it could be accessed via the "struct in_addr".

src/platform/nm-linux-platform.c
```diff
@@ -2374,6 +2374,7 @@ _wireguard_create_change_nlmsgs (NMPlatform *platform,
                                  const NMPlatformLnkWireGuard *lnk_wireguard,
                                  const NMPWireGuardPeer *peers,
                                  guint peers_len,
+                                 gboolean replace_peers,
```
lkundrak: This needs a better commit message.

libnm-core/nm-connection.c
```diff
@@ -102,10 +102,16 @@ setting_changed_cb (NMSetting *setting,
 	g_signal_emit (self, signals[CHANGED], 0);
 }
 
+static void
```
lkundrak: I don't get it how is this commit an improvement.

shared/nm-utils/nm-shared-utils.h
thom311: Because I'll need to assign from `NMIPAddr` to a field of type `in_addr`:

```
+         *((struct sockaddr_in *) sockaddr) = (struct sockaddr_in) {
+              .sin_family = AF_INET,
+              .sin_addr   = addrbin.addr4_struct,
+              .sin_port   = htons (self->port),
+         };
```
Yes, I know it could be
```
+              .sin_addr   = { .s_addr = addrbin.addr4, },       
```

But would that be better? I don't understand which problem you see with the union. `NMIPAddr` main purpose is to be a union type, having another alias doesn't seem ugly to me.

Changing `addr4` from `in_addr_t` to `struct in_addr` will require a lot of changes throughout . In particular, because most of our internal API for IPv4 addresses is based on `in_addr_t` / `guint32` / `be32_t`, instead of `struct in_addr`. For example, `nm_utils_ip4_address_*()`. I would not object to refactor the code to prefer `struct in_addr` everywhere, but I don't mind either way. Anyway, not as part of this pull-request.

src/platform/nm-linux-platform.c
thom311: done.

thom311:
merged to master

PR #285, opened 2019-01-14
[th/connectivity-empty-response]

thom311:
fix handling support for empty "connectivity.response" setting.

Previously, we would have accepted any response as valid.


thom311:
merged. Thanks

PR #286, opened 2019-01-19
[th/logging-mt]

thom311:
Make `NMPNetns` and (the logging part of) nm-logging thread-safe.


thom311:
C11s threads are not supported on Ubuntu (TravisCI). Changed to use pthreads and repushed.

thom311:
merged to master

PR #288, opened 2019-01-22
[th/secret-agent-cleanup] cleanups to SecretAgent in libnm and clients/common

thom311:


vbenes:
we have a crash:
https://faf.lab.eng.brq.redhat.com/faf//reports/10409/

thom311:
> we have a crash:

should be fixed. Repushed.

thom311:
merged to master

PR #290, opened 2019-01-28
[th/wifi-p2p-various]

thom311:
Follow-up after merging ce3f7bf812c21773251748693cd39d2fb54b9055. Most notable:

- rename stuff, to consistently name it "Wi-Fi P2P".

- drop `nm_device_p2p_wifi_start_find()` and `nm_device_p2p_wifi_stop_find()` for the moment.

### DIFF
--- a/libnm/nm-device-wifi-p2p.c
+++ b/libnm/nm-device-wifi-p2p.c
@@ -203,6 +203,7 @@ start_find_finished_cb (GObject      *obj,
 /**
  * nm_device_wifi_p2p_start_find:
  * @device: a #NMDeviceWifiP2P
+ * @options: (allow-none): optional options passed to StartFind.
  * @cancellable: a #GCancellable, or %NULL
  * @callback: a #GAsyncReadyCallback, or %NULL
  * @user_data: user_data for @callback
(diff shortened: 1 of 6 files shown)


benzea:
Looks good to me. Thanks for all the cleanups!

thom311:
@benzea added asynchronous find_start()/find_stop() API in https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/72 .

Since that patch is based on this pull-request, I am going to close merge-request 72 and take the patch on this pull-request.

libnm/nm-device-p2p-wifi.c
```diff
@@ -410,7 +389,11 @@ finalize (GObject *object)
 {
 	NMDeviceP2PWifiPrivate *priv = NM_DEVICE_P2P_WIFI_GET_PRIVATE (object);
 
+	g_clear_object (&priv->proxy);
 	g_free (priv->hw_address);
+	g_byte_array_unref (priv->wfd_ies);
```
benzea: `priv->wfd_ies` might be NULL here. So this should probably be `g_clear_pointer (&priv->wfd_ies. g_byte_array_unref);` or just an if as with the `peers`.

libnm/nm-device-wifi-p2p.c
```diff
@@ -183,6 +183,151 @@ nm_device_wifi_p2p_get_peer_by_path (NMDeviceWifiP2P *device,
...
+ * using nm_device_p2p_wifi_stop_find().
+ *
+ * Since: 1.16
+ **/
+void
+nm_device_wifi_p2p_start_find (NMDeviceWifiP2P     *device,
+                               GCancellable        *cancellable,
+                               GAsyncReadyCallback  callback,
+                               gpointer             user_data)
+{
```
bengal: The D-Bus method also supports an 'options' argument that allows to change the timeout and possibly other things in the future. I think we should add `GVariant *options` argument to the libnm API like we did for `nm_client_add_and_activate_connection2()`.

bengal:
The rest LGTM.

libnm/nm-device-wifi-p2p.c
thom311: Good point, yes should be done.

Let me merge the branch, as there are no other objections. I'd prefer to keep the time short, where master has an API that we are going to break again (after all, this branch does a significant amount of renaming).

thom311:
OK, I merged now all the patches except the last one to master (commit 027ef98cd4b44937d07c331868f1efa01b8e8e7c).

The pull-request remains open with the remaining patch "libnm: Add async start/stop routines for P2P find operations" is still left, to address https://github.com/NetworkManager/NetworkManager/pull/290#discussion_r253098974 .

thom311:
@benzea I will look into the missing part next week, but if you beat me to it: patch welcome!

thom311:
I think the remaining issue is now addressed.

+ a few follow-up fixes.


Repushed.

bengal:
New commits LGTM.

thom311:
pt2 also merged to master. Closing.

PR #291, opened 2019-01-28
[th/various-for-wireguard-2]

thom311:
Few patches to make libnm-core's settings extentable, so that sub-classes can override and better handle non-GObject-based properties.


thom311:
merged to master.

PR #292, opened 2019-02-01
[th/errno]

thom311:
Cleanup handling of error numbers.


thom311:
I was prompted to this, by being reminded that `strerror()` is not thread-safe (which I find problematic) and that `g_strerror()` leaks all strings (which I also find problematic).

thom311:
merged to master.

Unfortunately, despite CI all green, there was still a mishap. Fixed by https://github.com/NetworkManager/NetworkManager/commit/128099151d23818501e9d9600aec64fbfe999322

PR #293, opened 2019-02-05
[th/various-for-wireguard-3]

thom311:
Yet another branch in preparation for WireGuard support.

It already adds the internal `NMSockAddrEndpoint` API, which will be used by WireGuard implementation to (internally) track and parse the endpoint.


thom311:
merged to master

PR #294, opened 2019-02-05
platform: limit the maximum size of sysctl cache

bengal:
When the logging level is DEBUG or TRACE, we keep all the sysctl
values we read in a cache to log how they change. Currently there is
no limit on the size of this cache and it can take a large amount of
memory.

Implement a LRU cache where the oldest entries are deleted to make
space for new ones.

### DIFF
--- a/src/platform/nm-linux-platform.c
+++ b/src/platform/nm-linux-platform.c
@@ -376,8 +376,8 @@ typedef struct {
 
 	bool pruning[_DELAYED_ACTION_IDX_REFRESH_ALL_NUM];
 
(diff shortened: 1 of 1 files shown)


thom311:
lgtm

thom311:
or maybe reduce the number to 1000 cached entries? Even 10000 is really large.

bengal:
> or maybe reduce the number to 1000 cached entries? Even 10000 is really large.

Makes sense, changed.

Merged as 3a0f7114fed6ec4fa2fb976bf51721c4ad328e02.

PR #296, opened 2019-02-10
Fix some ovs memory leaks

bengal:


src/devices/ovs/nm-ovsdb.c
```diff
@@ -690,6 +690,9 @@ _delete_interface (NMOvsdb *self, json_t *params, const char *ifname)
 		_expect_ovs_bridges (params, priv->db_uuid, bridges);
 		_set_ovs_bridges (params, priv->db_uuid, new_bridges);
 	}
+
+	json_decref (bridges);
+	json_decref (new_bridges);
```
thom311: can we add and use a `nm_auto_decref_json` macro?

src/devices/ovs/nm-ovsdb.c
```diff
@@ -1552,9 +1553,6 @@ dispose (GObject *object)
 	g_clear_pointer (&priv->ports, g_hash_table_destroy);
 	g_clear_pointer (&priv->interfaces, g_hash_table_destroy);
 
-	g_cancellable_cancel (priv->cancellable);
-	g_clear_object (&priv->cancellable);
-
```
thom311: dispose() is not reentrant. Can we fix that too?

src/devices/ovs/nm-ovsdb.c
bengal: Done.

src/devices/ovs/nm-ovsdb.c
bengal: Do you mean that it can be called multiple times? I added a commit to call ovsdb_disconnect() only on the first time.

thom311: `dispose()` in glib might always be called multiple times. In practice, that is not the case for NetworkManager, or here. But we shold write our dispose() code in that manner.

That basically just means to use `g_clear_object()/nm_clear_g_free()` instead of `g_object_unref()/g_free()`. Don't leave dangling pointers, and assume `dispose()` would be called multiple times in a row.

bengal: > `dispose()` in glib might always be called multiple times. In practice, that is not the case for NetworkManager, or here. But we shold write our dispose() code in that manner.

Yeah, I was just wondering if you really meant 'reentrant' (i.e. thread-safe) or just that the function can be called multiple times.

> That basically just means to use `g_clear_object()/nm_clear_g_free()` instead of `g_object_unref()/g_free()`. Don't leave dangling pointers, and assume `dispose()` would be called multiple times in a row.

I think the only change needed is to avoid calling ovsdb_disconnect() multiple times.

thom311: and
```
     g_string_free (priv->input, TRUE);
     priv->input = NULL;
```
g_string_free() asserts that the argument is non-NULL.

And instead of guarding the call to `ovsdb_disconnect()`, the function should also accept to be called any time (and decide itself when there is nothing to do).

thom311:
lgtm

bengal:
Merged as 080f5ee76b037287ca7d104c8a2e4d3df265278e.

PR #298, opened 2019-02-15
[th/c-list-move]

thom311:
Add `nm_c_list_move_*()` and use it.

I'll also use this on another branch, so I figure it's worth adding such helper methods.


bengal:
LGTM

thom311:
thanks. Merged to master.

PR #301, opened 2019-02-22
device: do ARP announcements only after masters have a slave

bengal:
Delay ARP announcements for masters until the first interfaces gets
enslaved. There is no point in doing it before as the ARP packets
would be dropped in most cases; also, if the first slave is added when
we already started announcing, the MAC of the master is going to
change and so the remaining ARPs will have a wrong "sender mac
address" field.

https://bugzilla.redhat.com/show_bug.cgi?id=1678796


thom311:
lgtm

thom311:
merged to master

PR #302, opened 2019-02-25
wifi-p2p: fix failed assertion when releasing supplicant interface

bengal:
supplicant_interfaces_release() can be called during a state change (for example by device_state_changed()) and so it can't trigger another state change. Move the state change to callers, only when actually needed.

Fixes the following failed assertion:
```
 <debug> device[0x11dfec0] (p2p-dev-wlp4s0): P2P: Releasing WPA supplicant interface.
 <debug> supplicant: setting WFD IEs for P2P operation
 (../src/devices/nm-device.c:14769):_set_state_full: runtime check failed: (priv->in_state_changed == FALSE)
 <info>  device (p2p-dev-wlp4s0): state change: unmanaged -> unavailable (reason 'supplicant-failed', sys-iface-state: 'external')
 <debug> device[0x11dfec0] (p2p-dev-wlp4s0): add_pending_action (1): 'waiting-for-supplicant'
```
Fixes: 27bc2cb22a6bdd3dc0c1ad2a67464939dcf53049

### DIFF
--- a/src/devices/wifi/nm-device-wifi-p2p.c
+++ b/src/devices/wifi/nm-device-wifi-p2p.c
@@ -705,11 +705,10 @@ supplicant_iface_state_cb (NMSupplicantInterface *iface,
 			_set_is_waiting_for_supplicant (self, FALSE);
 		break;
 	case NM_SUPPLICANT_INTERFACE_STATE_DOWN:
+		supplicant_interfaces_release (self, TRUE);
 		nm_device_queue_recheck_available (device,
 		                                   NM_DEVICE_STATE_REASON_SUPPLICANT_AVAILABLE,
 		                                   NM_DEVICE_STATE_REASON_SUPPLICANT_FAILED);
-
-		supplicant_interfaces_release (self, TRUE);
 		break;
 	default:
 		break;
@@ -931,19 +930,13 @@ supplicant_interfaces_release (NMDeviceWifiP2P *self, gboolean set_is_waiting)
 
 	if (priv->mgmt_iface) {
 		_LOGD (LOGD_DEVICE | LOGD_WIFI, "P2P: Releasing WPA supplicant interface.");
-
 		nm_supplicant_manager_set_wfd_ies (priv->sup_mgr, NULL);
-
 		g_signal_handlers_disconnect_by_data (priv->mgmt_iface, self);
-
 		g_clear_object (&priv->mgmt_iface);
+		nm_clear_g_source (&priv->sup_timeout_id);
 	}
 
 	supplicant_group_interface_release (self);
-
-	nm_device_state_changed (NM_DEVICE (self),
-	                         NM_DEVICE_STATE_UNAVAILABLE,
-	                         NM_DEVICE_STATE_REASON_SUPPLICANT_FAILED);
 
 	if (set_is_waiting)
 		_set_is_waiting_for_supplicant (self, TRUE);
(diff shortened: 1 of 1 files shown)


thom311:
shouldn't this state-change only be done, if the current start > UNAVAILABLE?

bengal:
> shouldn't this state-change only be done, if the current start > UNAVAILABLE?

Thinking again about it, it seems to me that we shouldn't set state UNAVAILABLE directly but instead just call nm_device_queue_recheck_available() that will update the state asynchronously. 
supplicant_iface_state_cb() already does that and nm_device_wifi_p2p_set_mgmt_iface() needs to do it also when `!iface`, like in:

https://gist.github.com/bengal/bef223bcd98c5c8292206d8f6fc7aa3d

What do you think? @benzea ?

thom311:
this sounds good (better) to me

benzea:
That doesn't sound good to me unfortunately :)

AFAICT, we may have a state changed queued at the point that this call happens. If we call recheck_available, then this queued state change will *not* be removed and we may transition into it before the recheck is done.

I would prefer to just call `nm_device_queue_recheck_available`, but I don't think it is safe.

bengal:
> this queued state change will _not_ be removed and we may transition into it before the recheck is done.

Does this cause any trouble? If a state change was scheduled it will be run but in the end recheck_available() will bring the device in the correct state anyway, right?

If it's necessary to clear the pending state changes before scheduling a recheck, we could add a boolean parameter to nm_device_queue_recheck_available(). Or in alternative, restore the explicit nm_device_state_changed() like the initial patch did.

benzea:
> Does this cause any trouble?

I believe it may right now. Because `device_state_changed` will assert that `priv->mgmt_iface != NULL` if `new_state > NM_DEVICE_STATE_UNAVAILABLE`. And that assertion seems sensible to me to be honest.

About `UNAVAILABLE`, I think that is the lowest state the device can ever be in realistically. But it is true that we should never transition to a higher state there.

thom311:
I think, set-state and recheck-available should integrate better.

Meaning: `_set_state_full()` should: "if the new state is > UNAVAILABLE && recheck-available-is-queued", then as first thing the device should recheck availability. If the check determines that the device is available, then it proceeds setting the new state.

if the check determines the device is UNAVAILABLE, then clearly we want eventually reach unavailable. That could mean, that the device ignores the new start and just goes to UNAVAILABLE right away.

That would avoid this problem.

However, I am not sure that such a transition directly to UNAVAILABLE is right. The device cannot just skip states. E.g. the device cannot go from ACTIVATED directly to UNAVAILABLE, instead the device must go through all intermediate states, like DISCONNECTING and DISCONNECTED. That means, asserting that `priv->mgmt_iface != NULL` if `new_state > NM_DEVICE_STATE_UNAVAILABLE` is wrong.

benzea:
Hm, right, I think right the P2P code in NetworkManager does not guarantee that all those states are reached when switching to UNAVAILABLE. So, that is something that likely needs fixing.

I agree that such a modified approach to state changing would fix the potential race conditions.

Changing the assertion seems fair to me. I guess there might be a need to add guards elsewhere to ensure we never try to operate on a non-existing `mgmt_iface`.

bengal:
I updated the patch, how does it look now?

thom311:
lgtm   (it's not easy to understand all the implications)

benzea:
Yeah, looks good to me.

The only potential issue I see right now is that `act_stage2_config` does not have a check for `mgmt_iface` yet (in the same way that `act_stage1_prepare` has)?

bengal:
Thanks, fixed and merged (d2e95856e9b8613fee58384a554c423f83676a8b).

PR #303, opened 2019-02-28
build: replace intltool with gettext

lkundrak:
Gettext, since version 0.18, provides infrastructure equivalent to
intltool (generating po/Makefile and translating the .xml policy).

Also, it seems to work better in that it actually honors MSGFMT_OPTS
that are determined by the configure script. That one typically ends up
set to "-c", meaning "check". Without it the bugs in the .po files are
easy to overlook.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -192,6 +192,7 @@ test-*.trs
 /libnm/tests/test-secret-agent
 
 /m4/codeset.m4
+/m4/fcntl-o.m4
 /m4/gettext.m4
 /m4/glibc2.m4
 /m4/glibc21.m4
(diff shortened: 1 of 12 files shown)


thom311:
CI does not pass. Setting "needs-work" label.

thom311:
see also https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/96

thom311:
this branch is obsolete by https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/96.
They are mostly identical anyway. Closing.

thom311:
the difference can be seen here: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/96#note_676936

thom311:
this was eventually fixed with commit 19a6affd3ad40ce7fd7ca6201f20ddd89e99949a and [!1094](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1094).

PR #304, opened 2019-03-01
[th/wireguard-import]

thom311:
Implement `nmcli connection import type wireguard file "$FILE"` and various other WireGuard related fixes.

Note that this changes libnm API that we released as 1.16-rc1. Now is still the time.

### DIFF
--- a/clients/common/nm-vpn-helpers.c
+++ b/clients/common/nm-vpn-helpers.c
@@ -25,7 +25,13 @@
 
 #include "nm-vpn-helpers.h"
 
+#include <arpa/inet.h>
+#include <net/if.h>
+
+#include "nm-client-utils.h"
 #include "nm-utils.h"
+#include "nm-utils/nm-io-utils.h"
+#include "nm-utils/nm-secret-utils.h"
 
 /*****************************************************************************/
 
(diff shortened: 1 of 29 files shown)


clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,522 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+
+			if (line_match (line, "MTU", &matched_key, &value)) {
+				i64 = _nm_utils_ascii_str_to_int64 (value, 0, 0, G_MAXUINT32, -1);
+				if (i64 == -1)
+					goto fail_invalid_value;
+
+				/* TODO: not yet implemented. Ignore the value.
+				 *
+				 * wg-quick accepts the "MTU" value, but it also fetches routes to
+				 * autodetect it. */
```
zx2c4: NetworkManager is in a uniquely better position to manage this than wg-quick. Namely, it knows about and monitors all of the links already. Using the existing link monitoring, resolve the endpoint's interface, subtract 80, and keep that set.

clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,522 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+				continue;
+			}
+
+			if (line_match (line, "DNS", &matched_key, &value)) {
+				char addr_s[NM_CONST_MAX (INET_ADDRSTRLEN, INET6_ADDRSTRLEN)];
+				GPtrArray **p_data_dns;
+				NMIPAddr addr_bin;
+				int addr_family;
+
+				if (!nm_utils_parse_inaddr_bin (AF_UNSPEC,
```
zx2c4: Note that fields like Address=, DNS=, and AllowedIPs= can take multiple entries on a single line, separated by a comma.

clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,522 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+
+			if (line_match (line, "AllowedIPs", &matched_key, &value)) {
+
+				while (value[0] != '\0') {
+					const char *entry;
+					char *str;
+
+					entry = value;
+					str = strchrnul (value, ',');
+					if (str[0] == ',') {
```
zx2c4: Looks like you're doing the comma parsing here. Remember to have this in DNS= and Address= too.

clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,522 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+		}
+
+		if (data_table == _TABLE_AUTO) {
+			/* TODO: not implemented. */
+		} else if (data_table == _TABLE_OFF) {
+			/* TODO: not implemented. Well, currently we don't support adding direct routes for the
+			 * allowed-ips ranges. So, it implicitly is implemented.  */
+		} else {
+			/* TODO: setting NM_SETTING_IP_CONFIG_ROUTE_TABLE to a routing table is not exactly the same as
+			 * what wg-quick does. */
```
zx2c4: What is that doing, btw?

clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,522 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+			for (i = 0; i < data_addr->len; i++)
+				nm_setting_ip_config_add_address (s_ip, data_addr->pdata[i]);
+		}
+		if (data_dns) {
+			for (i = 0; i < data_dns->len; i++)
+				nm_setting_ip_config_add_dns (s_ip, data_dns->pdata[i]);
+		}
+
+		if (data_table == _TABLE_AUTO) {
+			/* TODO: not implemented. */
```
zx2c4: In order for NetworkManager to actually be useful to people using WireGuard, you'll probably need to implement some variety of this. At least solve the general question: 0.0.0.0/0 is in AllowedIPs. You now know what they expect. How will you configure their routes? Policy routing? Explicit routes? Sky's the limit, but you have to do something. The table parameter, I think, is a particularly good way of doing it, but that's only one method. (<https://www.wireguard.com/netns/> describes a few.)

libnm-core/nm-keyfile.c
```diff
@@ -2929,19 +2929,18 @@ _read_setting_wireguard_peer (KeyfileReaderInfo *info)
...
 	}
-	nm_wireguard_peer_set_public_key (peer, cstr);
+	nm_wireguard_peer_set_public_key (peer, cstr, TRUE);
 	nm_clear_g_free (&str);
 
 	key = NM_WIREGUARD_PEER_ATTR_PRESHARED_KEY;
 	str = nm_keyfile_plugin_kf_get_string (info->keyfile, info->group, key, NULL);
 	if (str) {
-		if (!_nm_utils_wireguard_decode_key (str, NM_WIREGUARD_SYMMETRIC_KEY_LEN, NULL)) {
+		if (!nm_wireguard_peer_set_preshared_key (peer, str, FALSE)) {
```
zx2c4: In case the verdict wasn't clear from our email/IRC discussions, on input, consider the all zero key as "null" -- an "unsetting" action. On output, consider any stored all zero key (say, from the kernel) as unset as well. But don't try to validate and reject any of these keys.

shared/nm-utils/nm-secret-utils.c
```diff
@@ -132,3 +133,31 @@ nm_secret_buf_to_gbytes_take (NMSecretBuf *secret, gssize actual_len)
...
+ * Returns: whether all bytes are zero.
+ */
+gboolean
+nm_utils_memeqzero_secret (gconstpointer data, gsize length)
+{
+	const guint8 *const key = data;
+	volatile guint8 acc = 0;
+	gsize i;
+
+	/* Based on key_is_zero() by Jason A. Donenfeld <Jason@zx2c4.com>. */
```
zx2c4: As mentioned via email, IANAL, but usually in my experience the right way (or at least, the common way) of crediting code that you take from some GPL-2.0 or LGPL-2.1 sources isn't a stray comment like this, but to just take the precise copyright line from the header and put that in the header of this file. On the other hand, it's not as though this is a particularly novel function, so whatever floats your boat...

src/platform/nm-linux-platform.c
```diff
@@ -2019,8 +2019,21 @@ _wireguard_update_from_peers_nla (CList *peers,
...
-		if (tb[WGPEER_A_LAST_HANDSHAKE_TIME])
-			nla_memcpy (&peer_c->data.last_handshake_time, tb[WGPEER_A_LAST_HANDSHAKE_TIME], sizeof (peer_c->data.last_handshake_time));
+		if (tb[WGPEER_A_LAST_HANDSHAKE_TIME]) {
+			if (nla_len (tb[WGPEER_A_LAST_HANDSHAKE_TIME]) >= sizeof (peer_c->data.last_handshake_time))
+				nla_memcpy (&peer_c->data.last_handshake_time, tb[WGPEER_A_LAST_HANDSHAKE_TIME], sizeof (peer_c->data.last_handshake_time));
+			else if (   sizeof (struct timespec) < sizeof (NMPTimespec64)
+			         && nla_len (tb[WGPEER_A_LAST_HANDSHAKE_TIME]) >= sizeof (struct timespec)) {
+				struct timespec ts;
+
+				/* older versions used "struct timespec". */
```
zx2c4: Personally, I'd prefer you don't leave support for older code around.

clients/common/nm-vpn-helpers.c
thom311: Let me quote the documentation (just to see whether that description is comprehensible):

```
»···/**
»··· * NMSettingIPConfig:route-table:
»··· *
»··· * Enable policy routing (source routing) and set the routing table used when adding routes.
»··· *
»··· * This affects all routes, including device-routes, IPv4LL, DHCP, SLAAC, default-routes
»··· * and static routes. But note that static routes can individually overwrite the setting
»··· * by explicitly specifying a non-zero routing table.
»··· *
»··· * If the table setting is left at zero, it is eligible to be overwritten via global
»··· * configuration. If the property is zero even after applying the global configuration
»··· * value, policy routing is disabled for the address family of this connection.
»··· *
»··· * Policy routing disabled means that NetworkManager will add all routes to the main
»··· * table (except static routes that explicitly configure a different table). Additionally,
»··· * NetworkManager will not delete any extraneous routes from tables except the main table.
»··· * This is to preserve backward compatibility for users who manage routing tables outside
»··· * of NetworkManager.
»··· *
»··· * Since: 1.10
»··· **/
```

src/platform/nm-linux-platform.c
thom311: OK, dropped (also, because the value is unused and because the code only mattered on 32bit). 

While at it, I think it's inconvenient to expose timestamps on netlink in `time(2)` scale, which is susceptible to clock changes. 
 I think `CLOCK_BOOTTIME` scale would be better, otherwise there is an unresolvable race computing `ago()`.

clients/common/nm-vpn-helpers.c
thom311: I don't think that link-monitoring suffices, because it's non-obvious which is the "entpoint's interface". It would require doing what`wg-quick` does, which looks up the interface (or the MTU) via `ip route get`. Basically, on every tiny change of IP configuration, NetworkManager would need to call `ip route get` for each endpoint, which seems also to have issues to scale. 

On #305 the MTU will simply be configurable, but in absence of an explicit value, it's left unset. NetworkManager also does that for other "ip-tunnel" types, where the same issue exists, and if this would be improved, it should be improved for other tunnel types as well.

Do you know how does systemd-networkd handles this?

libnm-core/nm-keyfile.c
thom311: ok

clients/common/nm-vpn-helpers.c
thom311: fixed

clients/common/nm-vpn-helpers.c
thom311: yes, fixed.

clients/common/nm-vpn-helpers.c
thom311: On #305, the AllowedIPs ranges will be added as routes. So that part will be covered.

Handling of routing-loops ("/0") is a related but separate issue. Fixing that problem is not restricted to WireGuard but affects also "ip-tunnels" (e.g. vxlan). It needs a general solution in NetworkManager, which won't happen before 1.16.0.

The *(non-)* solution for the moment is to let the user solve this manually. For example,
- by adding static routes to the endpoints on (the profile of) the interface with the default-route. This can be done by configuring the other profile in NetworkManager, or via script.
- or by (manually) setting up suitable routing policy rules (via a script). Support for configuring rules is work in progress and will be added soon.

zx2c4: > It needs a general solution in NetworkManager, which won't happen before 1.16.0.

Sounds like NM will not be useful for WireGuard, which is unfortunate. Hopefully GUIs don't arrive until 1.17, so that I don't have to deal with as many confused users on a regular basis or get in an unfortunate habit of telling people, "dont use NM; it's broken" or something. Certainly I don't want to start propagating alternative workarounds, since those tend to stick around, on blogs and in configs, after the deficiency is addressed and confuse people even more.

What's the purpose of putting this off for later?

> (e.g. vxlan)

People generally have different expectations of vxlan.

clients/common/nm-vpn-helpers.c
zx2c4: Wording makes sense to me, though I don't quite see how this is different from wg-quick's behavior, except the global override business.

clients/common/nm-vpn-helpers.c
zx2c4: > Basically, on every tiny change of IP configuration, NetworkManager would need to call ip route get for each endpoint, which seems also to have issues to scale.

No, the routing table is supposed to be extremely fast, since it's queried for incoming packets. IIRC, a cache was recently removed, even, because the main routing table implementation was considered sufficiently fast.

shared/nm-utils/nm-secret-utils.c
```diff
@@ -17,6 +17,7 @@
  * Boston, MA 02110-1301 USA.
  *
  * (C) Copyright 2018 Red Hat, Inc.
+ * (C) Copyright 2019 Jason A. Donenfeld <Jason@zx2c4.com>
```
zx2c4: The file you took the function from had:

```
 * Copyright (C) 2015-2019 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
```

IANAL and don't know particulars on if it's ever appropriate to change dates, but in my experience, I usually just copy the copyright line wholesale.

clients/common/nm-vpn-helpers.c
thom311: yeah, it's not too different. I updated the comment.

shared/nm-utils/nm-secret-utils.c
thom311: OK, adjusted. Hope it's right (only adjusted for style -- all our such lines have the `(C)` after Copyright and a space between the year).

clients/common/nm-vpn-helpers.c
thom311: With #305 merged, I think what we want to do about MTU is now implemented. The import-code of this branch is also adjusted.

We may improve futher, but not as part of this pull-request. We should open a separate issue, if there are concrete suggestions what to do exactly.

clients/common/nm-vpn-helpers.c
thom311: > What's the purpose of putting this off for later?

That it takes consideration (time+effort) to get this right.

We'll just document NM's WireGuard support in 1.16.0 as experimental.

zx2c4: Or just put the consideration in now before 1.16 drops? Policy routing isn't really that hard.

shared/nm-utils/nm-secret-utils.c
zx2c4: You really don't this now.

thom311: The comment is not for Copyright reasons, but to understand where the code originates from. For what it's worth, 2 lines of the functions are non-obvious to me, and it seems helpful to be able to find the original implementation, to possibly understand why these lines are the way they are.

Anyway, it would be hard to argue that any comment is ever really "needed". Dropped.

bengal:
lgtm

libnm-core/nm-utils.c
```diff
@@ -6738,3 +6740,4 @@ _nm_utils_wireguard_normalize_key (const char *base64_key,
 	nm_explicit_bzero (buf, required_key_len);
 	return TRUE;
 }
+
```
lkundrak: Why this extra line break?

libnm-core/nm-utils.c
```diff
@@ -6673,21 +6673,23 @@ nm_utils_version (void)
 /*****************************************************************************/
 
 /**
- * _nm_utils_wireguard_decode_key:
+ * nm_utils_base64secret_decode:
```
lkundrak: Why "secret"? Is anything the procedure does specific to secrets or keys?

libnm-core/nm-utils.c
thom311: bug. Will fix.

clients/common/nm-vpn-helpers.c
```diff
@@ -247,3 +253,556 @@ nm_vpn_openconnect_authenticate_helper (const char *host,
...
+	if (str[0] == ',') {
+		str[0] = '\0';
+		*line_remainder = &str[1];
+	} else
+		*line_remainder = str;
+	return TRUE;
+}
+
+NMConnection *
+nm_vpn_wireguard_import (const char *filename,
```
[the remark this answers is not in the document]
thom311: done

thom311:
merged to master.

PR #305, opened 2019-03-02
[th/wireguard-routes]

thom311:
Branch to automatically add direct routes to the AllowedIPs ranges of the WireGuard peers.

And, I couldn't refrain from refactoring NMDevice a bit :)


src/devices/nm-device-private.h
```diff
@@ -75,11 +75,12 @@ void nm_device_set_firmware_missing (NMDevice *self, gboolean missing);
 void nm_device_activate_schedule_stage1_device_prepare (NMDevice *device);
 void nm_device_activate_schedule_stage2_device_config (NMDevice *device);
 
-void nm_device_activate_schedule_ip4_config_result(NMDevice *device, NMIP4Config *config);
-void nm_device_activate_schedule_ip4_config_timeout (NMDevice *device);
+void nm_device_activate_schedule_ip_config_result (NMDevice *device,
+                                                   int addr_family,
+                                                   gpointer config /* (NMIPConfig *) */);
```
bengal: Why not a NMIPConfig * argument instead of the void pointer, and explicit casts where needed?

thom311: because most callers will actually be required to cast. Anyway, yes, maybe that's better. Fixing...

lkundrak:
LGTM

thom311:
Thanks. merged to master.

PR #306, opened 2019-03-04
[th/routing-rule-pt1]

thom311:
first part supporting routing rules.

Only add the platform caching part.


bengal:
LGTM

thom311:
merged to master, because all CI is green and @bengal  gave it thumbs up.

Also, there is not much time if this work is going to make it into 1.16.2. So better to merge a non-perfect work early, then aiming for the perfect solution with only little time left.

PR #308, opened 2019-03-05
[lr/goodbye-libnm-glib-❤] all: goodbye libnm-glib

lkundrak:
This removes libnm-glib, libnm-glib-vpn, and libnm-util for good.
The it has been replaced with libnm since NetworkManager 1.0, disabled
by default since 1.12 and no up-to-date distributions ship it for years
now.

Removing the libraries allows us to:

* Remove the horrible hacks that were in place to deal with accidental use
  of both the new and old library in a single process.
* Relief the translators of maintenance burden of similar yet different
  strings.
* Get rid of known bad code without chances of ever getting fixed
  (libnm-glib/nm-object.c and libnm-glib/nm-object-cache.c)
* Generally lower the footprint of the releases and our workspace

If there are some really really legacy users; they can just build
libnm-glib and friends from the NetworkManager-1.16 distribution. The
D-Bus API is stable and old libnm-glib will keep working forever.


thom311:
let's hold this off until after `1.16.0` is released (which is a matter of days).

(Because it's helpful to compare `master` with `nm-1-16` during 1-16-rc time, and see a reasonably small diff).

PR #309, opened 2019-03-05
platform/linux: fix detection of IFA_FLAGS support

lkundrak:
The condition got accidentally reversed, which means we're always
undecided and thus wrongly assuming support and never being able to set
any addresses.

This would bother the few that are struck with 3.4 android kernels. Very
few indeed, given this got unnoticed since 1.10.

Fixes: 8670aacc7ce ('platform: cleanup detecting kernel support for IFA_FLAGS and IPv6LL')


thom311:
lgtm

How did you find this? Just by noticing the log message about "unable to detect kernel support"?

lkundrak:
> How did you find this? Just by noticing the log message about "unable to detect kernel support"?

I've done terrible terrible things. Don't ask.

PR #310, opened 2019-03-07
[lr/rpm-madness] Some contrib/rpm tidy-ups

lkundrak:


thom311:
lgtm.

Keep in mind, I intend to sync the spec-files of Fedora 30 and Rawhide to this. I think that's desirable, but it does revert some changes that were done there intentionally.

Also, it reverts https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/93

PR #312, opened 2019-03-07
wwan/device-modem: don't enter available state until registered

lkundrak:
Based on Ubuntu's "Modify NMDeviceModem's available logic" patch by
Tony Espy <espy@canonical.com>. The original commit message:

This patch modifies NMDeviceModem's available logic such that the device
is only considered available if the modem_state is
>= NM_MODEM_STATE_REGISTERED.  NMDevice defines 'available' as meaning the
device is in such a state that it can be activated.  This change prevents
NM from trying to activate a modem which is not yet ready to be activated.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/1445080

### DIFF
--- a/src/devices/wwan/nm-device-modem.c
+++ b/src/devices/wwan/nm-device-modem.c
@@ -473,11 +473,6 @@ check_connection_available (NMDevice *device,
 	}
 
 	state = nm_modem_get_state (priv->modem);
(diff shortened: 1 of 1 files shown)


thom311:
looks not wrong to me. Dunno, it's not easy to understand the implications.

Can you wait for a CI run before merging?

PR #313, opened 2019-03-08
wwan/ofono: pass the right argument to manager proxy callback

lkundrak:
Otherwise it will be dereferencing NULL when invoked.

Fixes: 58712c95464a ('ofono: take D-Bus proxy for ConnectionManager asynchronously')

### DIFF
--- a/src/devices/wwan/nm-modem-ofono.c
+++ b/src/devices/wwan/nm-modem-ofono.c
@@ -651,7 +651,7 @@ handle_connman_iface (NMModemOfono *self, gboolean found)
 		                          OFONO_DBUS_INTERFACE_CONNECTION_MANAGER,
 		                          priv->connman_proxy_cancellable,
 		                          _connman_proxy_new_cb,
-		                          NULL);
+		                          self);
 	}
 }
 


thom311:
lgtm

---

In general, could we in the pull-requests (or the commit message) give more information about the motivation/origin of the change?

How did you discover it? Were you testing ofono? Did you just review the code? Did somebody tell you about the crash personally? Is there an upstream bug report? etc.

I think that is helpful to understand the motivation for a patch, and such information should be tracked. In this example, the "motivation" is clear, but the "origin" / "backstory" would be interesting, possibly to understand who is impacted by this (e.g. I am curious which distros use recent NM with ofono)

Of course, if you found the bug while doing code review, there is not much to say -- just drop one line "found via code-review/testing by myself".

lkundrak:
> In general, could we in the pull-requests (or the commit message) give more information about the motivation/origin of the change?
> 
> How did you discover it? Were you testing ofono? Did you just review the code? Did somebody tell you about the crash personally? Is there an upstream bug report? etc.
> 
> I think that is helpful to understand the motivation for a patch, and such information should be tracked. In this example, the "motivation" is clear, but the "origin" / "backstory" would be interesting, possibly to understand who is impacted by this (e.g. I am curious which distros use recent NM with ofono)
> 
> Of course, if you found the bug while doing code review, there is not much to say -- just drop one line "found via code-review/testing by myself".

I'd prefer not to comment on circumstances of discovering a bug there, because it could easily send the wrong message about their relevance. In fact I don't have a slightest clue about how relevant oFono in general is nor whether anyone is actually using it or maintaining the code.

PR #314, opened 2019-03-09
restore IPv6 addresses when the link comes up

bengal:
When the link goes down the kernel removes IPv6 addresses from the interface. In update_ext_ip_config() we detect that addresses were removed externally and drop them from various internal configurations. Don't do that if the link is down so that those addresses will be restored again on link up.

https://bugzilla.redhat.com/show_bug.cgi?id=1548237


thom311:
lgtm

PR #315, opened 2019-03-12
ovs-port: dissociate the link from the interface device on delete

lkundrak:
Open vSwitch is the special kid on the block -- it likes to be in charge of
the link lifetime and so we shouldn't be. This means that we shouldn't be
attempting to remove the link: we'd just (gracefully) fail anyways.

More importantly, this also means that we shouldn't care if we see the link
go away. We may already be activating another connection and shouldn't alter
the device state when OpenVSwitch decides to drop the old link.

https://bugzilla.redhat.com/show_bug.cgi?id=1543557

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -14219,6 +14219,7 @@ static void
 _cleanup_generic_post (NMDevice *self, CleanupType cleanup_type)
 {
 	NMDevicePrivate *priv = NM_DEVICE_GET_PRIVATE (self);
+	int ifindex;
 
 	priv->v4_commit_first_time = TRUE;
 	priv->v6_commit_first_time = TRUE;
@@ -14283,7 +14284,9 @@ _cleanup_generic_post (NMDevice *self, CleanupType cleanup_type)
 		/* Check if the device was deactivated, and if so, delete_link.
 		 * Don't call delete_link synchronously because we are currently
 		 * handling a state change -- which is not reentrant. */
-		delete_on_deactivate_check_and_schedule (self, nm_device_get_ip_ifindex (self));
+		ifindex = nm_device_get_ip_ifindex (self);
+		if (ifindex > 0)
+			delete_on_deactivate_check_and_schedule (self, ifindex);
 	}
 
 	/* ip_iface should be cleared after flushing all routes and addresses, since
--- a/src/devices/ovs/nm-device-ovs-port.c
+++ b/src/devices/ovs/nm-device-ovs-port.c
@@ -139,6 +139,13 @@ del_iface_cb (GError *error, gpointer user_data)
 static void
 release_slave (NMDevice *device, NMDevice *slave, gboolean configure)
 {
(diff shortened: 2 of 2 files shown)


src/devices/nm-device.c
```diff
@@ -14283,7 +14284,9 @@ _cleanup_generic_post (NMDevice *self, CleanupType cleanup_type)
 		/* Check if the device was deactivated, and if so, delete_link.
 		 * Don't call delete_link synchronously because we are currently
 		 * handling a state change -- which is not reentrant. */
-		delete_on_deactivate_check_and_schedule (self, nm_device_get_ip_ifindex (self));
+		ifindex = nm_device_get_ip_ifindex (self);
+		if (ifindex > 0)
+			delete_on_deactivate_check_and_schedule (self, ifindex);
```
thom311: wouldn't it be nicer if `delete_on_deactivate_check_and_schedule()` had this check for `ifindex > 0`? It already starts with a preamble of conditions when to do nothing.

Otherwise, there should be a `g_return_if_fail (ifindex > 0);` in `delete_on_deactivate_check_and_schedule()`.

src/devices/ovs/nm-device-ovs-port.c
```diff
@@ -139,6 +139,13 @@ del_iface_cb (GError *error, gpointer user_data)
 static void
 release_slave (NMDevice *device, NMDevice *slave, gboolean configure)
 {
+	/* Dissociate the platform link from the device -- openvswitch is going
+	 * to remove the link, not us. We shouldn't care about whatever happen to it.
+	 * In particular, we don't want to alter the NMDevice state if we see the
+	 * link go away, because the device may already have another activation
+	 * enqueued. */
+	nm_device_update_from_platform_link (slave, NULL);
```
thom311: is this really right? Our `NMDevice` instances are usually associated with the netdev device (i.e. the ifindex). Yes, not always, for example bluetooth, WWan and OVS device can exist without an underlying netdev device. But often have devices an `ip_ifindex` that links to the underlying kernel link. For example, we also expose this on D-Bus (as `IpInterface` property).

This means `We shouldn't care about whatever happen to it.` is not true. We do care, as long as there is a link, various places want to know about this association (e.g. D-Bus properties).

I don't think you should clear the ip-ifindex before platform says the link it's really gone. Also, while we may call `nm_device_update_from_platform_link()` from multiple places, I think it always should pass there the `NMPlatformLink` information that is in the platform cache at that point. We shouldn't lie to `NMDevice` about that.

If you don't want to do something with this link, then don't do it (by setting a flag on the device to not doing it). But don't tell NMDevice that the interface is gone, when it is not.

thom311:
I reverted the patch for the moment. It causes a crash and breaks CI. See for example https://beaker.engineering.redhat.com/recipes/6661811#task90117952,task90117953

vbenes:
crash here:
https://faf.lab.eng.brq.redhat.com/faf//reports/10985/

PR #316, opened 2019-03-14
fg/checkpoint devices rh1578335

fgiudici:
When a checkpoint is created, if the sofware device is destroyed, avoid listing it in the "Devices" property of the Checkpoint object.
https://bugzilla.redhat.com/show_bug.cgi?id=1578335

### DIFF
--- a/src/nm-checkpoint.c
+++ b/src/nm-checkpoint.c
@@ -260,6 +260,11 @@ restore_and_activate_connection (NMCheckpoint *self,
 		need_activation = TRUE;
 	}
 
+	if (!dev_checkpoint->device) {
+		_LOGD ("rollback: device cannot be restored");
+		return FALSE;
+	}
+
 	if (need_activation) {
 		_LOGD ("rollback: reactivating connection %s",
 		       nm_settings_connection_get_uuid (connection));
--- a/src/nm-checkpoint.c
+++ b/src/nm-checkpoint.c
@@ -39,6 +39,7 @@
 typedef struct {
 	char *original_dev_path;
 	char *original_dev_name;
+	NMDeviceType dev_type;
 	NMDevice *device;
 	NMConnection *applied_connection;
 	NMConnection *settings_connection;
@@ -61,6 +62,7 @@ NM_GOBJECT_PROPERTIES_DEFINE (NMCheckpoint,
 struct _NMCheckpointPrivate {
 	/* properties */
 	GHashTable *devices;
+	GPtrArray *removed_devices;
 	gint64 created_at_ms;
 	guint32 rollback_timeout_s;
 	guint timeout_id;
@@ -68,6 +70,7 @@ struct _NMCheckpointPrivate {
 	NMManager *manager;
 	NMCheckpointCreateFlags flags;
 	GHashTable *connection_uuids;
+	gulong dev_removed_id;
 
 	NMCheckpointTimeoutCallback timeout_cb;
 	gpointer timeout_data;
@@ -245,6 +248,15 @@ restore_and_activate_connection (NMCheckpoint *self,
 			g_clear_error (&local_error);
 			return FALSE;
 		}
+
+		/* If the device is software, a brand new NMDevice may have been created */
+		if (   dev_checkpoint->is_software
+		    && !dev_checkpoint->device) {
+			dev_checkpoint->device = nm_manager_get_device (priv->manager,
+			                                                dev_checkpoint->original_dev_name,
+			                                                dev_checkpoint->dev_type);
+			nm_g_object_ref (dev_checkpoint->device);
+		}
 		need_activation = TRUE;
 	}
 
@@ -294,9 +306,30 @@ nm_checkpoint_rollback (NMCheckpoint *self)
 	GHashTableIter iter;
 	NMDevice *device;
 	GVariantBuilder builder;
+	uint i;
 
 	_LOGI ("rollback of %s", nm_dbus_object_get_path (NM_DBUS_OBJECT (self)));
 	 g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{su}"));
+
+	/* Start creating removed devices (if any and if possible) */
+	if (priv->removed_devices) {
+		for (i = 0; i < priv->removed_devices->len; i++) {
+			guint32 result = NM_ROLLBACK_RESULT_OK;
+
+			dev_checkpoint = priv->removed_devices->pdata[i];
+			_LOGD ("rollback: restoring removed device %s (state %d, realized %d, explicitly unmanaged %d)",
+			       dev_checkpoint->original_dev_name,
+			       (int) dev_checkpoint->state,
+			       dev_checkpoint->realized,
+			       dev_checkpoint->unmanaged_explicit);
+
+			if (dev_checkpoint->applied_connection) {
+				if (!restore_and_activate_connection (self, dev_checkpoint))
+					result = NM_ROLLBACK_RESULT_ERR_FAILED;
+			}
+			g_variant_builder_add (&builder, "{su}", dev_checkpoint->original_dev_path, result);
+		}
+	}
 
 	/* Start rolling-back each device */
 	g_hash_table_iter_init (&iter, priv->devices);
@@ -379,7 +412,6 @@ next_dev:
 	if (NM_FLAGS_HAS (priv->flags, NM_CHECKPOINT_CREATE_FLAG_DELETE_NEW_CONNECTIONS)) {
 		NMSettingsConnection *con;
 		gs_free NMSettingsConnection **list = NULL;
-		guint i;
 
 		g_return_val_if_fail (priv->connection_uuids, NULL);
 		list = nm_settings_get_connections_clone (nm_settings_get (), NULL,
@@ -420,10 +452,51 @@ next_dev:
 }
 
 static void
+device_checkpoint_destroy (gpointer data)
+{
+	DeviceCheckpoint *dev_checkpoint = data;
+
+	nm_clear_g_signal_handler (dev_checkpoint->device, &dev_checkpoint->dev_exported_change_id);
+	g_clear_object (&dev_checkpoint->applied_connection);
+	g_clear_object (&dev_checkpoint->settings_connection);
+	g_clear_object (&dev_checkpoint->device);
+	g_free (dev_checkpoint->original_dev_path);
+	g_free (dev_checkpoint->original_dev_name);
+
+	g_slice_free (DeviceCheckpoint, dev_checkpoint);
+}
+
+static void
+_move_dev_to_removed_devices (NMDevice *device,
+                              NMCheckpoint *checkpoint)
+{
+	NMCheckpointPrivate *priv = NM_CHECKPOINT_GET_PRIVATE (checkpoint);
+	DeviceCheckpoint *dev_checkpoint;
+
+	g_return_if_fail (device);
+
+	dev_checkpoint = g_hash_table_lookup (priv->devices, device);
+	if (!dev_checkpoint)
+		return;
+
+	g_hash_table_steal (priv->devices, dev_checkpoint->device);
+	nm_clear_g_signal_handler (dev_checkpoint->device,
+	                           &dev_checkpoint->dev_exported_change_id);
+	g_clear_object (&dev_checkpoint->device);
+
+	if (!priv->removed_devices)
+		priv->removed_devices = g_ptr_array_new_with_free_func ((GDestroyNotify) device_checkpoint_destroy);
+	g_ptr_array_add (priv->removed_devices, dev_checkpoint);
+
+	_notify (checkpoint, PROP_DEVICES);
+}
+
+static void
 _dev_exported_changed (NMDBusObject *obj,
                        NMCheckpoint *checkpoint)
 {
-	_notify (checkpoint, PROP_DEVICES);
+
+	_move_dev_to_removed_devices (NM_DEVICE (obj), checkpoint);
 }
 
 static DeviceCheckpoint *
@@ -444,6 +517,7 @@ device_checkpoint_create (NMCheckpoint *checkpoint, NMDevice *device)
 	dev_checkpoint->device = g_object_ref (device);
 	dev_checkpoint->original_dev_path = g_strdup (path);
 	dev_checkpoint->original_dev_name = g_strdup (nm_device_get_iface (device));
+	dev_checkpoint->dev_type = nm_device_get_device_type (device);
 	dev_checkpoint->state = nm_device_get_state (device);
 	dev_checkpoint->is_software = nm_device_is_software (device);
 	dev_checkpoint->realized = nm_device_is_real (device);
@@ -472,21 +546,6 @@ device_checkpoint_create (NMCheckpoint *checkpoint, NMDevice *device)
 	}
 
 	return dev_checkpoint;
-}
-
-static void
-device_checkpoint_destroy (gpointer data)
-{
-	DeviceCheckpoint *dev_checkpoint = data;
-
-	nm_clear_g_signal_handler (dev_checkpoint->device, &dev_checkpoint->dev_exported_change_id);
-	g_clear_object (&dev_checkpoint->applied_connection);
-	g_clear_object (&dev_checkpoint->settings_connection);
-	g_clear_object (&dev_checkpoint->device);
-	g_free (dev_checkpoint->original_dev_path);
-	g_free (dev_checkpoint->original_dev_name);
-
-	g_slice_free (DeviceCheckpoint, dev_checkpoint);
 }
 
 static gboolean
@@ -586,6 +645,12 @@ nm_checkpoint_init (NMCheckpoint *self)
 	                                       NULL, device_checkpoint_destroy);
 }
 
+static void
+_device_removed (NMManager *manager, NMDevice *device, gpointer user_data)
+{
+	_move_dev_to_removed_devices (device, NM_CHECKPOINT (user_data));
+}
+
 NMCheckpoint *
 nm_checkpoint_new (NMManager *manager, GPtrArray *devices, guint32 rollback_timeout_s,
                    NMCheckpointCreateFlags flags)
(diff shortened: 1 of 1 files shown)


src/nm-checkpoint.c
```diff
@@ -604,7 +617,7 @@ nm_checkpoint_new (NMManager *manager, GPtrArray *devices, guint32 rollback_time
 		 *        a non-existing D-Bus path of a device. */
```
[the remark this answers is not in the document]
fgiudici: Oh right, I forgot... just pushed a fixup commit on top of the branch for it!
Leaving conversation open as we may want to make other changes that will require another update before the branch is ready...

fgiudici:
> OK, this certainly fixes the immediate problem that on D-Bus expose paths that no longer exist. That part is good.
> 
> However, the NMCheckPoint still keeps a reference to the removed device. If a device gets removed, does it still have a purpose to be tracked by the checkpoint? What exactly happens during rollback with this device?

This is the way I initially addressed the issue: dropping the reference to device and let it be destroyed. Than, during the rollback, I realized that I miss some information if the device obj is gone, namely the interface name (for the LOG) and the info that the device was a software one. While I can add extra fields to create a copy of them in the DeviceCheckpoint structure... why just not delay a bit letting the Device obj go? We would save extra fields and duplicating data.
> 
> If the device is not a software device, it's relatively easy to see that not much happens. That raises the question, whether non-software devices should be dropped earlier, instead of leaking them until the checkpoint gets destroyed.

We could do that. But still, we would have to store in DeviceCheckpoint the device name at least. What would be the issue of keeping the Device obj alive a bit more?
> 
> If the device is a software device, rollback proceeds below. If will for example check the device's unmanaged-state and state. None of this makes any sense, if the device instance is no longer tracked by NMManager. At the very least, it seems rollback should not touch devices that are unexported.

You are totally right here. This should be fixed.
> Better maybe even drop the reference to the device instance early.

I was on your page when starting working on this branch... but then I changed my mind. I think it is ok to keep the reference to the Device obj a bit more, I don't see any harm in it... the benefit is to have all the information related to the device handy, without the need to copy them.

thom311:
> Than, during the rollback, I realized that I miss some information if the device obj is gone, namely the interface name (for the LOG) and the info that the device was a software one. While I can add extra fields to create a copy of them in the DeviceCheckpoint structure... why just not delay a bit letting the Device obj go? 

After the device gets removed (from the list of devices in NMManager) most fields become meaningless. Did you actually check that `nm_device_get_iface()` still returns a valid name afterwards? Even if it does by accident, we should not guarantee (nor rely on) that NMDevice's properties still make sense after being removed.

Yes, you need to cache/copy the fields that you might require afterwards.

> I don't see any harm in it

The harm is that the NMDevice instance itself keeps references to all kinds of other objects. It's not easy to understand whether extending the lifetime of these objects (or NMDevice itself) is a problem or not. Assume it is.

fgiudici:
Still missing to drop the reference to the Device objects. WiP.

fgiudici:
While the branch is working and looks "good enough", better to drop the weak reference on the Device objs. After all, we don't really care if the Device obj is destroyed, and we should not rely on that. Let's instead register on the signal of the Device removal (or unexport) to invalidate the Device pointer...

src/nm-checkpoint.c
```diff
@@ -433,6 +463,7 @@ device_checkpoint_create (NMCheckpoint *checkpoint, NMDevice *device)
 	dev_checkpoint->device = g_object_ref (device);
 	dev_checkpoint->original_dev_path = g_strdup (path);
```
thom311: there is actually little point in preserving the `original_dev_path`.

Each path MUST be generated uniquely, and we should never export an `NMDBusObject` (like `NMDevice`) more than once. The D-Bus path ("exported path") is only useful as an ID when communicating with the user (== the D-Bus API). After the NMDevice gets unexported (the first and only time), the path becomes meaningless.

In particular:
```
 next_dev:
          g_variant_builder_add (&builder, "{su}", dev_checkpoint->original_dev_path, result);
```
is wrong.

Either the device is still exported (in which case you can just use `nm_exported_object_get_path()`) -- or it is not and then it is a bug to tell the user about a device that no longer exists.

src/nm-checkpoint.c
```diff
@@ -615,14 +691,17 @@ nm_checkpoint_new (NMManager *manager, GPtrArray *devices, guint32 rollback_time
 	for (i = 0; i < devices->len; i++) {
 		NMDevice *device = devices->pdata[i];
 
-		/* As long as the check point instance exists, it will keep a reference
-		 * to the device also if the device gets removed (by rmmod or by deleting
-		 * a connection profile for a software device). */
+		nm_assert (nm_dbus_object_get_path (NM_DBUS_OBJECT (device)));
 		g_hash_table_insert (priv->devices,
-		                     device,
+		                     g_strdup (nm_dbus_object_get_path (NM_DBUS_OBJECT (device))),
```
thom311: as said above, it's wrong to stick to a D-Bus path after the device gets unexpected. As such, it makes no sense to clone and preserve this path.

You can assume that a D-Bus path is never reused. Hence, you can instead just index the hash by `g_direct_hash()` and the "device".

If you ever want to find a `NMDBusObject` by path, then use `nm_dbus_manager_lookup_object()`. That is a hash lookup by path of all currently exported objects.


You need to keep track in `priv->devices` of devices which are still exported, and those that were but no longer are. For that, you don't need a new artificial key (the path) that preserves longer. Instead:

```
static void
_device_removed (NMManager *manager, NMDevice *device, gpointer user_data)
{
    NMCheckpoint *self = NM_CHECKPOINT (user_data);
    NMCheckpointPrivate *priv = NM_CHECKPOINT_GET_PRIVATE (self);
    DeviceCheckpoint *dev_checkpoint;

    dev_checkpoint = g_hash_table_lookup (priv->devices, device);
    if (!dev_checkpoint)
        return;
    g_hash_table_steal (priv->devices, device);
    nm_clear_g_signal_handler (dev_checkpoint->device,
                                    &dev_checkpoint->dev_exported_change_id);
    g_clear_object (&dev_checkpoint->device);
    
    // MOVE THE DELETED ENTRY TO AN ALTERNATIVE LIST
    if (!priv->deleted_objects)
          priv->deleted_objects = g_ptr_array_new_with_free_fun (...);
    g_ptr_array_add (priv->deleted_objects, dev_checkpoint);

    _notify (self, PROP_DEVICES);
}
```

fgiudici:
> the first patches are all good, and can be merged (IMO)
> 
> The last patch I think needs more work.
> 
> Originally I was leaning towards requiring both subscribing to the device's `NM_DBUS_OBJECT_EXPORTED_CHANGED` and manager's `NM_MANAGER_DEVICE_REMOVED`. To make less assumptions about how these two signals interact. But it's clear that `NM_MANAGER_DEVICE_REMOVED` is always emitted shortly _before_ exporting the device. I would still subscribe to both signals, but in both cases I'd do the same: consider the device as gone. Whichever fires first, wins.

Agreed
> 
> Note in the last patch:
> 
> ```
> +         if (!dev_checkpoint->device) {
> +              if (dev_checkpoint->realized) {
> +                   if (dev_checkpoint->is_software) {
> +                        /* try to recreate software device */
> +                        _LOGD ("rollback: software device removed, will re-create");
> +                        goto activate;
> +                   }
> +                   _LOGD ("rollback: device is not present anymore");
> +                   result = NM_ROLLBACK_RESULT_ERR_FAILED;
> +              }
> +              goto next_dev;
> ```
> 
> is wrong. `next_dev` will append the exported path to the GVariant. But the path is invalid and must no longer be used. That is part of what the bug is about -- that paths are exported on D-Bus that no longer exist.

Your reasoning sounds right, and I agree at some extent. This is anyway a bit trickier: if a device (not software) is removed, it cannot be restored. As the dbus path is the way we identify each device from the rollback, returning the *old* device path of a device which is no more present together with the failure bit seems better to me than skipping it from the returned list of restored devices at all: in this way we would not report an explicit error. It would be up to the caller to check that all the devices are there in order to see if the rollback was successful. Would this behavior be acceptable?
Side note: the *unrealized* case should be fixed to return error. This is not actually an issue for now as we don't accept unrealized devices in nm-checkpoint-mananger, but should be fixed (also if we keep the no more existent DBUS paths).

src/nm-checkpoint.c
fgiudici: But how do you report that a removed device has not been restored in the rollback?

thom311:
> As the dbus path is the way we identify each device from the rollback, returning the old device path of a device which is no more present together with the failure bit seems better to me than skipping it from the returned list of restored devices at all: in this way we would not report an explicit error. 

You are right, because
``` 
   <method name="CheckpointRollback">
      <arg name="checkpoint" type="o" direction="in"/>
      <arg name="result" type="a{su}" direction="out" />
    </method>
```
the return value is `s`, not `o`. Hence, we can (and should) return paths that for objects that no longer exist.

src/nm-checkpoint.c
fgiudici: I agree this could be a viable alternative.
The reasoning beyond using the dbus path as the key, was to keep a 1:1 mapping with what the API received as arguments. It doesn't look worse nor better to me than the approach proposed here: just different.

fgiudici:
Merged first part (fg/checkpoint-devices-rh1578335-pt1) on master.
Few changes on the remaining part:
1) Rebased on master
2) Made NM_DBUS_OBJECT_EXPORTED_CHANGED full managed (do as much as NM_MANAGER_DEVICE_REMOVED, as per Thomas advice)
3) Added minimal management of removed unrealized objects during rollback

Still to discuss if we want to keep proposed hash table indexing by dev_dbus_path or if we want to keep original obj Device indexing, keeping a separate Device list for removed devices.

src/nm-checkpoint.c
fgiudici: Updated branch as requested: keep NMDevice as the g_hash_table key and add a GPtrArray to track removed devices.

fgiudici:
Reworked the branch to keep NMDevice objects as keys in the g_hash_table tracking devices to rollback.

vbenes:
a crash here:
https://faf.lab.eng.brq.redhat.com/faf//reports/11121/

fgiudici:
> a crash here:
> https://faf.lab.eng.brq.redhat.com/faf//reports/11121/

Thanks, the assert on SW devices was clearly wrong. Fixed, branch updated.

src/nm-checkpoint.c
```diff
@@ -206,15 +206,93 @@ find_settings_connection (NMCheckpoint *self,
...
+		subject = nm_auth_subject_new_internal ();
+
+		/* Disconnect the device if needed. This necessary because now
+		 * the manager prevents the reactivation of the same connection by
+		 * an internal subject. */
+		if (   nm_device_get_state (dev_checkpoint->device) > NM_DEVICE_STATE_DISCONNECTED
+		    && nm_device_get_state (dev_checkpoint->device) < NM_DEVICE_STATE_DEACTIVATING) {
+			nm_device_state_changed (dev_checkpoint->device,
+						 NM_DEVICE_STATE_DEACTIVATING,
+						 NM_DEVICE_STATE_REASON_NEW_ACTIVATION);
```
thom311: Wrong indentation. Also, this is complained by checkpatch.pl. Are you running that as commit-hook? That would be a good idea.

src/nm-checkpoint.c
```diff
@@ -245,6 +248,15 @@ restore_and_activate_connection (NMCheckpoint *self,
...
 			return FALSE;
 		}
+
+		/* If the device is software, a brand new NMDevice may have been created */
+		if (   dev_checkpoint->is_software
+		    && !dev_checkpoint->device) {
+			dev_checkpoint->device = nm_manager_get_device (priv->manager,
+			                                                dev_checkpoint->original_dev_name,
+			                                                dev_checkpoint->dev_type);
+			g_object_ref (dev_checkpoint->device);
```
thom311: you need to handle the possibility that `nm_manager_get_device()` won't find the device.

that may just mean to use `nm_g_object_ref()` instead of `g_object_ref()`.

src/nm-checkpoint.c
```diff
@@ -586,6 +643,22 @@ nm_checkpoint_init (NMCheckpoint *self)
...
+	NMCheckpointPrivate *priv = NM_CHECKPOINT_GET_PRIVATE (self);
+	DeviceCheckpoint *dev_checkpoint;
+
+	dev_checkpoint = g_hash_table_lookup (priv->devices, device);
+	if (!dev_checkpoint)
+		return;
+
+	_move_dev_to_removed_devices (priv, dev_checkpoint);
+
+	_notify (self, PROP_DEVICES);
```
thom311: most of the code here is duplicated with `_dev_exported_changed()`. Add a function that does the common stuff.

src/nm-checkpoint.c
```diff
@@ -607,6 +680,7 @@ nm_checkpoint_new (NMManager *manager, GPtrArray *devices, guint32 rollback_time
 	priv->rollback_timeout_s = rollback_timeout_s;
 	priv->created_at_ms = nm_utils_get_monotonic_timestamp_ms ();
 	priv->flags = flags;
+	priv->removed_devices = NULL;
```
thom311: commonly we rely on the field to be pre-initialized to `%NULL`. I don't think this line is necessary.

src/nm-checkpoint.c
```diff
@@ -647,6 +725,10 @@ dispose (GObject *object)
 
 	g_clear_pointer (&priv->devices, g_hash_table_unref);
 	g_clear_pointer (&priv->connection_uuids, g_hash_table_unref);
+	if (priv->removed_devices)
+		g_ptr_array_unref (priv->removed_devices);
```
thom311: `dispose()` must be reentrant.

Use
```
    nm_clear_pointer (&priv->removed_devices, g_ptr_array_unref)
```

src/nm-checkpoint.c
```diff
@@ -647,6 +725,10 @@ dispose (GObject *object)
...
+
+	nm_clear_g_signal_handler (priv->manager, &priv->dev_removed_id);
```
thom311: `priv->manager` does not own a reference to NMManager.  Previously, that was less of a problem, because every use of `priv->manager` happened at points where the manager clearly was still alive (e.g. during `nm_checkpoint_rollback()`).

It's not clear that at the moment when the last reference to NMCheckpoint is dropped, that the manger is still alive. You need to add a reference, and `g_object_clear()` it in dispose.

Note that then NMManager and NMCheckpoint hold a reference to each other. But that is fine, because during shutdown, NMManager must first destroy all checkpoints (giving up it's reference and break the cycle).

Of course, currently we just utterly leak NMManager and it wouldn't matter. But if we ever hope to fix that, all places must get this right (first).

src/nm-checkpoint.c
fgiudici: thanks, fixed

src/nm-checkpoint.c
fgiudici: done

src/nm-checkpoint.c
fgiudici: _move_dev_to_removed_devices was already the function to do the common stuff.
The few lines similar to _dev_exported_changed() were there to show immediately the difference wrt here, i.e., asserting the presence of the device in the checkpoint devices hash table in the latter.
Anyway moving all in the _move_dev_to_removed_devices () would also have its points...
Done

src/nm-checkpoint.c
fgiudici: removed

src/nm-checkpoint.c
fgiudici: fixed, thanks

src/nm-checkpoint.c
fgiudici: But shouldn't we expect to be the manager the last one to be finalized (the scenario is NM shutting down, right)?
Anyway, agreed, linking an object without taking a reference is ugly.
Added a new commit on top of the branch for this.

src/nm-checkpoint.c
```diff
@@ -245,6 +248,15 @@ restore_and_activate_connection (NMCheckpoint *self,
...
 			return FALSE;
 		}
+
+		/* If the device is software, a brand new NMDevice may have been created */
+		if (   dev_checkpoint->is_software
+		    && !dev_checkpoint->device) {
+			dev_checkpoint->device = nm_manager_get_device (priv->manager,
+			                                                dev_checkpoint->original_dev_name,
+			                                                dev_checkpoint->dev_type);
+			nm_g_object_ref (dev_checkpoint->device);
```
thom311: Now if no device exits, you won't get an assertion failure during `g_object_ref()`.

But a few lines below, you'll pass `%NULL` to `nm_device_get_state (dev_checkpoint->device)`. I meant, %NULL must be handled, and that *may* be as simple as using `nm_g_object_ref()`. But looking at it, it requires a bit more...

fgiudici: Right. Pushed a fixup commit on top of the branch to fix this.

thom311: there are still several other code paths that lead to
```
»···»···if (   nm_device_get_state (dev_checkpoint->device) > NM_DEVICE_STATE_DISCONNECTED
»···»···    && nm_device_get_state (dev_checkpoint->device) < NM_DEVICE_STATE_DEACTIVATING) {
```

you need to handle the possibility of device being %NULL there...

fgiudici: Ok, the point is that we have to account a software device NOT created by NM that is removed after checkpoint creation. This is what is missing in the logic here.
Updated the fixup commit.

src/nm-checkpoint.c
```diff
@@ -260,6 +260,12 @@ restore_and_activate_connection (NMCheckpoint *self,
 		need_activation = TRUE;
 	}
 
+	if (!dev_checkpoint->device) {
+		nm_assert (dev_checkpoint->is_software);
```
thom311: where does this assert come from? I don't see why this is correct. Could you add a comment explaining why it's here?

fgiudici: OK, I can add a comment there. Does the rest looks good?

thom311:
lgtm

fgiudici:
Thanks, merged

PR #317, opened 2019-03-18
[th/nmcli-list-options]

thom311:
a large branch refactoring nmcli.

Still WIP, but I want to see how CI fares.

---
*update: Wed Mar 20 15:11:19 2019*

So what does it do?

- subtract operation like `nmcli connection modify "$PROFILE" -ipv4.addresses 192.168.7.24/24`:

  - no longer fail if the address is not yet added. The `'-'` only fails, if the value is not parsable/invalid.

  - support a list of values (not just one). As such, `'-'` and `'+'` and plain set work all the same

  - still support removing by index. That already worked for one index. But now a list of (zero based) indexes is supported: `nmcli connection modify "$PROFILE" -ipv4.addresses 1,2,3`. Again, nonexisting indexes are silently ignore.

- heavily refactor `clients/common/nm-meta-setting-desc.c`. In particular get rid of the DEFINE macros like `DEFINE_SETTER_*` and `DEFINE_REMOVER_*`. Instead, there are now all functions (property-type-handlers) and code is unified by using the same implementation (but differing `property_info->property_typ_data`). Note in particular the new property types `_pt_multilist`, `_pt_objlist`, and `_set_fcn_optionlist`. For the last type (optionlist), the implementations define different `get_fcn()`, so there is not *one* `_pt_optionlist`. But they optionlist types all share `_set_fcn_optionlist`.

- move logic of setting properties from higher layers down to `clients/common/nm-meta-setting-desc.c`. Previously, nmcli would try to parse set-values as integers and call `remove_fcn()` with the index to remove. Whether a string accepts integer indexes, must be decided by the property. For example, `sriov-vfs` don't support such numbers. Instead, nmcli now just calls `nmc_setting_set_property()` with modifiers `'+'`, `'-'` or `'\\0'`.

- merge `remove_fcn()` with `set_fcn()`. Previously, `set_fcn()` would only append values (in case of list values). And `remove_fcn()` would just remove one value. That meant, that remove and append was implemented separately (with separate parsing, which in the best case led to duplication, in the worst case to inconsistent behaviour).

- binary size is basically the same. Just to show it doesn't increase the binary. On F29, x86_64, `./contrib/fedora/rpm/build_clean.sh -g -W debug -W test` gives:
```
   - master: /usr/bin/nmcli   922536
   - new:    /usr/bin/nmcli   908224  (-1.55%)
```

---

*update Fri Mar 22 09:46:02 2019*

btw, what this also fixes. If you issue `nmcli connection modify "$PROFILE" setting.property value` (not the `+setting.property` form to append), then nmcli previously would always set the GObject property twice: first once to reset the default, and then during `set_fcn()`. Now of course that's no longer happening, because for plain properties `set_fcn()` knows that there is no need to reset the default first. Only for list properties it needs to first clear the list.

### DIFF
--- a/clients/cli/connections.c
+++ b/clients/cli/connections.c
@@ -7380,6 +7380,7 @@ editor_menu_main (NmCli *nmc, NMConnection *connection, const char *connection_t
 					                          prop_name);
 				}
 
+				/* setting a value in edit mode "appends". That seems unexpected behavior. */
 				if (!nmc_setting_set_property (nmc->client,
 				                               ss,
 				                               prop_name,
@@ -8101,6 +8102,9 @@ editor_init_existing_connection (NMConnection *connection)
 	NMSettingProxy *s_proxy;
 	NMSettingWireless *s_wireless;
 	NMSettingConnection *s_con;
+
+	/* FIXME: this approach of connecting handlers to do something is fundamentally
+	 * flawed. See the comment in nmc_setting_ip6_connect_handlers(). */
 
 	s_ip4 = nm_connection_get_setting_ip4_config (connection);
 	s_ip6 = nm_connection_get_setting_ip6_config (connection);
--- a/clients/cli/settings.c
+++ b/clients/cli/settings.c
@@ -142,6 +142,25 @@ ipv6_addresses_changed_cb (GObject *object, GParamSpec *pspec, gpointer user_dat
 		}
 	} else {
 		answered = FALSE;
+		/* FIXME: editor_init_existing_connection() and registering handlers is not the
+		 *  right approach.
+		 *
+		 * This only happens to work because in nmcli's edit mode
+		 * tends to append addresses -- instead of setting them.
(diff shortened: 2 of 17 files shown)


vbenes:
crashing here:
https://faf.lab.eng.brq.redhat.com/faf/reports/10968/

bengal:
LGTM

thom311:
merged to master.

Note that one CI test is failing: tests.team_abs_overwrite_watchers
this of course needs fixing.

I looked into it, trying to bisect the issue, but I wasn't unable to find a good version. I also didn't understand how this test could work. Probably I made a mistake in reproducing the issue.

The reason for merging it with one CI failure is that
- this is a large branch. I'd argue that it fixes more issues that it breaks. Also, I consider it the right direction anyway. By merging it early, we hopefully give it more testing. The time for ironing out the issues are limited, and it only takes longer by leaving the work in an pull-request
- other work already depends on this branch, and new work should be implemented in the way this branch suggests.

sorry about the CI failure.

PR #318, opened 2019-03-21
[th/libnm-team-fixes]

thom311:
CI tests for #317 fail now test `team_abs_set_runner_tx_hash`.

It reveals that `g_object_set (s_team, "runner-tx-hash", NULL, NULL)` does not actually clear the settings. That was hid by `set_fcn()` in nmcli wrongly calling `nm_setting_team_remove_runner_tx_hash()` [[1]](https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/clients/common/nm-meta-setting-desc.c?id=cfcd746260b0dc1af80dd769d67dfa2649569b36#n4082).

This branch attempts to address these issues for `NMSettingTeam`.

As of now,  `NMSettingTeamPort` is still not changed. Please first ACK this approach, before I am going to spend time doing the wrong thing. Thanks.

### DIFF
--- a/libnm-core/nm-utils-private.h
+++ b/libnm-core/nm-utils-private.h
@@ -179,7 +179,8 @@ _nm_utils_json_extract_strv (char *conf,
 	if (   !t_value
 	    || !G_TYPE_CHECK_VALUE_TYPE (t_value, G_TYPE_STRV))
 		return NULL;
-	return g_strdupv (g_value_get_boxed (t_value));
+	return    g_strdupv (g_value_get_boxed (t_value))
+	       ?: g_new0 (char *, 1);
 }
 
 static inline GPtrArray *
--- a/libnm-core/nm-utils-private.h
+++ b/libnm-core/nm-utils-private.h
@@ -134,7 +134,8 @@ _nm_utils_json_extract_int (char *conf,
 	nm_auto_unset_and_free_gvalue GValue *t_value = NULL;
 
 	t_value = _nm_utils_team_config_get (conf, key.key1, key.key2, key.key3, is_port);
-	if (!t_value)
+	if (   !t_value
+	    || !G_VALUE_HOLDS_INT (t_value))
 		return key.default_int;
 	return g_value_get_int (t_value);
 }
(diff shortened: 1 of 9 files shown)


libnm-core/nm-utils-private.h
```diff
@@ -105,6 +105,16 @@ gboolean    _nm_sriov_vf_parse_vlans (NMSriovVF *vf, const char *str, GError **e
 
 /* JSON to GValue conversion macros */
 
+static inline void
+_nm_auto_gvalue_unset_and_free (GValue **ptr)
+{
```
bengal: other similar macros have the `nm_auto_<verb>_<object>` form, i.e. `_nm_auto_unset_and_free_gvalue`

thom311: fixed.

thom311:
ok, this breaks some tests and needs more thought. I think we need here a larger effort ([[rhbz#1691619]](https://bugzilla.redhat.com/show_bug.cgi?id=1691619)). That is for another day.

I dropped the patches that cause (bad) change in behaviour and intend to merge this early.

I'll add a workaround to #317 for the moment.


The full branch is at branch [th/libnm-team-fixes-pt2](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commits/th/libnm-team-fixes-pt2)

fgiudici:
> ok, this breaks some tests and needs more thought. I think we need here a larger effort ([[rhbz#1691619]](https://bugzilla.redhat.com/show_bug.cgi?id=1691619)). That is for another day.
> 
> I dropped the patches that cause (bad) change in behaviour and intend to merge this early.
> 
I was just looking into the failures and what broke the sync from NMSettingTeam properties to the JSON config... no more need now.
Agreed the effort is not trivial: keeping the sync may be tricky.

> I'll add a workaround to #317 for the moment.
> 
> The full branch is at branch [th/libnm-team-fixes-pt2](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commits/th/libnm-team-fixes-pt2)

Ok for the approach in this PR. If tests are ok I would merge this branch.

thom311:
merged to master.

PR #320, opened 2019-03-22
ovs: don't traverse interface through disconnected when the ovsdb entry is removed

lkundrak:
Go straight to unmanaged. That's what all the other devices do when
their backing resources vanish. If the device reached disconnected
state, an autoconnect check would try to connect it back, in vain.

### DIFF
--- a/src/devices/ovs/nm-ovs-factory.c
+++ b/src/devices/ovs/nm-ovs-factory.c
@@ -120,22 +120,13 @@ ovsdb_device_removed (NMOvsdb *ovsdb, const char *name, NMDeviceType device_type
                       NMDeviceFactory *self)
 {
 	NMDevice *device;
-	NMDeviceState device_state;
 
 	device = nm_manager_get_device (nm_manager_get (), name, device_type);
 	if (!device)
 		return;
 
-	device_state = nm_device_get_state (device);
-	if (   device_type == NM_DEVICE_TYPE_OVS_INTERFACE
-	    && device_state > NM_DEVICE_STATE_DISCONNECTED
-	    && device_state < NM_DEVICE_STATE_DEACTIVATING) {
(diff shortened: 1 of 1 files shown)


src/devices/ovs/nm-ovs-factory.c
```diff
@@ -120,22 +120,13 @@ ovsdb_device_removed (NMOvsdb *ovsdb, const char *name, NMDeviceType device_type
...
-	device_state = nm_device_get_state (device);
-	if (   device_type == NM_DEVICE_TYPE_OVS_INTERFACE
-	    && device_state > NM_DEVICE_STATE_DISCONNECTED
-	    && device_state < NM_DEVICE_STATE_DEACTIVATING) {
-		nm_device_state_changed (device,
-		                         NM_DEVICE_STATE_DEACTIVATING,
-		                         NM_DEVICE_STATE_REASON_REMOVED);
-	} else if (device_state == NM_DEVICE_STATE_UNMANAGED) {
+	if (device_type == NM_DEVICE_TYPE_OVS_INTERFACE)
 		nm_device_unrealize (device, TRUE, NULL);
```
thom311: OK, other devices do that? Can't argue with that, and it's fine with me for the moment.

But generally, it seems to be we should be very strict about from which states we switch to other states. e.g. when we are activated, we should always deactivate by hitting the states DISCONNECTING, and possibly FAILED/DISCONNECTED, then UNAVAILABLE, then UNMANAGED. If that causes issues (autoconnect during disconnected), then that should be solved differently, not by skipping states. Of course, that's for another day.

fgiudici:
Change looks good to me, but this seems to hit CI when ovs devices are removed.

lkundrak:
Folded into #324 so that it's tested together.

PR #321, opened 2019-03-23
[th/routing-rule-pt2] add libnm API for Routing Rules

thom311:
adds
 - libnm API for routing rules (`NMIPRoutingRule`)
 - keyfile support
 - ifcfg-rh support
 - nmcli support

Such a setting is still completely ignored by NetworkManager. That will follow later.

Also missing a more attributes for routing rules. Those too will be added later.


shared/nm-utils/nm-shared-utils.c
```diff
@@ -2505,6 +2505,59 @@ _nm_utils_unescape_plain (char *str, const char *candidates, gboolean do_strip)
 	return str;
 }
 
+char *
+nm_utils_str_simpletokens_extract_next (char **p_line_start)
+{
```
bengal: What the functions does is not clear from the name, can you add a comment?

libnm-core/nm-setting-ip-config.c
```diff
@@ -1384,6 +1385,2125 @@ _nm_ip_route_attribute_validate_all (const NMIPRoute *route)
...
+	if (!self->from_str) {
+		nm_assert (self->from_valid);
+		((NMIPRoutingRule *) self)->from_str = nm_utils_inet_ntop_dup (_ip_routing_rule_get_addr_family (self),
+		                                                               &self->from_bin);
+	}
+	return self->from_str;
+}
+
+const NMIPAddr *
+nm_ip_routing_rule_get_from_bin (const NMIPRoutingRule *self)
```
bengal: Maybe, add an underscore prefix to internal-only functions?

bengal:
The rest LGTM.

shared/nm-utils/nm-shared-utils.c
thom311: I did not, but the parsing/escaping here needs some further improvements. I will do that in a follow-up.

libnm-core/nm-setting-ip-config.c
thom311: Maybe. We are not consistent about this (and I don't even think we have a preferred/recommended style). I didn't change that now.

thom311:
merged to master

PR #323, opened 2019-03-26
logging: don't leave protected log domain unset

lkundrak:
Otherwise the following in the config file results in VPN_PLUGIN (and
others) not logging at all.

  [logging]
  level=trace

It should go to the default level (INFO) instead.

### DIFF
--- a/src/nm-logging.c
+++ b/src/nm-logging.c
@@ -157,9 +157,9 @@ NMLogDomain _nm_logging_enabled_state[_LOGL_N_REAL] = {
 	 *
 	 * Note: LOGD_VPN_PLUGIN is special and must be disabled for
 	 * DEBUG and TRACE levels. */
-	[LOGL_INFO] = LOGD_DEFAULT,
-	[LOGL_WARN] = LOGD_DEFAULT,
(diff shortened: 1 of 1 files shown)


thom311:
```
»···LOGD_ALL       = (((__LOGD_MAX - 1LL) << 1) - 1LL),
»···LOGD_DEFAULT   = LOGD_ALL & ~(
»···                              LOGD_DBUS_PROPS |
»···                              LOGD_WIFI_SCAN |
»···                              LOGD_VPN_PLUGIN |
»···                              0),
```

This patch does not only change VPN_PLUGIN domain, but others too. The commit message makes no mention about what happens to them, and why the patch is correct for them.

The commit message is anyway unclear what it tries to fix. If you set

```
[logging]
level=TRACE
```
in config, then that should have the same effect as
```
[logging]
level=DEFAULT
```
Does it? Does it not? What does not work?

thom311:
Another way to see it: this changes the initialization of the static variable from
```
-	[LOGL_*]  = LOGD_DEFAULT,
+	[LOGL_* = LOGD_ALL,
```
This looks wrong. The static variable should be initialized with the default, so this is a red flag.

PR #324, opened 2019-03-27
ovs-interface: dissociate the link on disconnection

lkundrak:
Open vSwitch is the special kid on the block -- it likes to be in charge of
the link lifetime and so we shouldn't be. This means that we shouldn't be
attempting to remove the link: we'd just (gracefully) fail anyways.

More importantly, this also means that we shouldn't care if we see the link
go away. Once the device reaches DISCONNECTED state, its configuration is
cleaned up and we may already be activating another connection. We shouldn't
alter the device state when OpenVSwitch decides to drop the old link.

https://bugzilla.redhat.com/show_bug.cgi?id=1543557

### DIFF
--- a/src/devices/ovs/nm-device-ovs-interface.c
+++ b/src/devices/ovs/nm-device-ovs-interface.c
@@ -118,6 +118,16 @@ link_changed (NMDevice *device,
 	}
 }
 
+static void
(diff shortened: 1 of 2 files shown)


fgiudici:
LGTM

PR #326, opened 2019-03-31
cli: fix a crash on "nmcli d wifi hotspot"

lkundrak:
Call the correct _finish() function for
nm_client_add_and_activate_connection_async(). add_and_activate_cb()
somewhat confusingly alternates between two different ones depending on
whether info->create is set.

Fixes: 35932375272a ('cli: reuse connections in nmcli dev wifi con')


bengal:
LGTM

vbenes:
still see a crash http://faf.lab.eng.brq.redhat.com/faf/reports/11132/

PR #327, opened 2019-04-01
clients: only ask secrets for settings that require them

bengal:
When nmcli needs secrets for a connection it asks them for every known setting. nmtui is a bit smarter and asks them only for settings that actually exist in the connection. Make a step further and let clients ask secrets only for setting that exist *and* have any secret property. This decreases the number of D-Bus calls when editing or showing a connection with secrets.

https://bugzilla.redhat.com/show_bug.cgi?id=1506536


lkundrak:
Looks good, thanks.

thom311:
lgtm

bengal:
Merged as 5b5a768b693a3685ee0acd3c9ea31d1f7ce70d19.

PR #328, opened 2019-04-02
secret-agent-simple: pass hints to the auth dialog

lkundrak:


thom311:
```
+    gs_unref_ptrarray GPtrArray *auth_dialog_argv = g_ptr_array_new ();
```
There are code paths that return early without requiring the `auth_dialog_argv` array. Maybe create the pointer array later, before it's actually needed.

Rest lgtm

PR #329, opened 2019-04-02
[th/strstrip]

thom311:
few patches to add `nm_strstrip_avoid_copy_a()`.

In most cases, we don't need to clone the string at all, because trailing whitespace not the usual case. Avoid that. Even if we need to copy the string, usually the string is short and we can just stack allocate it. Add  and use `nm_strstrip_avoid_copy_a()` for that.


bengal:
LGTM

thom311:
merged to master

PR #331, opened 2019-04-04
platform: assume 'sriov_drivers_autoprobe' is 1 when sysfs file is missing

bengal:
'sriov_drivers_autoprobe' was added in kernel 4.12. With previous
kernel versions NM is currently unable to set any SR-IOV parameter
because it tries to read 'sriov_drivers_autoprobe' which doesn't
exist, assumes that current value is -1 and tries to change it,
failing.

When the file doesn't exist, drivers are automatically probed so we
can assume the value is 1. In this way NM is able to activate a
connection with sriov.autoprobe-drivers=1 (the default) even on older
kernel versions.

Fixes: 1e41495d9a1b ('platform: sriov: write new values when we can't read old ones')

https://bugzilla.redhat.com/show_bug.cgi?id=1695093

### DIFF
--- a/libnm-core/nm-keyfile.c
+++ b/libnm-core/nm-keyfile.c
@@ -643,7 +643,7 @@ _build_list_create (GKeyFile *keyfile,
 	for (i_keys = 0; i_keys < n_keys; i_keys++) {
 		const char *s_key = keys[i_keys];
 		gint32 key_idx;
-		gint8 key_type;
+		gint8 key_type = 0;
 
 		switch (build_list_type) {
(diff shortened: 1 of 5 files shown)


thom311:
The analysis sounds good, but I'd be more specific about the cases when to employ the workaround (and not treat all failures to read the sysctl the same).

I'd propose a bit different solution, pushed two patches on top of yours.

bengal:
> I'd propose a bit different solution, pushed two patches on top of yours.

LGTM, merged.

PR #332, opened 2019-04-04
[th/strsplit]

thom311:
Recently, when refactoring nmcli (f224d6a8537dd58fafc8b7ba977ec7fffe8c31e9), handling of list-typed properties improved.

Still, the splitting such values should be improved. For example, currently there is no way to escape the ',' separator. The code (and behavior) for this should be unified and powerful to support arbitrary strings (that contain the delimiter characters).

In a first step, improve our `nm_utils_strsplit_set*()` implementation which will be used for that.

`nm_utils_strsplit_set*()` is a replacement for `g_strsplit_set()`, but differs in some aspects:
- it may (configurably) skip empty tokens right away (repeated delimiters)
- it only clones the string once, instead of returning a full strv array.
- it may (optionally) honor backslash as escape characters. We will need this to make splitting options in nmcli escapeable.

Add `NM_UTILS_STRSPLIT_SET_FLAGS_PRESERVE_EMPTY` flag, so that empty tokens are preserved. As such, `nm_utils_strsplit_set_with_empty()` is now very similar to `g_strsplit_set()`.

Also, replace uses of `g_strsplit_set()` with our variant. That avoids the overhead of cloning all items of the strv array and we can leverage our variant -- that we already have and need for `NM_UTILS_STRSPLIT_SET_FLAGS_PRESERVE_EMPTY`.

### DIFF
--- a/clients/cli/connections.c
+++ b/clients/cli/connections.c
@@ -1528,15 +1528,12 @@ split_required_fields_for_con_show (const char *input,
                                     char **active_flds,
                                     GError **error)
 {
-	char **fields, **iter;
-	char *dot;
-	GString *str1, *str2;
-	gboolean found;
+	gs_free const char **fields = NULL;
+	const char *const*iter;
+	nm_auto_free_gstring GString *str1 = NULL;
+	nm_auto_free_gstring GString *str2 = NULL;
 	gboolean group_profile = FALSE;
 	gboolean group_active = FALSE;
-	gboolean success = TRUE;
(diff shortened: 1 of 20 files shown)


thom311:
TODO:

> ./tools/run-nm-test.sh -v -m libnm-core/tests/test-general

bengal:
LGTM

bengal:
LGTM

thom311:
thanks. Merged to master.

PR #333, opened 2019-04-04
[lr/multi-level-master-slave] fixes for multi-level maser-slave relationship handling

lkundrak:
$SUBJ, with extra fixes for Open VSwitch (that always uses two-level master-slave realtionships).

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -5075,7 +5075,6 @@ nm_device_slave_notify_release (NMDevice *self, NMDeviceStateReason reason)
 {
 	NMDevicePrivate *priv = NM_DEVICE_GET_PRIVATE (self);
 	NMConnection *connection = nm_device_get_applied_connection (self);
-	NMDeviceState new_state;
 	const char *master_status;
 
 	g_return_if_fail (priv->master);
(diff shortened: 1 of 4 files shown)


thom311:
This topic is complicated, so I cannot tell whether this is all correct. It does look reasonable to me, so +1. Let's wait how CI fares.

---

Let me add something about state-reasons. I always found it ugly to make decisions based on the device-state-reason, in particular, because when we change the device's state (e.g. from DISCONNECTED to PREPARE), then usually the device's state progresses further. For these future state changes it's often not clear which state-reason to pass on (likely, the state-reason should be preserved for a while, but sometimes something else might happen, and we'd like to use a different state-reason for the following state changes). There is also [another reason](https://github.com/NetworkManager/NetworkManager/blob/ab65c085af25876aeefa6a56e8d2f761759b0f35/src/devices/nm-device.h#L57).

Hence, I'd prefer if state reasons are *only* exported on D-Bus (in the `StateChanged` signal of the device). That's not happening today yet, so, I marked all places where we make decisions based on the state-change-reason with `nm_device_state_reason_check()`. Grep for that. That's the reason why `nm_device_state_reason_check()` exists.
It also means: avoid making decisions based on the state-change-reason, and if you really have to, use `nm_device_state_reason_check()`.

vbenes:
rebuild

PR #334, opened 2019-04-05
libnm/infiniband: lift restriction of MTU to 2044 for IPoIB in "datag…

thom311:
…ram" mode

Traditionally, the MTU in "datagram" transport mode was restricted to
2044. That is no longer the case, relax that.

In fact, choose a very large maximum and don't differenciate between
"connected" mode (they now both use now 65520). This is only the
limitation of the connection profile. Whether setting such large MTUs
actually works must be determined when activating the profile.

Initscripts "ifup-ib" from rdma-core package originally had a limit of 2044.
This was raised to 4092 in rh#1186498. It is suggested to raise it further
in bug rh#1647541.

In general, kernel often does not allow setting large MTUs. And even if it
allows it, it may not work because it also requires the entire network to
be configured accordingly. But that means, it is generally not helpful to
limit the MTU in the connection profile too strictly. Just allow large
MTUs, we need to see at activation time whether the configuration works.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1186498
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1593334
         (rdma-core: raise limit from 2044 to 4092 in ifup-ib)

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1647541
         (rdma-core: raise limit beyond 4092 in ifup-ib)

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1532638#c4
         (rdma-core: MTU related discussion)

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1534869
       (NetworkManager bug about this topic, but with lots of unrelated
        discussion. See in particular #c16)

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1653494

### DIFF
--- a/libnm-core/nm-setting-infiniband.c
+++ b/libnm-core/nm-setting-infiniband.c
@@ -170,7 +170,6 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
 {
 	NMSettingConnection *s_con = NULL;
 	NMSettingInfinibandPrivate *priv = NM_SETTING_INFINIBAND_GET_PRIVATE (setting);
-	guint32 normerr_max_mtu = 0;
 
 	if (priv->mac_address && !nm_utils_hwaddr_valid (priv->mac_address, INFINIBAND_ALEN)) {
 		g_set_error_literal (error,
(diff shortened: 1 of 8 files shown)


libnm-core/nm-setting-infiniband.c
```diff
@@ -271,12 +265,21 @@ verify (NMSetting *setting, NMConnection *connection, GError **error)
...
+		 *
+		 * Note that this is the MTU in the connection profile. Whether
+		 * we will be able to configure large MTUs later (during activation)
+		 * is unknown at this point. */
 		g_set_error (error,
 		             NM_CONNECTION_ERROR,
 		             NM_CONNECTION_ERROR_INVALID_PROPERTY,
-		             _("mtu for transport mode '%s' can be at most %d but it is %d"),
-		             priv->transport_mode, normerr_max_mtu, priv->mtu);
+		             _("mtu for transport mode '%s' can be at most %u but it is %u"),
```
bengal: Since the max mtu now doesn't depend on the transport mode, perhaps the "for transport mode %s" part can be dropped.

thom311: fixed.

thom311:
Repushed. Also fixed unit-tests.

lkundrak:
LGTM

fgiudici:
LGTM

thom311:
merged to master as 69765c11fd8e175e203a2a13265f7984f943c493

PR #335, opened 2019-04-05
[th/no-wifi-scan-rand-mac-address]

thom311:
https://bugzilla.redhat.com/show_bug.cgi?id=1695696

---

Quote:

    contrib/rpm: install "/usr/lib/NetworkManager/conf.d/10-no-wifi-scan-rand-mac-address.conf"
    
    On Ubuntu 18.04 (bionic), package "wpasupplicant" installs a file
    "/usr/lib/NetworkManager/conf.d/no-mac-addr-change.conf" with similar
    content and purpose.
    
    These drivers are known to work badly with NetworkManager changing the
    MAC address. There is no point in causing pain for users for something
    that they cannot fix (short of buying different hardware).
    
    Install the snippet with "NetworkManager-wifi" package.
    
    [1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=833507
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1695696

### DIFF
--- a/Makefile.examples
+++ b/Makefile.examples
@@ -147,8 +147,8 @@ EXTRA_DIST += \
 	examples/lua/lgi/change-vpn-username.lua \
 	examples/lua/lgi/deactivate-all.lua \
 	\
+	examples/nm-conf.d/10-no-wifi-scan-rand-mac-address.conf \
 	examples/nm-conf.d/30-anon.conf \
-	examples/nm-conf.d/31-mac-addr-change.conf \
 	\
 	examples/python/dbus/nm-state.py \
 	examples/python/dbus/add-connection.py \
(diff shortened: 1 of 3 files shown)


vbenes:
rebuild

lkundrak:
No no no. No. Why is this even needed? It should be obvious why this is a horrible thing to do. If the drivers are broken then the drivers should be fixed. If we slap a band-aid like this, then we don't even give them a chance to ever work correctly.

For quite some time we detect when we fail to change a MAC address, log a warning and back off. We've certainly tested this with some of the drivers in the blacklist.

If we ever conclude that a driver is unfixable and abandoned to the point it would make sense to impose such a blacklist, we should include a reference to the attempts to contact the driver maintainer and linux.git/MAINTAINERS patch that marks the driver unmaintained.

thom311:
> For quite some time we detect when we fail to change a MAC address

Note that at [rhbz#1695696](https://bugzilla.redhat.com/show_bug.cgi?id=1695696), the "wl" driver appears to set the MAC address, but later something seems broken. So, it's not merely the non-ability to change the MAC (which we handle fine, I guess), but that something doesn't work afterwards.

Anyway, ok, I don't really mind either way.

PR #336, opened 2019-04-07
[th/systemd-resolved-dbus-activated]

thom311:
See

- https://wiki.archlinux.org/index.php/NetworkManager#Unit_dbus-org.freedesktop.resolve1.service_not_found
- https://bugs.archlinux.org/task/62138

### DIFF
--- a/src/dns/nm-dns-systemd-resolved.c
+++ b/src/dns/nm-dns-systemd-resolved.c
@@ -44,8 +44,9 @@
 #include "NetworkManagerUtils.h"
 #include "nm-dbus-compat.h"
 
-#define SYSTEMD_RESOLVED_DBUS_SERVICE "org.freedesktop.resolve1"
-#define SYSTEMD_RESOLVED_DBUS_PATH "/org/freedesktop/resolve1"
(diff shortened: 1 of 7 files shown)


src/dns/nm-dns-systemd-resolved.c
```diff
@@ -124,13 +125,23 @@ call_done (GObject *source, GAsyncResult *r, gpointer user_data)
...
 	if (!v) {
-		if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
-			return;
-		_LOGW ("send-updates failed: %s", error->message);
-	}
+		if (!priv->send_updates_warn_ratelimited) {
+			_LOGW ("send-updates failed to update systemd-resolved: %s", error->message);
+			priv->send_updates_warn_ratelimited = TRUE;
+		}
+		_LOGD ("send-updates failed: %s", error->message);
```
bengal: The error message is printed 2 times.

thom311: that was intentional, but I agree, it's better not to do this. Changed.

thom311:
merged to master

PR #338, opened 2019-04-09
Add some more TC attributes

lkundrak:


libnm-core/nm-setting-tc-config.c
```diff
@@ -1246,7 +1366,9 @@ _qdiscs_to_variant (GPtrArray *qdiscs)
 	if (qdiscs) {
 		for (i = 0; i < qdiscs->len; i++) {
 			NMTCQdisc *qdisc = qdiscs->pdata[i];
+			gs_strfreev char **attrs = nm_tc_qdisc_get_attribute_names (qdisc);
 			GVariantBuilder qdisc_builder;
+			int y;
```
thom311: the names are inside a `GHashTable`, which has `guint` type for the number of elements. We should not use a integer type that is smaller than the number of elements that we theoretically could have.

libnm-core/nm-setting-tc-config.c
```diff
@@ -1259,6 +1381,11 @@ _qdiscs_to_variant (GPtrArray *qdiscs)
 			g_variant_builder_add (&qdisc_builder, "{sv}", "parent",
 			                       g_variant_new_uint32 (nm_tc_qdisc_get_parent (qdisc)));
 
+			for (y = 0; attrs[y]; y++) {
+				g_variant_builder_add (&qdisc_builder, "{sv}", attrs[y],
+				                       nm_tc_qdisc_get_attribute (qdisc, attrs[y]));
```
thom311: do we really need to call `nm_tc_qdisc_get_attribute()` here which asserts that the name is non-empty?

libnm-core/nm-setting-tc-config.c
```diff
@@ -1308,8 +1437,17 @@ _qdiscs_from_variant (GVariant *value)
 			goto next;
 		}
 
-		if (g_variant_lookup (qdisc_var, "handle", "u", &handle))
-			nm_tc_qdisc_set_handle (qdisc, handle);
+		g_variant_iter_init (&qdisc_iter, qdisc_var);
+		while (g_variant_iter_next (&qdisc_iter, "{&sv}", &key, &attr_value)) {
+			if (   strcmp (key, "kind") == 0
```
thom311: leaks `attr_value`

```
    _nm_unused gs_unref_variant GVariant *attr_value_free = attr_value;
```

libnm-core/nm-utils.c
```diff
@@ -2309,6 +2309,15 @@ static const NMVariantAttributeSpec * const tc_object_attribute_spec[] = {
 	NULL,
 };
 
+typedef struct {
+	const char *kind;
+	const NMVariantAttributeSpec * const *attrs;
+} NMQdiscAttributeSpec;
+
+NMQdiscAttributeSpec * const tc_qdisc_attribute_spec[] = {
```
thom311: `static const NMQdiscAttributeSpec *const tc_qdisc_attribute_spec[]`

libnm-core/nm-utils.c
```diff
@@ -2323,8 +2332,12 @@ static const NMVariantAttributeSpec * const tc_object_attribute_spec[] = {
 void
 _nm_utils_string_append_tc_qdisc_rest (GString *string, NMTCQdisc *qdisc)
 {
+	gs_unref_hashtable GHashTable *ht = NULL;
 	guint32 handle = nm_tc_qdisc_get_handle (qdisc);
 	const char *kind = nm_tc_qdisc_get_kind (qdisc);
+	gs_strfreev char **attr_names = NULL;
+	gs_free char *str = NULL;
+	int i;
```
thom311: guint

libnm-core/nm-utils.c
```diff
@@ -2333,6 +2346,20 @@ _nm_utils_string_append_tc_qdisc_rest (GString *string, NMTCQdisc *qdisc)
 	}
 
 	g_string_append (string, kind);
+
+	ht = g_hash_table_new_full (nm_str_hash, g_str_equal, NULL, NULL);
```
thom311: don't allocate `@ht` if there are no attribute names.

libnm-core/nm-utils.c
```diff
@@ -2455,21 +2482,45 @@ nm_utils_tc_qdisc_from_str (const char *str, GError **error)
...
+	GHashTableIter iter;
+	gpointer key, value;
+	int i;
 
 	nm_assert (str);
 	nm_assert (!error || !*error);
 
-	ht = nm_utils_parse_variant_attributes (str,
+	options = nm_utils_parse_variant_attributes (str,
 	                                        ' ', ' ', FALSE,
```
thom311: intentation

libnm-core/nm-utils.c
```diff
@@ -2455,21 +2482,45 @@ nm_utils_tc_qdisc_from_str (const char *str, GError **error)
...
 	                                        tc_object_attribute_spec,
 	                                        error);
-	if (!ht)
+	if (!options)
 		return NULL;
+	g_clear_pointer (&options, g_hash_table_unref);
```
thom311: `nm_clear_pointer()` instead of `g_clear_pointer()`. It does not cast the pointer to void, and you'd get a compiler error when the function's signature does not match.

libnm-core/nm-utils.c
```diff
@@ -2455,21 +2482,45 @@ nm_utils_tc_qdisc_from_str (const char *str, GError **error)
 	gs_free char *kind = NULL;
 	gs_free char *rest = NULL;
 	NMTCQdisc *qdisc = NULL;
-	gs_unref_hashtable GHashTable *ht = NULL;
+	gs_unref_hashtable GHashTable *options = NULL;
+	GHashTableIter iter;
+	gpointer key, value;
+	int i;
```
thom311: guint

libnm-core/nm-utils.c
```diff
@@ -2455,21 +2482,45 @@ nm_utils_tc_qdisc_from_str (const char *str, GError **error)
...
+			break;
+		}
+	}
+
+	if (options) {
+		value = g_hash_table_lookup (options, "");
+		if (value)
+			rest = g_variant_dup_string (value, NULL);
+		else
+			rest = NULL;
```
thom311: this looks not right. What does it do? What's special about `""`?

Also, the ownership handling of `@rest` seems not right. Above, I'd expect `rest` to own some string. Here, you just reset the value.

libnm-core/nm-utils.c
thom311: This just throws away `@options`. That doesn't look right, shouldn't it consider what was parsed there?

libnm-core/nm-utils.c
```diff
@@ -2309,12 +2309,25 @@ static const NMVariantAttributeSpec * const tc_object_attribute_spec[] = {
...
+	NULL,
+};
+
 typedef struct {
 	const char *kind;
 	const NMVariantAttributeSpec * const *attrs;
 } NMQdiscAttributeSpec;
 
 NMQdiscAttributeSpec * const tc_qdisc_attribute_spec[] = {
+	&(NMQdiscAttributeSpec) { "fq_codel", tc_qdisc_fq_codel_spec },
```
thom311: the value should be const. In this context it allows the linker to set the memory read-only.

src/devices/nm-device.c
```diff
@@ -6527,6 +6527,46 @@ tc_commit (NMDevice *self)
 			qdisc->parent = nm_tc_qdisc_get_parent (s_qdisc);
 			qdisc->info = 0;
 
+			if (strcmp (qdisc->kind, "fq_codel") == 0) {
+				GVariant *var;
+
+				var = nm_tc_qdisc_get_attribute (s_qdisc, "limit");
+				if (var && g_variant_is_of_type (var, G_VARIANT_TYPE_UINT32))
+					qdisc->fq_codel.limit = g_variant_get_uint32 (var);
```
thom311: see https://github.com/NetworkManager/NetworkManager/blob/f5e8bbc8e082520af714df290da262ad731ccc4d/src/nm-ip4-config.c#L848 . I like that better.

src/platform/nm-platform.h
```diff
@@ -596,13 +596,27 @@ typedef struct {
...
 
+typedef struct {
+	guint32 limit;
+	guint32 flows;
+	guint32 target;
+	guint32 interval;
+	guint32 quantum;
+	guint32 ce_threshold;
+	guint32 memory;
+	gboolean ecn;
```
thom311: `bool ecn:1;`

src/platform/nm-platform.c
```diff
@@ -6407,6 +6441,17 @@ nm_platform_qdisc_cmp (const NMPlatformQdisc *a, const NMPlatformQdisc *b)
 	NM_CMP_FIELD (a, b, handle);
 	NM_CMP_FIELD (a, b, info);
 
+	if (strcmp (a->kind, "fq_codel") == 0) {
```
thom311: the function must not crash if certain fields are unset. `nm_streq0()`.

src/platform/nm-platform.c
```diff
@@ -6369,17 +6369,40 @@ const char *
...
+	char *p = opt_buf;
+	gsize l = sizeof (opt_buf);
 
 	if (!nm_utils_to_string_buffer_init_null (qdisc, &buf, &len))
 		return buf;
 
-	g_snprintf (buf, len, "%s%s family %d handle %x parent %x info %x",
+	if (nm_streq (qdisc->kind, "fq_codel")) {
+		if (qdisc->fq_codel.limit)
+			nm_utils_strbuf_append (&p, &l, " limit %d", qdisc->fq_codel.limit);
```
thom311: these are all `guint32`. Use `%u`.

src/platform/nm-platform.c
```diff
@@ -6369,17 +6369,40 @@ const char *
...
 	char str_dev[TO_STRING_DEV_BUF_SIZE];
+	char opt_buf[300] = { 0, };
+	char *p = opt_buf;
+	gsize l = sizeof (opt_buf);
 
 	if (!nm_utils_to_string_buffer_init_null (qdisc, &buf, &len))
 		return buf;
 
-	g_snprintf (buf, len, "%s%s family %d handle %x parent %x info %x",
+	if (nm_streq (qdisc->kind, "fq_codel")) {
```
thom311: we must also to-string ~invalid~ (or at least ~incomplete~) instances. `nm_streq0()`.

src/platform/nm-platform.c
```diff
@@ -6369,17 +6369,40 @@ const char *
...
+
+	g_snprintf (buf, len, "%s%s family %d handle %x parent %x info %x%s",
 	            qdisc->kind,
 	            _to_string_dev (NULL, qdisc->ifindex, str_dev, sizeof (str_dev)),
 	            qdisc->addr_family,
 	            qdisc->handle,
 	            qdisc->parent,
-	            qdisc->info);
+	            qdisc->info,
+	            opt_buf);
```
thom311: `opt_buf` is a buffer that's appended at the end. So, replace `g_snprintf()` by `nm_utils_strbuf_append()`, and then don't use `opt_buf`, but let it append to the correct buffer right away.

src/platform/nm-linux-platform.c
```diff
@@ -4246,6 +4289,31 @@ _nl_msg_new_qdisc (int nlmsg_type,
...
 	NLA_PUT_STRING (msg, TCA_KIND, qdisc->kind);
 
+	if (!(tc_options = nla_nest_start (msg, TCA_OPTIONS)))
+		goto nla_put_failure;
+
+	if (strcmp (qdisc->kind, "fq_codel") == 0) {
+		if (qdisc->fq_codel.limit)
+			NLA_PUT_U32 (msg, TCA_FQ_CODEL_LIMIT, qdisc->fq_codel.limit);
+
+			if (qdisc->fq_codel.flows)
```
thom311: indentation

src/platform/nm-linux-platform.c
```diff
@@ -3574,6 +3582,40 @@ _new_from_nl_qdisc (struct nlmsghdr *nlh, gboolean id_only)
 	obj->qdisc.parent = tcm->tcm_parent;
 	obj->qdisc.info = tcm->tcm_info;
 
+	if (tb[TCA_OPTIONS]) {
+		struct nlattr *options_attr;
+		int remaining;
+
+		nla_for_each_nested (options_attr, tb[TCA_OPTIONS], remaining) {
+			switch (nla_type (options_attr)) {
```
thom311: I think this lacks validation, doesn't it? You just assume that `options_attr` contains a u32 type.

libnm-core/nm-setting-tc-config.c
```diff
@@ -62,11 +62,19 @@ nm_tc_qdisc_new (const char *kind,
...
 
+	if (!kind || !*kind) {
+		g_set_error (error,
+		             NM_CONNECTION_ERROR,
+		             NM_CONNECTION_ERROR_INVALID_PROPERTY,
+		             _("kind is missing."));
+		return NULL;
+	}
+
 	if (!kind || !*kind || strchr (kind, ' ') || strchr (kind, '\t')) {
```
fgiudici: Minor note: here we can now remove the check on "!kind" and "!*kind" as already checked above:
`if (strchr (kind, ' ') || strchr (kind, '\t'))`

lkundrak:
First half split off here: https://github.com/NetworkManager/NetworkManager/pull/343

src/platform/nm-linux-platform.c
```diff
@@ -3506,6 +3514,43 @@ _new_from_nl_qdisc (struct nlmsghdr *nlh, gboolean id_only)
...
+	if (tb[TCA_OPTIONS]) {
+		struct nlattr *options_attr;
+		int remaining;
+
+		nla_for_each_nested (options_attr, tb[TCA_OPTIONS], remaining) {
+			if (nla_len (options_attr) < sizeof (uint32_t))
+				continue;
+
+			switch (nla_type (options_attr)) {
+			case TCA_FQ_CODEL_LIMIT:
```
bengal: I think we should check that the kind is "fq_codel" before parsing options as TCA_FQ_CODEL_* because other qdiscs use the same numerical attribute values.

src/platform/nm-platform.c
```diff
@@ -6430,13 +6430,32 @@ nm_platform_qdisc_to_string (const NMPlatformQdisc *qdisc, char *buf, gsize len)
...
+	                        qdisc->kind,
+	                        _to_string_dev (NULL, qdisc->ifindex, str_dev, sizeof (str_dev)),
+	                        qdisc->addr_family,
+	                        qdisc->handle,
+	                        qdisc->parent,
+	                        qdisc->info);
+
+	if (nm_streq0 (qdisc->kind, "fq_codel")) {
+		if (qdisc->fq_codel.limit)
+			nm_utils_strbuf_append (&buf, &len, " limit %d", qdisc->fq_codel.limit);
```
bengal: Since the member is unsigned, `"%u"` instead of `"%d"`?

src/platform/nm-linux-platform.c
lkundrak: True. Will fix.

src/platform/nm-platform.c
lkundrak: Yes

PR #339, opened 2019-04-09
team: clean up state when connection to teamd fails

bengal:
If NM fails to connect to teamd, it currently just sets the device
state to FAILED and waits that deactivate() is called later. However,
the 5 seconds timeout on teamd process start can hit in the meantime,
which fails with an assertion "nm_device_is_activating (device)".

Clean up the device state when the connection to teamd fails.

https://bugzilla.redhat.com/show_bug.cgi?id=1697900


thom311:
lgtm

bengal:
Merged as c48698d74778f1ac1c06bb8713b18e8d1edc1e24.

PR #343, opened 2019-04-12
core/qdisc: add support for attributes

lkundrak:


lkundrak:
@thom311 i believe i addressed points raised in #338 Thanks for the review there!

I'm wondering if you could take another look?

thom311:
> @thom311 i believe i addressed points raised in #338 Thanks for the review there!

Hi. You did not change the API ( https://github.com/NetworkManager/NetworkManager/pull/338#discussion_r273816177 ) and not comment why you think `char **` is so much better than `const char **` to warrant the copying overhead.

--- 

```
+         if (value)
+              rest = g_variant_dup_string (value, NULL);
+    }
+
     if (rest) {
          g_set_error (error, 1, 0, _("unsupported qdisc option: '%s'."), rest);
          return NULL;
     }
```

This is confusing, and I wonder whether it's wrong (or what it does). You didn't comment on what is special about key `""`. The documentation of `nm_utils_parse_variant_attributes()`  also doesn't mention anything special about `""`. If this code is really correct, then the `if (rest)` should just move above inside the `if (value)` block (and don't bother to `g_variant_dup_string()`).

---

I am going to merge the branch, to do the `1.18-rc1` release. We can still change API, if we want to (we also did that after `1.16-rc1`).

PR #345, opened 2019-04-15
build: default to internal DHCP client

lkundrak:
Meson builds already seem to default this and RHEL & Fedora switched
already. Everyone else also should.


thom311:
lgtm.

Maybe this is worth a NEWS entry? Can you add it now (on master branch), so we don't forget? It seems relevant for a package maintainer on update.

lkundrak:
Thanks. Applied with the NEWS entry.

PR #347, opened 2019-04-15
build: drop HAVE_WIFI define

lkundrak:
It's unused.


thom311:
It doesn't seem unused to me:
```
git grep -w WITH_WIFI origin/master
```

thom311:
ah, the define is unused. The automake condidtional is not. Never mind.

lgtm.

PR #348, opened 2019-04-18
contrib: Provide list of units to %systemd_postun

tyll:
%systemd_postun is meant to be run with arguments and Fedora Rawhide
seems to enforce this now. Therefore provide the units there, too.

This might also need to be fixed in  Fedora itself. Actual build error without this patch:
https://copr-be.cloud.fedoraproject.org/results/till/NetworkManager/fedora-rawhide-x86_64/00884799-NetworkManager/builder-live.log

```
error: This macro requires some arguments
error: line 809: %systemd_postun
```

### DIFF
--- a/contrib/fedora/rpm/NetworkManager.spec
+++ b/contrib/fedora/rpm/NetworkManager.spec
@@ -41,6 +41,7 @@
 
 %global real_version_major %(printf '%s' '%{real_version}' | sed -n 's/^\\([1-9][0-9]*\\.[1-9][0-9]*\\)\\.[1-9][0-9]*$/\\1/p')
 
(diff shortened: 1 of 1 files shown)


vhumpa:

tyll:
for reference, the systemd postun macro is defined here: https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in#L72

thom311:
Nice. I guess this fixes our build failures on Rawhide (e.g. https://gitlab.freedesktop.org/NetworkManager/NetworkManager/pipelines/31716)

thom311:
merged to master and nm-1-18.

thanks!

PR #351, opened 2019-05-24
WIP: dhcp: add nettools backend for DHCPv4

teg:
This adds a new dhcp backend, named 'nettools', which works largely like the current 'internal' one. See the commit message for details.

This is work in progress and mostly intended to gather feedback on both the n-dhcp4 API and on the integration itself.

By comparison, n-dhcp4 is more low-level than sd-dhcp4 (and other clients), as all options are considered opaque and left up to NetworkManager to interpret. The API is also more finegrained than the existing ones, so will allow usecases that were not possible before. In particular, we are able to integrate with DAD handling, and we do not require explicit support in the library for any options in order to use them from NetworkManager.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,9 +63,11 @@ test-*.trs
 
 /clients/cli/nmcli
 /clients/common/settings-docs.h
+/clients/common/tests/test-general
+/clients/common/tests/test-libnm-core-aux
+/clients/nm-online
 /clients/tui/newt/libnmt-newt.a
 /clients/tui/nmtui
-/clients/nm-online
 
 /data/NetworkManager.service
 /data/NetworkManager-wait-online.service
@@ -146,8 +148,8 @@ test-*.trs
 /examples/C/qt/list-connections
 /examples/C/qt/change-ipv4-addresses
 
-/shared/nm-version-macros.h
 /shared/nm-utils/tests/test-shared-general
+/shared/nm-version-macros.h
 
 /introspection/org.freedesktop.NetworkManager*.[ch]
 
@@ -277,8 +279,6 @@ test-*.trs
 /src/tests/test-systemd
 /src/tests/test-utils
 /src/tests/test-wired-defname
-
-/clients/common/tests/test-general
 
 /vapi/*.vapi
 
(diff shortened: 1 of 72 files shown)


vhumpa:

thom311:
Hi Tom, I merged the import of n-dhcp4 subtree to master (because it's annoying to rebase a merge-commit). I thereby transplanted the merge commit on master and added a comment to the commit message. Aside that, the import should be unchanged and correct. But that is why you get the merge-conflict now.

---

Do you have an account at https://gitlab.freedesktop.org/NetworkManager/NetworkManager ? That would be good, because if you open the pull request there, our CI runs as well.

---

Regarding our white space: we use everywhere "smart tabs" (tabs only used for indentation, but further alignment with space). I personally don't like that, because it requires the programmer to be smart about it, as the editor tends do to a poor job automatically. I find it cumbersome, but that's the style...

---

> nm_utils_error_set_literal (error, NM_UTILS_ERROR_UNKNOWN, "nettools plugin does not support IPv6");

Ah right. I wonder if we can beat nettools' DHCPv4 support into shape to put it into rhel-8.1. For example, it would potentially fix the DHCPDECLINE on ACD failure bug and it's the way forward anyway. Currently our DHCP plugins both are for IPv4 and IPv6. So, if we would do that, we would need to use "internal" (systemd) plugin for DHCPv6. That would be doable...

---

> You should have received a copy of the GNU General Public License along

`libnm` is LGPL2.0+ licensed while `src` is GPL2+. That mixing has historic reasons and is cumbersome, and we would very much like to go through the effort of relicense everything under LGPL2.0+. For that reasons, we require all new contributions to be LGPL2.0+, even if they are under "src/". Yes, we should document that better and in clean up this mess (TODO). Point being, do you agree that your contributions are under LGPL2.0+?

Usually, for all new source files we also want that they have a LGPL header. In this case, the code partly originates from a GPL licensed file, so doing so a bit problematic (although that file is almost entirely written by people who would agree with this change (`git log --pretty=format:'%an - %ai - %h - %s' --no-merges --follow -- src/dhcp/nm-dhcp-systemd.? | sort`) and at most it's inspired by systemd-networkd (which is LGPL already). *sigh*

teg:
Hi Thomas,

Thanks for the quick feedback! I'll rebase and file a PR on gitlab.

I confirm that all my contributions are under LGPL2.0+.

Would it make sense to just support DHCP 6 and 4 being used from different backends in general?

I'll go over the whitespace again.

Cheers,

Tom

thom311:
This merge request migrated to https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/171 . Closing...

PR #352, opened 2019-10-03
src/devices/nm-device.c: resolve possible null pointer dereference

chipitsine:
found by cppcheck

[src/devices/nm-device.c:3032] -> [src/devices/nm-device.c:3025]: (warning) Either the condition '!handle' is redundant or there is possible null pointer dereference: handle.

### DIFF
--- a/src/devices/nm-device.c
+++ b/src/devices/nm-device.c
@@ -3024,14 +3024,14 @@ concheck_cb (NMConnectivity *connectivity,
 	/* keep @self alive, while we invoke callbacks. */
 	priv = NM_DEVICE_GET_PRIVATE (self);
 
+	nm_assert (handle && c_list_contains (&priv->concheck_lst_head, &handle->concheck_lst));
+
+	seq = handle->seq;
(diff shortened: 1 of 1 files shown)


vhumpa:

chipitsine:
.ok\W+to\W+test.

thom311:
How does moving the logging line prevent the warning? 

That's probably due to the `nm_assert()`, which allows (and checks) for `handle`.

I think the only bug here is the assertion, that should instead be fixed to do:
```
nm_assert (handle && c_list_contains (&priv->concheck_lst_head, &handle->concheck_lst));
 ```

thom311:
if the assertion gets fixed, it does make sense to move the logging line after it!!

chipitsine:
yep, I meant that check `handle` against null should be done be dereferencing

chipitsine:
logging dereference `handle`, it should not occur before assert

thom311:
merged to master.

Thanks!!

PR #353, opened 2019-10-03
cppcheck findings

chipitsine:
found by cppcheck

!stream is true from line 666 and below, no need to check again

[src/nm-config-data.c:677] -> [src/nm-config-data.c:677]: (warning) Either the condition '!stream' is redundant or there is possible null pointer dereference: stream.

### DIFF
--- a/src/platform/nm-platform.c
+++ b/src/platform/nm-platform.c
@@ -4138,10 +4138,11 @@ nm_platform_ip4_address_sync (NMPlatform *self,
 		}
 	}
 	ip4_addr_subnets_destroy_index (plat_subnets, plat_addresses);
-	ip4_addr_subnets_destroy_index (known_subnets, known_addresses);
 
 	if (!known_addresses)
 		return TRUE;
+
+	ip4_addr_subnets_destroy_index (known_subnets, known_addresses);
 
 	ifa_flags =   nm_platform_kernel_support_get (NM_PLATFORM_KERNEL_SUPPORT_TYPE_EXTENDED_IFA_FLAGS)
 	            ? IFA_F_NOPREFIXROUTE


vhumpa:

chipitsine:
.ok\W+to\W+test.

thom311:
> .ok\W+to\W+test.

this has no effect and does not trigger a CI test. That is because you are not one of the trusted people :)

Also, I think nowadays the CI integration from github is no longer working. We use nowdays gitlab: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests

chipitsine:
github has very ancient "trusty" travis-ci images.
I thought it was done in purpose

thom311:
we have:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/pipelines
https://semaphoreci.com/lkundrak/networkmanager
https://travis-ci.org/NetworkManager/NetworkManager

probably one day in the future, travis and semaphore will be retired. gitlab-ci should do all the relevant unit tests.


Anway, above are only unit-tests. Those run always, and you don't need to get white listed (.ok\W+to\W+test). This would be necessary for our CI tests from https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci . These run more elaborate integration tests and run in internal infrastructure, which is why they are not run for "untrusted" pull-requests.

chipitsine:
I also found https://scan.coverity.com/projects/networkmanager-networkmanager
it is not being run on daily basis.

I can help to setup scheduled builds (either travis or gitlab-ci)

thom311:
automated coverity runs would be great.

Best triggered via a manual job in the `.gitlab-ci` file.

chipitsine:
ok, I'll make another pull request on that

src/platform/nm-platform.c
```diff
@@ -4138,11 +4138,12 @@ nm_platform_ip4_address_sync (NMPlatform *self,
 		}
 	}
 	ip4_addr_subnets_destroy_index (plat_subnets, plat_addresses);
-	ip4_addr_subnets_destroy_index (known_subnets, known_addresses);
 
 	if (!known_addresses)
 		return TRUE;
 
+	ip4_addr_subnets_destroy_index (known_subnets, known_addresses);
```
thom311: This warning is a false positive.

We can merge this commit, if only to shut up cppcheck.

But could you reword the commit message? I think there is no "possible" null pointer dereference.

And in the git-commit message, please always leave an empty line after the first one. `git log --oneline` will unwrap such line continuations, which is probably not intended here. Same with the parent commit.

src/nm-config-data.c
```diff
@@ -674,7 +674,7 @@ nm_config_data_log (const NMConfigData *self,
 		is_atomic = nm_config_data_is_intern_atomic_group (self, group);
 
 		_LOG (stream, prefix, "");
-		_LOG (stream, prefix, "[%s]%s", group, is_atomic && !stream ? " # atomic section" : "");
+		_LOG (stream, prefix, "[%s]%s", group, is_atomic ? " # atomic section" : "");
```
thom311: Do you know why does cppcheck complain about line 666? This doesn't make sense to me.

Anyway, it seems that glib's keyfile parser does not actually allow comments after `[group]`. While this is only for logging, we should log valid keyfile format, hence this comment should move to the following line, like
```C
_LOG (stream, prefix, "[%s]", group);
if (is_atomic)
    _LOG (stream, prefix, "# atomic section");
```

src/platform/nm-platform.c
chipitsine: function `ip4_addr_subnets_destroy_index` checks first argument against NULL, but does not check second one.

if known_addresses is NULL, so it will be dereferenced inside `ip4_addr_subnets_destroy_index`

moving `if(!known_addresses)` earlier than calling `ip4_addr_subnets_destroy_index` eliminates that situation.

I do not see it is being false positive.

src/nm-config-data.c
chipitsine: I need to have a deeper look at `_LOG` macro. probably, I'll drop this commit for now and reopen it later

src/platform/nm-platform.c
thom311: `@known_subnets` is only not NULL, if `@known_addresses` is not NULL. You can easily see that by looking at `nm_platform_ip4_address_sync()`.

chipitsine: there're couple of things

1) checking whether known_addresses is NULL or not
2) passing known_addresses to ip4_addr_subnets_destroy_index which dereferences known_addresses

either comparision should not happen at all, or it should happen earlier than dereference

thom311: The point is that the commit message claims there is a bug there (NULL pointer dereference). That would be a serious thing. Commit messages are important, and it should not make incorrect claims about the seriousness. I only complain here about the commit message!

> 2. passing known_addresses to ip4_addr_subnets_destroy_index which dereferences known_addresses

`known_addresses` is only dereferences if `known_subnets` is not NULL, and `known_subnets` is only not NULL if `known_addresses` is not NULL. The code has no possible NULL pointer dereference. 

Yes, the logic there might be improved (for the human reader, and not to confuse cppcheck). Not confusing tools (like cppcheck) and getting rid of false positives is important. Readable code is also important. So no argument there.

But note that while your patch makes it easier to see that `@known_addresses` is never dereferenced with NULL, it now might appear that `@known_subnets` gets leaked. Of course, with and without the patch neither happens, it's just not very obvious.

What would make it clear are assertions:

```diff
diff --git i/src/platform/nm-platform.c w/src/platform/nm-platform.c
index 4ec161adc3e4..754f2b1f64ef 100644
--- i/src/platform/nm-platform.c
+++ w/src/platform/nm-platform.c
@@ -3905,10 +3905,12 @@ ip4_addr_subnets_destroy_index (GHashTable *subnets, const GPtrArray *addresses)
     gpointer p;
 
     if (!subnets)
          return;
 
+    nm_assert (addresses);
+
     g_hash_table_iter_init (&iter, subnets);
     while (g_hash_table_iter_next (&iter, NULL, &p)) {
          if (!ip4_addr_subnets_is_plain_address (addresses, p))
               g_ptr_array_free ((GPtrArray *) p, TRUE);
     }
```

Assertions with `nm_assert()` are not compiled in production code, but enabled during tests. They can be used liberally.

chipitsine: well, it is not clear about `nm_assert`, what will happen in production then ?

chipitsine:
@thom311 , I have reviewed patch one more time.

patch suggests the following

```
	if (!known_addresses)
		return TRUE;

	ip4_addr_subnets_destroy_index (known_subnets, known_addresses);
```

instead of 

```
       ip4_addr_subnets_destroy_index (known_subnets, known_addresses);

	if (!known_addresses)
		return TRUE;

```


I see it is correct by its nature. I do not understand your point regarding that

thom311:
> I do not understand your point regarding that

The point is that the commit message claims that there is a possible NULL pointer reference, when there is not.

chipitsine:
commit message is correct.

comparing known_addresses against NULL implies it might be NULL. passing it to ip4_addr_subnets_destroy_index dereferences it.

it does not make sense to pass (i.e. dereference) first and compare against null after that

thom311:
If the commit message is correct, how could the "possible null pointer derefence" happen?

`ip4_addr_subnets_destroy_index()` only dereferences `@known_addresses` if `@known_subnets` is not `%NULL`. On master, no possible code path will call `ip4_addr_subnets_destroy_index()` so that `@known_subnets` is not equal `%NULL` while `@known_addresses` is `%NULL`.

Are you aware that above there is:
```C
»···»···if (known_addresses) {
...
»···»···»···»···if (!known_subnets)
»···»···»···»···»···known_subnets = ip4_addr_subnets_build_index (known_addresses, FALSE, FALSE);
```

It seems, no possible code path (on master) will lead to a "null pointer dereference".

chipitsine:
so, should we just move `if (!known_addresses)` inside `ip4_addr_subnets_build_index` ?

chipitsine:
ok, is commit message "make cppcheck a bit happier" better ?

bengal:
As explained by Thomas, this is a false positive. I've reworded the message of the commit and merged it to master as c0f4c2f2e67a3afd62c3a154887c11f412647ed9. Thanks!

PR #355, opened 2019-10-09
contrib: move scripts from test/ to scripts/

lkundrak:
There's no reason to have them separate. Also add the missing executable
bit to btmodem.pl.


thom311:
lgtm

PR #356, opened 2019-10-09
Handle routes even when there are no rules

yann-soubeyrand:
Hi,

The script exits when rule configuration files don't exist. This prevents it from handling routes when there are no rules. The check for the rule configuration file is also done later and checks for additional rule files. Therefore, it seems that we can safely remove this early check.

### DIFF
--- a/examples/dispatcher/10-ifcfg-rh-routes.sh
+++ b/examples/dispatcher/10-ifcfg-rh-routes.sh
@@ -22,10 +22,6 @@ file_regex='^/etc/sysconfig/network-scripts/ifcfg-([^/]+)$'
 
 profile="${BASH_REMATCH[1]}"
 
-if [ ! -f "/etc/sysconfig/network-scripts/rule-$profile" ] && [ ! -f "/etc/sysconfig/network-scripts/rule6-$profile" ]; then
-    exit 0
-fi
-
 MATCH='^[[:space:]]*(\#.*)?$'
 
 handle_file () {


vhumpa:

thom311:
if there are no rule files, then NetworkManager handles the routes.

Why do you want the script to handle them?

yann-soubeyrand:
@thom311 During my test, NetworkManager did not create routes with multiple next hops like this one
```
10.10.21.0/24 nexthop via 10.10.20.101 nexthop via 10.10.20.102
```
The solution was to add the dispatcher script and remove the rule configuration file existence check.

thom311:
Yes, NetworkManager does not support multipath routes. It's a missing feature, of course it should. 

As NM does not support it, it begs the question how are you going to configure such routes?

The dispatcher scripts are a possible solution and exist for this reason. I just don't think that it's right to install such a script by default. If such a modification of the script works for you, you may very well copy the script, adjust it, and use it. That's why dispatcher scripts exist.

Of course, you have different problems. You can only configure the routes by editing the ifcfg-rh files yourself (which is not the problem). The problem is when you use NetworkManager API to change the profile, then your modifications are overwritten. I guess, you could avoid that by adding an (empty) rule file. The presence of rule files tells NM to assume there are no manual routes, and it's up to the user to create them.
If you anyway have your own set of dispatcher scripts, you could also patch it to not use the "route-" files, but your own files where you edit such routes. Then whenever NM would rewrite the ifcfg files, it would not mess with your manually edited parts.

yann-soubeyrand:
Hum, handling of rule-* and route-* files by NM seems to be a lot more complicated than what I understood, so feel free to close this PR if you think this is a bad idea (or simply a wrong thing) to suppress this early check ;-)

thom311:
OK.

It doesn't seem so complicated though :)

If you want that the existing dispatcher script handles the routes for you, just create an empty rule file. That will work pretty much was you wish, with the only downside that you cannot use D-Bus API in NetworkManager to configure manual routes. But that is because NM should not rewrite your manually crafted route-files (that it doesn't fully understand).


The real solution is to support multipath routes in NM, but that's more work.

PR #360, opened 2020-03-29
Change nmcli to use XDG_DATA_HOME instead of HOME

estheruary:
Right now nmcli saves its history into ~/.nmcli-history. This would change it so that it stores it in XDG_DATA_HOME (i.e. ~/.local/share/nmcli-history for most users).

### DIFF
--- a/clients/cli/connections.c
+++ b/clients/cli/connections.c
@@ -6344,7 +6344,7 @@ nmcli_editor_tab_completion (const char *text, int start, int end)
 	return match_array;
 }
 
-#define NMCLI_EDITOR_HISTORY ".nmcli-history"
+#define NMCLI_EDITOR_HISTORY "nmcli-history"
 
 static void
 load_history_cmds (const char *uuid)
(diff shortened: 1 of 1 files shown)


vhumpa:

thom311:
Thank you. I think this is something we want to do.

But we cannot just abandon the old file location. There needs to be fallback code to check whether `~/.nmcli-history` exists, and use that, if present.

There is some code duplication in the pre-existing code, that makes it harder to implement this change. First, the duplicated code should be unified by having a function that opens the keyfile from the right location.

Unrelated, we use github only as a mirror. We would really appreciate merge requests being made against https://gitlab.freedesktop.org/NetworkManager/NetworkManager . Did some outdated docs point you here? Which?
gitlab works essentially the same as github (log-in, fork the repo, create a pull/merge request). I think in github, pull requests cannot be disabled, otherwise we would. Could you please re-open the issue against our gitlab instance? Thank you!!

thom311:
Ah, the topic at https://github.com/NetworkManager/NetworkManager was misleading about creating pull requests here. Updated.

The reason we prefer pull requests at freedesktop.org is because then there is only one location for pull requets, and we run there our unit tests with gitlab-ci.

PR #363, opened 2022-03-27
Revert "libnm: buffer reads in nm_vpn_service_plugin_read_vpn_details()"

brjacobsatlassian:
This partially reverts commit 4a9fcb0fc32ed67646843312a6cbc706ec3fee45, which
replaced one-byte reads with buffered ones in the VPN service plugin.

Unfortunately the buffering means that commands coming after the magic "DONE"
string were being pulled into the buffer. Secrets agents expect a "QUIT" to
come after the "DONE", and since with buffering "QUIT" was in the buffer, this
led to a twenty-second delay on every VPN connection using a secrets manager.

Yes, EVERY SINGLE connection. A twenty second delay while the secrets manager timed out.

Performance optimizations are only good where they improve performance; the intent of the previous commit was good but the outcome was not.

### DIFF
--- a/src/libnm-client-impl/nm-vpn-service-plugin.c
+++ b/src/libnm-client-impl/nm-vpn-service-plugin.c
@@ -729,54 +729,6 @@ nm_vpn_service_plugin_secrets_required(NMVpnServicePlugin *plugin,
 
 /*****************************************************************************/
 
-typedef struct {
-    char *buf;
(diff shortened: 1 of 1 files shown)


thom311:
this was reopened at https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1164. Closing here.

PR #365, opened 2022-12-05
fix: to fix 'Wstringop-truncation' related compile warning

kklionz:
There are some compile warning while building with gcc 7
```
./src/libnm-glib-aux/nm-macros-internal.h:1379:54: warning：‘#pragma GCC diagnostic’ unknown option  [-Wpragmas]
     NM_PRAGMA_WARNING_DISABLE("-Wstringop-truncation");
                                                      ^
  CC       src/libnm-lldp/libnm_lldp_la-nm-lldp-network.lo
```

Refer to https://gcc.gnu.org/gcc-8/changes.html, Wstringop-truncation is introduced with GCC 8.1

So,  this PR is created to fix the compile warning

Could U help to review it? thanks

### DIFF
--- a/src/libnm-glib-aux/nm-macros-internal.h
+++ b/src/libnm-glib-aux/nm-macros-internal.h
@@ -1376,13 +1376,17 @@ nm_memdup_nul(gconstpointer data, gsize size)
 static inline char *
 _nm_strndup_a_step(char *s, const char *str, gsize len)
 {
+#if defined(__GNUC__) && (__GNUC__ > 8 || (__GNUC__ == 8 && __GNUC_MINOR__ >= 1))
     NM_PRAGMA_WARNING_DISABLE("-Wstringop-truncation");
+#endif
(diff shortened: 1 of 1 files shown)


thom311:
While NetworkManager enables many warnings and wants to build without warnings, it explicitly sets `-Wno-pragmas`. This warning just doesn't seem useful, because we mostly use pragmas to suppress warnings. If that generates a warning saying it doesn't understand the warning that should be suppressed, it's not useful.

How do you build NetworkManager? Both meson and autotools should pass `-Wno-pragmas`.

Btw, we don't do pull requests on github. The right place would be https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests

thom311:
closing. Please open merge requests against our [gitlab](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests).

Also, the patch would require a different solution, as I commented earlier.

PR #368, opened 2024-01-05
8021x openssl ciphers

ebikt:
Allow user to specify TLS ciphers for wpa_supplicant in 802.1x.
Some old RADIUS servers support only ciphers that were recently moved to DEFAULT@SECLEVEL=0 in openssl.
Debian based distributions compile wpa_supplicant with default cipher string DEFAULT@SECLEVEL=1, which is good thing in general.

This patch aims to allow to connect to networks with old radius servers while not compromising security on other networks.

Patch is targeted only at `wpa_supplicant` because that 802.1x implementation is used in debian and because I do not have access to test other implementations.

Redhat settings conversion is provided without testing (I do not use any redhat-based distribution).

I also added support for this config to nmtui, so that user can edit their connections to old radius servers at least there, because I do not expect graphical user interfaces to incorporate this change (soon).

Tested against Windows 2008 radius servers from debian testing (patch applied to network manager version 1.44).


ebikt:
Created issue with link to this pull-request on gitlab https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1457

thom311:
As you found out, NetworkManager is maintained at gitlab.freedesktop.org. The github project is only a fork (and the option to create pull requests here cannot be disabled).

Please open the merge request [there](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests).

Closing this one. Sorry for the inconvenience.

ebikt:
@thom311 I cannot open merge request there, as my `fork` button is greyed out in freedesktop gitlab.

ebikt:
I'll try to open issue similar to https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/575, to let administrators know about my problem.

ihuguet:
Please see: https://gitlab.freedesktop.org/freedesktop/freedesktop/-/wikis/home#how-can-i-contribute-to-an-existing-project-or-create-a-new-one

PR #370, opened 2024-09-12
wifi: add block scan list to a device section of key file.

nazarenko-roman:
On system startup (not all wireless devices are configured) NM starts periodic scans on all of interfaces. It creates a load on the dbus and NM, some services are failing with timeout because of it. Usually, it happens when a system has several virtual interfaces on one physical chip - wpa_supplicant sends scan results to all siblings. For example, if the system has 3 virtual interfaces, NM starts 3 scans and receives 3 scan results on each. So NM receives 9 scan results (mainly with the same APs since all scans are done on one chip).
This commit extends the configuration file, adds a possibility to specify device list for which a scan is unnecessary to avoid that situation.

scan-issue-3-interfaces

### DIFF
--- a/man/NetworkManager.conf.xml
+++ b/man/NetworkManager.conf.xml
@@ -1399,6 +1399,16 @@ managed=1
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term><varname>wifi.periodic-scan-block-list</varname></term>
+          <listitem>
+            <para>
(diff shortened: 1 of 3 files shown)


bengal:
This repository on github is only a mirror of the one on https://gitlab.freedesktop.org/NetworkManager/NetworkManager . Please open a merge request there, thank you.

PR #371, opened 2024-11-12
dispatcher: Check for pending dbus stop job before disptching action.

RN97:
We see a spurious log when NetworkManger attempts to invoke dispatcher via the dbus when a dbus stop job is queued.
This issue has been reported to the officual github under issue-1643.

This patch aims to fix this by checking for a pending dbus stop job and not call NM-dispatcher in that case.

### DIFF
--- a/src/core/nm-dispatcher.c
+++ b/src/core/nm-dispatcher.c
@@ -50,6 +50,10 @@
     }                                                         \
     G_STMT_END
 
+#define SYSTEMD_DBUS_SERVICE   "org.freedesktop.systemd1"
+#define SYSTEMD_DBUS_PATH      "/org/freedesktop/systemd1"
+#define SYSTEMD_DBUS_INTERFACE "org.freedesktop.systemd1.Manager"
+
 /*****************************************************************************/
 
 /* Type for generic callback function; must be cast to either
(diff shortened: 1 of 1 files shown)


RN97:
This is the PR that addresses the issue being tracked in issue #1643
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1643

thom311:
please open merge requests against https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests

src/core/nm-dispatcher.c
```diff
@@ -836,6 +844,56 @@ _dispatcher_call(NMDispatcherAction    action,
...
+                                       SYSTEMD_DBUS_SERVICE,
+				       SYSTEMD_DBUS_PATH,
+				       SYSTEMD_DBUS_INTERFACE,
+				       "GetUnit",
+				       g_variant_new("(s)", "dbus.service"),
+				       G_VARIANT_TYPE("(o)"),
+				       G_DBUS_CALL_FLAGS_NONE,
+				       CALL_TIMEOUT,
+				       NULL,
+				       &sd_err);
```
thom311, line 857: I think there should be no blocking calls to external services from NetworkManager.

Also, the overhead that every call to the dispatcher involves now two additional D-Bus calls doesn't seem acceptable to me.

thom311: you will have to monitor (and cache) what is going on on systemd. That way the value is cached and readily accessible.

Also, we should support systems without systemd.

PR #372, opened 2025-02-26
NetworkManager periodically executes the periodic_update method to re…

ut003640:
…fresh signal measurements. When external active scanning occurs, it triggers wpa_supplicant to update BSS information. Upon receiving BSS update signals, NetworkManager proactively retrieves current WiFi properties from wpa_supplicant for updates. However, discrepancies exist between the signal strength values reported by wpa_supplicant and those obtained through NetworkManager's periodic_update method. Therefore, we should refrain from updating signal strength metrics when handling bss_update events to maintain measurement consistency


bengal:
This is a read-only mirror of https://gitlab.freedesktop.org/NetworkManager/NetworkManager , please open a MR there. Thank you.

PR #373, opened 2025-03-17
Ih/nmcli ovs port type

ihuguet:
If the connection is a port we need to set the connection.port-type
property. Usually this property is guessed by nmcli depending on the
connection type or the chosen controller, so it doesn't need to be
specified by the user. However, if it is explicitly set by the user
we should not guess, but just use it.

When we process arguments like "controller" or "type" we call custom
functions like set_connection_controller that will guess the port-type
if needed. By processing port-type first, it will be set in the
connection by the time that these other properties are processed, so they
won't try to guess.

After port-type, process connection.type, as we are usually capable of
deducing the port-type from it.

Also, add some other minor fixes and improvements to nmcli's handling of
port-type.

Resolves: https://issues.redhat.com/browse/RHEL-80273

### DIFF
--- a/src/nmcli/connections.c
+++ b/src/nmcli/connections.c
@@ -4615,8 +4615,11 @@ set_connection_type(NmCli            *nmc,
                     gboolean          allow_reset,
                     GError          **error)
 {
-    GError     *local     = NULL;
-    const char *port_type = NULL;
+    NMSettingConnection *s_con     = nm_connection_get_setting_connection(con);
+    GError              *local     = NULL;
+    const char          *port_type = NULL;
+
+    nm_assert(s_con);
 
     value = check_valid_name_toplevel(value, &port_type, &local);
     if (!value) {
@@ -4631,7 +4634,7 @@ set_connection_type(NmCli            *nmc,
         return FALSE;
     }
 
-    if (port_type) {
+    if (!nm_setting_connection_get_port_type(s_con) && port_type) {
         if (!set_property(nmc->client,
                           con,
                           NM_SETTING_CONNECTION_SETTING_NAME,
--- a/src/nmcli/connections.c
+++ b/src/nmcli/connections.c
@@ -3860,6 +3860,11 @@ check_valid_name_toplevel(const char *val, const char **port_type, GError **erro
         return NM_SETTING_WIRED_SETTING_NAME;
     }
 
+    if (nm_streq(str, "ovs-port"))
+        NM_SET_OUT(port_type, NM_SETTING_OVS_BRIDGE_SETTING_NAME);
+    else if (NM_IN_STRSET(str, "ovs-interface", "ovs-patch"))
+        NM_SET_OUT(port_type, NM_SETTING_OVS_PORT_SETTING_NAME);
+
     setting_info = nm_meta_setting_info_editor_find_by_name(str, TRUE);
     if (setting_info)
         return setting_info->general->setting_name;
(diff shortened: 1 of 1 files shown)


src/nmcli/connections.c
```diff
@@ -3829,8 +3829,11 @@ check_valid_name_toplevel(const char *val, const char **port_type, GError **erro
         if (setting_info->alias)
             g_ptr_array_add(tmp_arr, (gpointer) setting_info->alias);
     }
+    g_ptr_array_add(tmp_arr, "bond-port");
     g_ptr_array_add(tmp_arr, "bond-slave");
+    g_ptr_array_add(tmp_arr, "bridge-port");
     g_ptr_array_add(tmp_arr, "bridge-slave");
+    g_ptr_array_add(tmp_arr, "team-port");
```
bengal, line 3836: Those `*-slave` connection types are now obsolete and should never be used. They implicitly create an "ethernet" connection and this is confusing because the name suggests it's just a port/slave that can be of any type. For this reason, it's always better to use the explicit `type * port-type *` syntax. All the documentation now should refer to the new, better way and I think we need to find a way to deprecate and remove those `*-slave` connection types.

src/nmcli/connections.c
```diff
@@ -4615,8 +4615,11 @@ set_connection_type(NmCli            *nmc,
                     gboolean          allow_reset,
                     GError          **error)
 {
-    GError     *local     = NULL;
-    const char *port_type = NULL;
+    NMSettingConnection *s_con     = nm_connection_get_setting_connection(con);
```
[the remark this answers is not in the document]
ihuguet: I didn't test, but if you look at the code of set_connection_type, it calls to check_valid_name_toplevel where it may deduce a port-type (not in all cases). If it got a port-type, it always set it, without checking if we already had one.

I guess that you can pass `port-type bridge bond-port.queue-id 123` and it will overwrite port-type=bridge with port-type=bond.

I will test it to be sure, but the patch is OK because it's more defensive without breaking any valid use case, IMO.

src/nmcli/connections.c
```diff
@@ -5236,9 +5236,69 @@ nmc_process_connection_properties(NmCli              *nmc,
...
     /* First check if we have a port-type, as this would mean we will not
      * have ip properties but possibly others, port-type specific.
+     * Then check connection.type, as port-type might be deduced from it.
      */
+    if (!nmc->complete) {
+        const char       **dst;
+        const char *const *src;
+        const char        *option;
+
+        dst  = g_alloca(*argc * sizeof(const char *) + 1);
```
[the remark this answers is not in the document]
ihuguet: Yes, you are right. Changed to g_malloc.

src/nmcli/connections.c
ihuguet: Dropped that change.

src/nmcli/connections.c
```diff
@@ -5236,9 +5244,70 @@ nmc_process_connection_properties(NmCli              *nmc,
...
+        const char *const *src;
+        const char        *option;
+
+        to_free = g_malloc(*argc * sizeof(const char *) + 1);
+        args = dst = to_free;
+
+        for (src = *argv; src < *argv + *argc - 1; src += 2) {
+            option = (**src == '+' || **src == '-') ? *src + 1 : *src;
+            if (NM_IN_STRSET(option,
+                             "connection.port-type",
```
bengal, line 5265: Consider using macros for setting and property names (e.g. `NM_SETTING_CONNECTION_SETTING_NAME "." NM_SETTING_CONNECTION_PORT_TYPE` . It is less readable, but allows one to grep the code for the property or the setting.
Apart from this optional suggestion, the rest of the branch LGTM.

